PR #758, opened 2013-06-24
Softdeleteable related updates

### LINKED ISSUE #554: Nested Set and Soft delete combination problem
Hello, I tried to use Nested Set tree and Soft Delete extensions on the same Entity. But I saw that it produce problems. For example, if I delete (SoftDelete) some entity in the middle of the tree, other entity property values (lft, rgt, ...) will be properly updated, but the same properties of softly deleted entity will not be updated. This cause problem when I want to verify Tree structure:

$entityRepository->verify();

It returns errors.

So it looks like combining of Tree Extension and Soft Delete extension is not compatible. Maybe I am doing something wrong? Or there is some solution?

### LINKED ISSUE #739: [SoftDeleteable] How to undelete? (and make it play nicely with Tree)
Hi,

I'm having trouble figuring out how to undelete an item that has been SoftDeleted. I tried doing `$item->setDeletedAt(null);` but it doesn't seem to do anything when flush()ed.

Even if it did... does this fire an 'undelete' event? Other DoctrineExtensions would need to be able to detect this so they could work their magic. E.g. If you have a SoftDeleteable Tree, then Tree would need to reinsert the previously deleted item back into the hierarchy _somewhere_ ... maintaining it's previously set parent, perhaps throwing an Exception if the parent is also deleted?

Is there any way to do this and if not where should I start to look for more info about how to implement it!

### LINKED ISSUE #754: [Translatable]  vs [SoftDeleteable]
I has entity with translatable and softdeletable turned on:

http://pastebin.com/m7MtpgSX

Now when I delete object of this entity, it is soft deleted, but translations for it are permanently removed. Am I missing something, or it is bug?

How can I correct this behavior?

l3pp4rd:
closes #739
closes #554
fixes #754 

Handles softdeleted slug uniqueness checks implicitly. Here is some minnor BC break if `addManagedFilter($name)` was used explicitly for sluggable.
### Points of improvement
- [x] make filter ignorance available at base listener class, to share common behavior between listeners
- [x] implicitly manage softdeleteable filter for sluggable listener without any user intervention.
- [x] update tree **lft** **rgt** values, when nodes were softdeleted
- [x] handle translations related to an entity, do not remove them

### DIFF
--- a/lib/Gedmo/Mapping/MappedEventSubscriber.php
+++ b/lib/Gedmo/Mapping/MappedEventSubscriber.php
@@ -114,13 +114,35 @@ abstract class MappedEventSubscriber implements EventSubscriber
             $filters = $this->getFilterCollectionFromObjectManager($om);
             $enabled = $filters->getEnabledFilters();
             foreach ($enabled as $name => $filter) {
-                if (in_array(get_class($filter), $this->ignoredFilters)) {
-                    $filters->disable($name);
-                    $this->disabledFilters[] = $name;
+                foreach ($this->ignoredFilters as $filterClassName) {
+                    if (is_a($filter, $filterClassName)) {
+                        $filters->disable($name);
+                        $this->disabledFilters[] = $name;
+                        break;
+                    }
                 }
             }
         }
     }
+
+    /**
+     * Checks if $filterClassName filter is enabled
+     *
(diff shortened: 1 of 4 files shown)


RaedsLab:
Is this still unmanaged ?

PR #764, opened 2013-07-01
[WIP] v3.0.0

### LINKED ISSUE #100: [Sluggable] uniqueness breaks when combined with translations
Entity Definition:

```
// Article.php
/**
 * @Gedmo\Sluggable
 * @Gedmo\Translatable
 * @ORM\Column(type="string", length="128")
 */
protected $title;

/**
 * @Gedmo\Slug
 * @Gedmo\Translatable
 * @ORM\Column(type="string", length="128", unique=true)
 */
protected $slug;
```

This will generate two identical slugs in the translation table:

```
$article = new Article();
$article->setTitle("title");
$em->persist($article);
$em->flush();
$em->clear();

$em->persist($article);
$em->flush();
```

### LINKED ISSUE #212: [Sortable] refactor
Refactor sortable in order to achieve these goals:
- Multiple sortable columns per object, like Slugs etc. support
- Synchronization with identityMap of UnitOfWork so that all loaded objects synchronized with updates executed during flush() in single request **(fixed)**
- ODM support

### LINKED ISSUE #267: translatable : strange cast error when using the query hint way
Hi
Using the query hint way as explained at
https://github.com/l3pp4rd/DoctrineExtensions/blob/master/doc/translatable.md#orm-query-hint
The SQL query returns a cast error between the integer id of the entity and the string(64) foreign_key of the translation table. I resolved it changing the foreign_key type to integer, in the mapping file Resources / config / doctrine / AbstractTranslation.orm.xml, but I suppose that isn't the good way !

### LINKED ISSUE #282: Incorrect results while using Translatable + 'order by'
The incorrect result seem to come from the fact that all the values are stored in plain text and sorting text column is entirely different world than sorting integer column.

Example:
id | title | sort_order
1  | a     | 5
2  | a     | 28

Query: select.. order by sort_order DESC, title ASC

While checking the results you will notice that the item 1 is before item 2, which would mean that the sort_order is higher, which is not the case.

### LINKED ISSUE #284: Refactor Translatable
Recently some new issues were reported conserning the **cast** problems for translation hints. The issue maintly is based on the fact that all translation are storted in the **long text** field. Which in nature, also does not allow to do a searching or indexing efficient based on translations. Regarding these facts, I thought of refactoring for version 2.4.x of extensions, which would include:
### Main points of improvement
- [ ] have translation entities related for each entity as currently personal translations do
- [ ] translations should contain all translatable fields
- [ ] other extensions should be able to apply directly on translation (sluggable as an example)
- [ ] query hint for translation should do only one join (inner or left), should not require cast for columns
- [ ] should be able to support default locale translation as fallback
### Optional points of improvement
- [ ] may have translation generator command
### References
- https://groups.google.com/forum/?fromgroups#!topic/doctrine-user/bXRzMzRhlFQ
- https://gist.github.com/sandermarechal/5661187#file-post-php

The example model of **Article** and **ArticleTranslation**

``` php
<?php

namespace Entity;

/**
 * @ORM\Entity
 */
class Article
{
    /**
     * @ORM\Id
     * @ORM\GeneratedValue
     * @ORM\Column(type="integer")
     */
    private $id;

    /**
     * @Gedmo\Translatable
     * @ORM\Column(length=128)
     */
    private $title;

    /**
     * @Gedmo\Translatable
     * @ORM\Column(type="integer")
     */
    private $level;

    /**
     * Custom typed content
     *
     * @Gedmo\Translatable
     * @ORM\Column(type="custom", nullable=true)
     */
    private $content;
}
```

And the generated **ArticleTranslation**

``` php
<?php

namespace Entity;

use Gedmo\Translatable\Entity\AbstractTranslation;

/**
 * Indexes should by added by user
 *  translation generator command if executed twice
 *  should only add new columns, or skip modifications at all.
 *  at best, user should control, and update the schema
 *
 * @ORM\Entity
 */
class ArticleTranslation extends AbstractTranslation // contains locale property
{
    /**
     * Can be and should be modified by user:
     *   - specify inverse list of translations
     *   - cascade options
     *
     * Regarding the inversed translations, there could be a Trait
     * generated also. It could contain functionality like TranslatorProxy
     * does, array access and so on.
     *
     * @ORM\ManyToOne(targetEntity="Article", inversedBy="translations")
     * @ORM\JoinColumn(name="object_id", referencedColumnName="id", onDelete="CASCADE")
     */
    private $object;

    /**
     * @ORM\Column(length=128)
     */
    private $title;

    /**
     * @ORM\Column(type="integer")
     */
    private $level;

    /**
     * Custom typed content
     *
     * @ORM\Column(type="custom", nullable=true)
     */
    private $content;

    // setters and getters follow
}
```

As mentioned before, it is planned for version 2.4.x and won't be available very soon. Please post in the comments your opinion and or ideas to be included.

### LINKED ISSUE #370: [sortable] removing entities causes DQL error
When removing entities from sortable crashes, in sortableListener at line 328 with invalid SQL:

``` sql
UPDATE SCMS\Entity\NewsArticlePhoto n SET n.position = n.position - 1 WHERE n.position >=  AND n.news_article = :val___0
```

``` php
array (size=3)
  'start' => null
  'stop' => int -1
  'delta' => int -1
```

tested on v2.3

### LINKED ISSUE #371: Translatable behavior breaks for all translatable fields if UniqueEntity constraint is applied on any field
Hello,

I am using Gedmo Translatable and Sluggable behaviors with StofDoctrineExtensionsBundle. 
Here is the code:

``` php
use Symfony\Component\Validator\Constraints as Assert;
use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;

/**
 * GD\AdminBundle\Entity\Tag
 *
 * @ORM\Table(name="tags")
 * @ORM\Entity(repositoryClass="GD\AdminBundle\Repository\TagRepository")
 * @Gedmo\TranslationEntity(class="GD\AdminBundle\Entity\TagTranslation")
 * @UniqueEntity("slug")
 */
class Tag
{
    /**
     * @var integer $id
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var string $name
     *
     * @Gedmo\Sluggable(slugField="slug")
     * @ORM\Column(name="name", type="string", length=255)
     * @Gedmo\Translatable
     */
    private $name;

    /**
     * @var string $slug
     * @Gedmo\Slug(updatable=true)
     * @ORM\Column(name="slug", type="string", length=255)
     */
    private $slug;
```

When using a UniqueEntity constraint, the "name" field stops updating. That is, the corresponding row is not updated in the `tag_translations` table for the locale.

Is this expected behavior?

I have tested the same scenario with other entities that have more multiple translatable fields too.

### LINKED ISSUE #379: Doctrine + Multiple entity managers + Translation
In one of my projects i'm using DoctineExtensions (Translation) and this works fine. Now that i've added a second entity manager, i get MappingExceptions. For example:

No identifier/primary key specified for Entity 'Company\ShopBundle\Entity\PageTranslation'. Every Entity must have an identifier/primary key.

The error appears:
1. When updating my doctrine schema using the commandline
2. When accessing a page in my project that contains a ....Translation table and the default locale is not active.

There are two things i can at the moment to get rid of the error:
1. Disable the 2nd entity manager (not really an option because i need this)
2. Disable everything from the stof_doctrine_extensions in the config.yml. (this breaks the translation, so not really an option too)

I'm using the Symfony version 2.0.12.

### LINKED ISSUE #381: "You cannot define a mapping item when in a sequence" error
I have upadated vendors yesterday and such error occured:

You cannot define a mapping item when in a sequence in "[...]BackendBundle\Resources\config\doctrine\/Entity.orm.yml" 

Stack Trace
1. in [...]\vendor\symfony\symfony\src\Symfony\Component\Yaml\Parser.php at line 114 
2. at Parser ->parse ('- translatable slug: separator: - style: default updatable: true fields: - name ')
   in [...]\vendor\symfony\symfony\src\Symfony\Component\Yaml\Parser.php at line 179
3. at Parser ->parse ('type: string length: 255 gedmo: - translatable slug: separator: - style: default updatable: true fields: - name')
   in [...]\vendor\symfony\symfony\src\Symfony\Component\Yaml\Parser.php at line 179 

Seems that error comes from translatable extension or maybe translation of slug triggers error

working on SF2 vendors older then few days gives me no errors

my definition for translatable slug field:

```
 slug:
      type: string
      length: 255
      gedmo:
        - translatable
        slug:
          separator: -
          style: default
          updatable: true
          fields:
            - name
```

removing translatable fixed error

please help

### LINKED ISSUE #391: more detailed info about the advance use of the translation listener
hi im having a problem when updating records with the same field value in multiple languages, in example lets say i want to translate the company entity, and my default locale is 'es' so i save the record for The Cola Company with the title The Cola Company, description in spanish and then i want to translate that record usin the uri locale, that works only if i change the title field , so i'm wondering if that can be solve by setting the translation fall back to true, and if so how can i instantiate the translation listener ?.

### LINKED ISSUE #404: [Translatable] issue with HINT_INNER_JOIN
I can't get this to work on some queries, is this a known issue ? Sometime I juste end up in an endless loop or so it seems, request taking endless time to load but no memory limit crash.

If I die($query->getSQL()) just after setting the hint same issue.

Here is a query that does not work, but is working if I remove the inner_join hint :

``` php
<?php
        $repo  = $this->getDoctrine()->getRepository('FooBundle:Hotel');
        $queryBuilder = $repo
            ->createQueryBuilder('h')
        ;

        $query = $queryBuilder
            ->getQuery()
            ->setHint(\Doctrine\ORM\Query::HINT_CUSTOM_OUTPUT_WALKER, 'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker')
            ->setHint(\Gedmo\Translatable\TranslatableListener::HINT_TRANSLATABLE_LOCALE, $this->getActiveCulture())
            ->setHint(\Gedmo\Translatable\TranslatableListener::HINT_FALLBACK, 0)
            ->setHint(\Gedmo\Translatable\TranslatableListener::HINT_INNER_JOIN, 1)
        ;
```

### LINKED ISSUE #415: personal translations and sluggable
Hello, 

I'm using personal translations and a sluggable field. When I insert a record I get the slug only for default locale. The slug for other locales is not created. 
How is it supposed to work?

Thanks.

### LINKED ISSUE #417: Elements without translation, how to show the word from entity table?
Hi,

I have a nested tree (tree doctrine extension) called place_tree with countries, counties, and towns. I created the translations for the countries, but I do not need translations for the names of the towns, so I don't know how to show the town's name stored in the Place table..(Now the names of the are just now shown when I invoque using a language different to the default locale).

Javi

### LINKED ISSUE #430: Translatable with Symfony 2.1 doesn't work
Hi,

I would like to use Translatable Extension with Symfony 2.1 but only my translation on parent returned.

On my table ext_translations, i have a lot of translations for the same entity.

I don't like to use the HINT. Do you have a best solution for that ?

Thanks for your help

Bertrand / Garfield-fr

### LINKED ISSUE #436: [documentation] metadata cache driver more then required
What I noticed lately is that DoctrineExtensions really need metadata cache driver like apc or memcached. 
Standard array setting does not work at all - performance drop while metadata object is generated is huge. 
Same entity without any extensions enabled can have metadata generated in unnoticeable time but with enabled 3-4 extensions this time can be over 1 second and its just for one entity class.

Imho adding tips about this matter to docs (maybe here: https://github.com/l3pp4rd/DoctrineExtensions/blob/master/doc/symfony2.md) would be really useful for many users.

For example I use metadata cache even on dev env to save around 2-3 seconds on every request.

Regards

### LINKED ISSUE #448: [translatable] postLoad not working with inheritance
When I try to translate entity that is part of inheritance with personal translations, it queries default translation table(instead one with personal translations). Reason for this is that TranslationListener::$configuration is not populated with ineherited entity configuration(only base class configuration) and translatable listener uses default table instead.

### LINKED ISSUE #464: [Tree] Allow childrenHierarchy() to return a nested set of Entities instead of flat arrays
It probably is by design but in many cases i don't need a flat array representation of my Entities but the Object form. Is it possible somehow to retrieve a tree structure of Entities in object form instead of flat arrays?

### LINKED ISSUE #479: [Translatable] PostPresist on Postgresql results into null value for column "id" .
When TranslatableListener calls $ea->insertTranslationRecord($translation); id is null. It should pull next id from sequnce.  It works on mysql if you have an autoincrement for an id field, but not in case of postgres.

$product = new Product();
$em = $this->getEntityManager();
$repo = $this->getEntityManager()->getRepository('Gedmo\Translatable\Entity\Translation');
$repo->translate($product, 'name', 'ru', 'assad');
$em->persist($product);
$em->flush();

Results into 'SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "id" violates not-null constraint'

Stack trace;
#0 vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(754): PDOStatement->execute(Array)
#1 vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(525): Doctrine\DBAL\Connection->executeUpdate('INSERT INTO ext...', Array, Array)
#2 vendor/gedmo/doctrine-extensions/lib/Gedmo/Translatable/Mapping/Event/Adapter/ORM.php(199): Doctrine\DBAL\Connection->insert('ext_translation...', Array)
#3 vendor/gedmo/doctrine-extensions/lib/Gedmo/Translatable/TranslatableListener.php(372): Gedmo\Translatable\Mapping\Event\Adapter\ORM->insertTranslationRecord(Object(Gedmo\Translatable\Entity\Translation))
#4 vendor/symfony/symfony/src/Symfony/Bridge/Doctrine/ContainerAwareEventManager.php(61): Gedmo\Translatable\TranslatableListener->postPersist(Object(Doctrine\ORM\Event\LifecycleEventArgs))

### LINKED ISSUE #480: Soft-deleting a Loggable entity does not produce a LogEntry
Using DoctrineExtensions 2.3.0-dev (updated today via composer), I attached the SoftDeleteable filter to an entity that already had the Loggable annotation. Logging worked as expected previously. While I realize that Loggable would no longer make a "remove" LogEntry with this setup, I did expect that an "update" LogEntry would be created with the data field containing something like {"deletedAt": "2012-10-15 14:56:18"}. Unfortunately, no LogEntry is created. How can I make these two work together?

If it helps, I'm using Symfony 2.1.3-dev, and StofDoctrineExtensionsBundle 1.1.x-dev.

### LINKED ISSUE #509: [translatable] Set @Gedmo\Locale field on hydration
It would be very useful to set the field marked `@Gedmo\Locale` when hydrating objects from a query. Example:

``` php
class Document implements Translatable
{
    /**
     * @Gedmo\Locale
     */
    private $locale;

    /* ... */
}

$query = $em->createQuery('SELECT d FROM Document');
$query->setHint(Query::HINT_CUSTOM_OUTPUT_WALKER, 'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker')
      ->setHint(TranslatableListener::HINT_TRANSLATABLE_LOCALE, 'nl');

$documents = $query->getResults();
```

Currently, the `locale` field will be `NULL` in the hydrated objects. It would be very useful if the Translatable behaviour sets this field to the locale that was used for the query. Whether the locale was given in a hint, or using the `TranslatableListerer`.

### LINKED ISSUE #554: Nested Set and Soft delete combination problem

### LINKED ISSUE #739: [SoftDeleteable] How to undelete? (and make it play nicely with Tree)

### LINKED ISSUE #754: [Translatable]  vs [SoftDeleteable]

### LINKED ISSUE #763: [Discussion] Translatable multiple fallbacks
Hi, I wanted to open a discussion regarding the translatable extension.

I want to use translation fallback, but without a default locale.

For example, according to the user's language it would fallback as such:
- `en` -> fallback to `fr`
- `fr` -> fallback to `en`
- `de` -> fallback to `en`

Ideally it would even fallback to several locales (for example `en` would fallback to `de`, or `fr`).

So the fallback is not always the same, since there is no concept of "default" language.

---

If I am not mistaken, the main thing I need is to improve the `TranslationWalker`.

I need to join the translation table twice:
- a join to get the target translation
- a join to get the fallback translation

and then: `COALESCE('translation', 'translationFallback')`

---

Have you ever thought of doing that? Started any work or anything?

I wanted to see with you first.
I started looking around the code but the Doctrine API regarding the tree walkers is quite obscure IMO (a lot of things in arrays, not objects), so that wasn't really productive.

I you have any idea/remarks thanks!

### LINKED ISSUE #821: [References] Adding the ReferencesListener breaks all other Doctrine extensions.
Hi - so I've been using Tree, Blameable and Timestampable with no problems at all.  I've recently had a need for the ORM/ODM cross-relational functionality, so added in the ReferencesListener, and although it works fine none of the other extensions work at the same time.

Blame and Timestamp both just insert null values now, and Tree gives me the error "Tree listener was not found on your entity manager, it must be hooked into the event manager".

I've done a var_dump in the Gedmo\Tree\Entity\Repository\AbstractTreeRepository code and it seems that whenever this line (from my services.yml) is called

```
        - [ registerManager, [ 'entity', "@doctrine.orm.entity_manager" ] ]
```

then $em->getListeners() returns an empty array.  No exceptions are thrown, just the other extensions fail to work.  If I comment out that one line, the situation is reversed (Everything works apart from references).

I've tried with and without Stof, nothing changes.  I've tried injecting the entity managers in the constructor, that doesn't work either.  I've also tried with just one listener where I pass in 2 entity managers.  I've included the relevant bit of services.yml below.

Thanks,
Adam

```
# Gedmo
gedmo.listener.reference.mysql:
    class: Gedmo\References\ReferencesListener
    tags:
        - { name: doctrine.event_subscriber, connection: game }
    calls:
        - [ setAnnotationReader, [ "@annotation_reader" ] ]
        - [ registerManager, [ 'document', "@doctrine_mongodb.odm.document_manager" ] ]

gedmo.listener.reference.mongodb:
     class: Gedmo\References\ReferencesListener
     tags:
        - { name: doctrine_mongodb.odm.event_subscriber }
     calls:
       - [ setAnnotationReader, [ "@annotation_reader" ] ]
        - [ registerManager, [ 'entity', "@doctrine.orm.entity_manager" ] ]

gedmo.listener.tree:
    class: Gedmo\Tree\TreeListener
    tags:
        - { name: doctrine.event_subscriber, connection: game }
    calls:
        - [ setAnnotationReader, [ @annotation_reader ] ]

gedmo.listener.timestampable:
    class: Gedmo\Timestampable\TimestampableListener
    tags:
        - { name: doctrine.event_subscriber, connection: game }
    calls:
        - [ setAnnotationReader, [ @annotation_reader ] ]

gedmo.listener.blameable:
    class: Gedmo\Blameable\BlameableListener
    tags:
        - { name: doctrine.event_subscriber, connection: game }
    calls:
        - [ setAnnotationReader, [ @annotation_reader ] ]
```

### LINKED ISSUE #836: Determine Slug before flush
Is there a way to determine what the slug will be before the Entity is flushed?

Also, if there is a way, would this take into account duplicates correctly. Eg, if slug was set using the 'name' field and I persisted 2 entities with the name 'Joe', can I get the correct slugs as 'joe', and 'joe-1'?

### LINKED ISSUE #850: [Tree] Allow tree root to be a separate entity
This would be a nice feature if we could define TreeRoot as for instance, ManyToOne relationship. Consider such a use case:
Entity\Navigation
Entity\Page (nested set).
Now each Entity\Navigation has it's own nested set of Entity\Page. 
Currently two solutions are possible: fake root node for each Entity\Navigation or InheritanceMapping, eliminating Entity\Navigation.

### LINKED ISSUE #859: [Sluggable] keeps generating unique slugs, unique=false not working!
I have implemented the Sluggable into a Nested Tree.  My entity partially looks like this:

```
/**     
 * @ORM\Column(type="string", length=255)
 */
private $title;

/**
 * @Gedmo\Slug(fields={"title"})
 * @ORM\Column(type="string", length=255, unique=false)
 */
 private $slug;
```

When i store the same title (this is an example) multiple times i get slug strings like this: "this-is-an-example-1", "this-is-an-example-2". 

I thought the unique=false setting should prevent this from happening. 
Any thoughts?

l3pp4rd:
# Refactoring for v3.0.0

Realease is planned next year **2014** will support:
- ORM **2.3.0** and higher versions
- ODM MongoDB **1.0.0** stable and higher versions.

After 3.0.0 is released, it is planned to avoid any further BC brakes. 

General backward incompatible changes for **v3.0.0**
## General

closes #381
closes #436 
- [x] for extension metadata, support relative class names based on current file namespace like ORM does.
- [x] refactor tests to support testing multiple database types, make general cleanup
- [x] get rid of Wrappers, makes static references, might cause memory usage issues
- [x] refactor mapping drivers based on 2.3 ORM features and improve support for inheritance
- [x] remove unused interfaces for each extension
- [x] get rid of Event adapters, use more general approach
- [ ] ensure YAML mapping compatibility for all extensions, ref #381 
- [x] get rid of **useObjectClass** in each metadata config, use rootEntityName or rootDocumentName instead (still need it anyway for inheritance, example slug definition (root class) may take fields from all inherited classes as sluggable ones, it is important for queries being issued)
- [ ] do not guess location of metadata and drivers for each extension, make it explicit listener responsibillity
- [ ] maybe provide a single class which would hook all drivers and listeners to a given **ObjectManager**
- [ ] update documentation, ref #436 
- [ ] make a command to list active listeners and watched entities
## Softdeleteable

closes #739
closes #554
closes #754

Handles softdeleted slug uniqueness checks implicitly. Here is some minnor BC break if `addManagedFilter($name)` was used explicitly for sluggable.
- [x] make filter ignorance available at base listener class, to share common behavior between listeners
- [x] implicitly manage softdeleteable filter for sluggable listener without any user intervention.
- [x] update tree **lft** **rgt** values, when nodes were softdeleted
- [x] handle translations related to an entity, do not remove them
- [ ] update documentation
## Translatable

closes #100 
closes #267
closes #282
closes #284 
closes #371
closes #379
closes #391 
closes #404
closes #415 
closes #417
closes #430 
closes #448 
closes #479 
closes #509 
closes #763 

Generally drop support for non ORM standard way to manage translations. This change will have more features and performance related improvements.
- [x] have translation entities related for each entity as currently personal translations do
- [x] translations should contain all translatable fields
- [x] other extensions should be able to apply directly on translation (sluggable as an example)
- [x] query hint for translation should do only one join (inner or left), should not require cast for columns
- [x] no more default locale, all locales are persisted as translations, translatable entity fields serve only as proxy and representation
- [x] support multiple locale fallbacks, for onload and translation query through hint
- [x] require hints for translatable queries, there were issues when values from listener were used, query cache was not working.
- [x] adapt ODM MongoDB for same updated behavior
- [x] create a command to update, generate translations
- [x] test generator command with orm different mapping types
- [ ] test generator command with odm mongodb different mapping types
- [ ] test generator command with inheritance mapping
- [ ] consider adding some migration commands, too 
- [ ] maybe put Translator inside Translatable if it was used or may be useful
- [ ] update documentation, ref #391
## Tree

closes #464 
closes #850
- [ ] support repository to construct children hierarchy of entities and as array #464.
- [ ] support tree root as any single valued non nullable association
## Sortable

closes #212 
closes #370 
- [x] add MongoDB ODM support
- [x] do not allow nullable sortable position
- [x] support multiple sortable positions, like sluggable does, refactor it to Gedmo\Sortable(groups={"field", "associationField"})
- [ ] test and support mapping drivers
- [ ] update documentation
## References

closes #821 
- [x] use manager registry to register different managers #821 
## Sluggable

closes #859 
closes #836
- [ ] find a better way to replace handlers
- [x] remove unique setting in slug annotation, reuse one for column #859 (rejected change)
- [x] support slugging identifier fields (sets slug on persist) #836 
## Loggable

closes #480
- [ ] play nice with softdeleteable and track removal #480 
### Concerns regarding translatable
- https://groups.google.com/forum/?fromgroups#!topic/doctrine-user/bXRzMzRhlFQ
- https://gist.github.com/sandermarechal/5661187#file-post-php

Ongoing work examples can be found in [tests](https://github.com/l3pp4rd/DoctrineExtensions/tree/wip-v2.4.0/tests/Translatable)


gonzalovilaseca:
I have started to work on adding phpcr odm to 'References' (and using registries instead of managers) on master branch, but I wonder if it would be a waste of time and It might be better to do it for version 3?
I will also need @stof bundle for symfony, so if the new versions of both bundles are going to be released within the following weeks I guess it's better to do it for version 3, otherwise I will do it for current version first.

l3pp4rd:
no, this branch most probably won't be merged. unless someone would take over the development, currently I'm working mostly with golang..

gonzalovilaseca:
ok good to know.
I'm learning golang too, any interesting repos that might be worth contributing to?

JHGitty:
Any updates? Has the development be discontinued?

l3pp4rd:
Yes, it is too much of work. current 2.x.x versions will be maintained as is for stability. My work profile has changed to java and golang, I do not use much of php, which means that I do not have much interest to move extensions to the next level, maybe @Atlantic18 is going to move it further some day.

JHGitty:
But there are some good improvements in this work.

For example:
```
 update tree lft rgt values, when nodes were softdeleted
```

Could you add these improvements to 2.x.x or create a 3.x.x branch with the current improvements?

l3pp4rd:
If that is important to you, could you care to open a pull request with the copy of the related code portion to 2.x. I just do not have time and interest.

AkenRoberts:
Since much of this progress currently exists in the 2.4 branch, and no further work is being done on this PR and 3.0 release, I'm going to close this.

Trying my best to reinvigorate this package a bit for 2019, and get it more stable and feature rich.

Thank you for all of your contributions @l3pp4rd!

PR #779, opened 2013-07-15
[WIP] [Uploadable] Allow build path relative to default using "pathMethod" property

Koc:
Quote from the documentation:

> For example, you can set specific paths for specific entities, or you can get the path from other sources (like a framework configuration) instead of hardcoding it in the entity.

But this is not possible get path from fw config yet. I've added basic support of this:

``` php

$listener->setDefaultPath('/my/path');

/**
 * @ORM\Entity
 * @Gedmo\Uploadable(pathMethod="getPath")
 */
class Photo
{
    public function getPath($defaultPath)
    {
        return $defaultPath.'/photos/'.mt_rand(10, 99).'/'.mt_rand(10, 99);
    }
}
```

**Let's discuss this before merging. How it could be better (put array with file information for adding possibility generate path depends on generated filename)?**

### DIFF
--- a/doc/uploadable.md
+++ b/doc/uploadable.md
@@ -50,7 +50,8 @@ on how to setup and use the extensions in most optimized way.
     * **pathMethod** - Similar to option "path", but this time it represents the name of a method on the entity that
     will return the path to which the files represented by this entity will be moved. This is useful in several cases.
     For example, you can set specific paths for specific entities, or you can get the path from other sources (like a
-    framework configuration) instead of hardcoding it in the entity. Default: ""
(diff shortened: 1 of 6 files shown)


webdevilopers:
Sorry for commenting on this old merged issue but has the functionality supposed by @Koc been implemented too?
https://github.com/Atlantic18/DoctrineExtensions/pull/779#issuecomment-21062937
-> `what about support parameters injection like @Uploadable(path="%upload_dir%/photos")`

I would like to use the `default_file_path` option from my config and add the ID of my entity as subfolder.

PR #838, opened 2013-09-18
Subclasses of valid types for Timestampables are considered too

josch1710:


josch1710:
Any progress on that?

jtreminio:
This doesn't look like it's going to be merged. Since there's no real way to set the `$validTypes` from outside the class, there's only a few other options.

* Use reflection to set the values you need at runtime, or
* Use Composer to point the class to one you control:

```
    "autoload": {
        "classmap": [
            "src/YourBundle/Vendor/GedmoTimestampableMappingDriverAnnotation.php"
        ]
    },```

PR #864, opened 2013-10-17
[Translatable] cast the id column to character varying in WHERE clause

greg0ire:
The foreign key column appears to use the character varying type.
Comparing these types does not work with postgres. When using the translation walker, it results in an error message saying this :

> SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer
> LINE 1: ...aBundle\Entity\FilmEditorial' AND t1_.foreign_key = c0_.id L...
> ^
> HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts. 

I solved this problem by casting the id column to character varying


TriAnMan:
@greg0ire  you may find my workaround for this problem useful. See #1512

greg0ire:
Well, thanks! Better late than never ;)

PR #871, opened 2013-10-22
Makes index usable for non-varchar foreign keys

ansata2001:
If field foreign_key varchar(64) references integer PK none of the indexes can be used (at least in MySQL though). Converting any foreign PK to varchar-compatible type solves the problem. Performance hit is much lower than if the index is not used at all.


TriAnMan:
@ansata2001 you may find my workaround for this problem useful. See #1512

PR #908, opened 2013-11-20
[Sortable] Test reproducing a bug: moving an entity to the last position

mnapoli:
Moving an object to the last position using the `-1` position doesn't work.

The new position is not the last one, but the last one + 1, which result in a gap:
- 0
- 1
- 3 (should be 2)

This PR contains a test reproducing the problem.

I tried figuring out the source of the problem, I think it's that when moving to a negative number, the new position is not the correct one:
- A (0)
- B (1)

`A.setPosition(-1)`

In SortableListener: `$newPosition += $this->maxPositions[$hash] + 2; // position == -1 => append at end of list`
- A.position = -1 + MAX + 2 = 2 (should be 1)
- B.position = 0 (correct)

### DIFF
--- a/tests/Gedmo/Sortable/SortableTest.php
+++ b/tests/Gedmo/Sortable/SortableTest.php
@@ -542,6 +542,35 @@ class SortableTest extends BaseTestCaseORM
         $this->assertEquals(4, $nodes[4]->getPosition());
     }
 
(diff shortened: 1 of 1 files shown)


AkenRoberts:

Thank you!

PR #1018, opened 2014-03-19
Bug fix for empty or "0" translated values

TonyBogdanov:
It is very bad to rely on the translation value to check if the field has been translated. The check would never return true if the value in the translation entity is either empty ("") or zero.

A better way would be using a separate flag just for the check.


AkenRoberts:

Thank you!

PR #1047, opened 2014-04-16
issue #945 fix

### LINKED ISSUE #945: [Sluggable] unique_base option and getSimilarSlugs() function
This issue is a fix-request to solve a bug when using unique_base option.

The bug : when generating slug when other similar slugs exists, the max count is used, but the unique_base is ignored.

Explanations : 
There are 2 unique bases : 
User1 (U1) and User2 (U2), each user owns one or many Page entities.

U1 (page, page-1, page-2, page-3) ;
U2 (page, page-1) ;

When U2 persists a new Page, the page slug should be « page-2 » (the next available valid slug), but the result is « page-4 » (the next available slug without filtering on unique_base). So the problem happens in getSimilarSlugs() function in ORM.php at line 42.

I've found this fork from ajanssen (https://github.com/ajanssen/DoctrineExtensions-1/commit/772f3f97a88a89a799dc122ffc6d2c18cbd2c95a) , this fix would be nice to be added in your DoctrineExtensions.

This fix consist to add a join clause when checking similar slugs when unique_base is set, in order to ignore slugs that belongs to other base uniqueness criteria element.

Replacing this line

``` php
$qb->join($associationMappings[$config['unique_base']]['targetEntity'], 'unique_'.$config['unique_base']);
```

With these 3 lines

``` php
$qb->join('rec.'.$associationMappings[$config['unique_base']]['fieldName'], 'unique_'.$config['unique_base']);
$qb->andWhere('unique_'.$config['unique_base'] . ' = :unique_base');
$qb->setParameter(':unique_base', $ubase);
```

Is it possible to you to add this fix in your project ? This would be great, because I need, after a composer-update, to fix again the file ORM.php manually.

Regards

jpass:
fixed #945: [Sluggable] unique_base option and getSimilarSlugs() function

### DIFF
--- a/lib/Gedmo/Sluggable/Mapping/Event/Adapter/ORM.php
+++ b/lib/Gedmo/Sluggable/Mapping/Event/Adapter/ORM.php
@@ -40,7 +40,9 @@ class ORM extends BaseAdapterORM implements SluggableAdapter
                 $qb->setParameter(':unique_base', $ubase);
             } elseif (array_key_exists($config['unique_base'], $wrapped->getMetadata()->getAssociationMappings())){
                 $associationMappings = $wrapped->getMetadata()->getAssociationMappings();
-                $qb->join($associationMappings[$config['unique_base']]['targetEntity'], 'unique_'.$config['unique_base']);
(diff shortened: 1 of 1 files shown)


stephanvierkant:
See https://github.com/Atlantic18/DoctrineExtensions/issues/1058#event-352807829

PR #1075, opened 2014-05-12
Fix - Add control - if newValue is object

### LINKED ISSUE #1074: [Doctrine\ORM\ORMInvalidArgumentException]  Persist - verify if $newValue is object
Since last commit, we have the following error when gedmo blameable is fired : 

```
[Doctrine\ORM\ORMInvalidArgumentException]                                       
  EntityManager#persist() expects parameter 1 to be an entity object, NULL given. 
```

Fix : Just add a check in the new added condition "if $newValue is object"

NG-iloWeb:
[Doctrine\ORM\ORMInvalidArgumentException]  
  EntityManager#persist() expects parameter 1 to be an entity object, NULL given.

Closes #1074

### DIFF
--- a/lib/Gedmo/Blameable/BlameableListener.php
+++ b/lib/Gedmo/Blameable/BlameableListener.php
@@ -82,7 +82,7 @@ class BlameableListener extends TimestampableListener
         $newValue = $this->getUserValue($meta, $field);
 
         //if blame is reference, persist object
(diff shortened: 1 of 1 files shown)


fungio:
+1

NG-iloWeb:
My apologies for the late reply

Would you have an example so we can create the test ?

Thks

l3pp4rd:
solved in master

PR #1126, opened 2014-07-16
Versioned class and inherited fileds for annotation configuration

chkris:
This annotation allows to define list of fields which will be versioned. By default all inherited fields and marked as `@Gedmo\Versioned` will be logged but sometimes we would like to specify that only few of them should be versioned.

Example:

``` php

class Person
{
    /**
     * @Gedmo\Versioned
     */
    protected $age;
}

/**
 * @Gedmo\VersionedClassAndInheritedFields(fieldList={"skill"})
 */
class Worker extends Person
{
    /**
     * @Gedmo\Versioned
     */
    protected $skill;
}
```

Now - because annotation VersionedClassAndInheritedFields has been used - versioned will be only this field. If this annotation will be not used than all fields from class Worker and inherited form Person (age, skill) will be versioned.


AkenRoberts:

Thank you!

PR #1131, opened 2014-08-01
Add support for a treeRoot with value NULL

vansante:
Previously you would get an SQL error when trying to insert a node into
a tree where treeRoot was NULL. Now it handles it fine.


l3pp4rd:
fixed in recent commit

PR #1145, opened 2014-08-26
Do not update field if blame is null

kadryjanek:
This one can produce exception when blame is unable to determine:

EntityManager#persist() expects parameter 1 to be an entity object, NULL given.
Scenario: updating entity through CRON tasks.


l3pp4rd:
solved by other commit

Issue #1177, opened 2014-10-21
[Sluggable] should try a slug without numeric postfix first.


maximium:
Hi!
There is an issue(I believe) when you have entity with title 'Title 777' and slug 'title-777' in db then add new entity with title 'Title' and expect this new entity will have a slug 'title' but it will have a slug 'title-1'.
I hope you understand my English :)

cmfcmf:
This is fixed in #1316 and can be closed.

PR #1178, opened 2014-10-21
#1177 [Sluggable] should try a slug without numeric postfix first.

maximium:
Fix for issue #1177


cmfcmf:
This is fixed in #1316 and can be closed.

PR #1191, opened 2014-11-16
Allow storing NULL as blameable reference (for the case when user is not logged in)

geekdevs:


tulanowski:
I encountered the same problem. I totally agree with geekdevs and luishdez. 

Moreover in my opinion ecf70f0 (which caused this problem) is wrong, because listener **should not** persist referenced user entity. Even if it would be needed in some cases, it is always possible to set cascade={"persist") for given relation. Commit ecf70f0 is wrong way to fix this failing test.

In my opinion either ecf70f0 should be reverted or (at least) this pull request should be merged. Until this is fixed I'm forced to use 2.3.9 version.

ajanssen:
+10

NielsJanssen:
I'm also running in to this problem. I agree with the fix proposed by @tulanowski, a persist in this location can cause unexpected results and should be handled via cascades or by manually persisting.

arlin2050:
+1

Same case for me, persisting an object from CLI breaks since v2.3.10.

PR #1197, opened 2014-11-26
MySql Numeric casting quick fix

birko:
a Quick fix for issue 
https://github.com/Atlantic18/DoctrineExtensions/issues/1196

where MySQL Numeric type should be casted as Decimal, because of 
http://dev.mysql.com/doc/refman/5.0/en/cast-functions.html#function_cast

### DIFF
--- a/lib/Gedmo/Translatable/Query/TreeWalker/TranslationWalker.php
+++ b/lib/Gedmo/Translatable/Query/TreeWalker/TranslationWalker.php
@@ -339,7 +339,13 @@ class TranslationWalker extends SqlWalker
                     (!($this->platform instanceof MySqlPlatform) &&
                     !in_array($fieldMapping["type"], array("datetime", "datetimetz", "date", "time")))) {
                     $type = Type::getType($fieldMapping["type"]);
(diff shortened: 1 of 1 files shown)


samgavinio:
Ping please.

l3pp4rd:
@birko can you rebase it?

birko:
@l3pp4rd ok I did a rebase. hope I did it correctly since it is first time I rebased a branch after such long time

stof:
I think you did the wrong rebase given the number of commits in this PR

birko:
@stof Fixed it with help of a friend @krivulcik

stof:
It would be great to have a test preventing regression on this

birko:
will add one

l3pp4rd:
thanks, if you add a test, submit another PR since it may take some time

birko:
yeah. kinda. is there some connection setting I could use to mock Mysql. since it is platform specific test and I saw only sqlite connection in the test suite

l3pp4rd:
there is an ormtestcase where it creates a driver with some mocks, it could have a separate function which gets a driver with mocked platform..

PR #1204, opened 2014-12-15
Fix the id strategy for closure and translation entities for PostgreSQL and Oracle

stof:
The Closure strategy and the translatable listener are inserting the data in SQL without providing the identifier, assuming it is generated by the database. However, the default id strategy for Doctrine is ``AUTO``, which chooses the strategy depending on the platform. When the sequence strategy is preferred (PostgreSQL and Oracle), these listeners were broken. Forcing the strategy to ``IDENTITY`` rather than ``AUTO`` fixes it.

### DIFF
--- a/lib/Gedmo/Translatable/Entity/MappedSuperclass/AbstractPersonalTranslation.php
+++ b/lib/Gedmo/Translatable/Entity/MappedSuperclass/AbstractPersonalTranslation.php
@@ -16,7 +16,7 @@ abstract class AbstractPersonalTranslation
      *
      * @ORM\Column(type="integer")
      * @ORM\Id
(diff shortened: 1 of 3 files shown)


Legenyes:
This don't work for Oracle.

Oracle require AUTO or SEQUENCE

Can you do rollback ?

JPernerstorfer:
@Legenyes is right, I had to edit the AbstractTranslation class and leave it like this:

    /**
     * @var integer $id
     *
     * @ORM\Column(type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue
     */

This way it works, but editing vendors... not my jam.

PR #1212, opened 2015-01-06
added support for embedded documents

data219:
Support for embedded documents in mongodb-odm

### DIFF
--- a/lib/Gedmo/Loggable/Document/Repository/LogEntryRepository.php
+++ b/lib/Gedmo/Loggable/Document/Repository/LogEntryRepository.php
@@ -46,7 +46,6 @@ class LogEntryRepository extends DocumentRepository
         if ($result instanceof Cursor) {
             $result = $result->toArray();
         }
-
         return $result;
     }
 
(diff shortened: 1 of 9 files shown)


simensen:
I have a need for this. Is there anything missing from this PR to that would keep it from being merged?

simensen:
@l3pp4rd Thanks for the merge! We've been trying some of these patches out over the last few days and found #1069 to be a better fit. It looks to be more complete and supports EmbedMany. Any chance that you would be able to work out the differences or would you consider reverting this one in favor of the other? You beat me to the punch on this one; after testing all day yesterday, I was going to come here and report my findings.

l3pp4rd:
hm, could you diff it? if that is important, I do not have much time to dive in it. I promise to merge it quicker than this one :)

simensen:
@l3pp4rd it might be above my knowledge to make sense of the diff between the two. I could try to revert this commit and merge the other into a new PR?

Alternately, does GitHub present you with a "revert" button on this PR? If so, maybe you could revert it that way and then merge the other? :)

PR #1213, opened 2015-01-06
Make members protected instead of private

FranzBruckner:
Please make members protected instead of private so that the class can
be easily extended.


l3pp4rd:
I think you wanted to push less commits than that?

Issue #1209, opened 2014-12-31
[Sortable] Positions are updated on entity remove, even if removing failed


naitsirch:
The *Sortable* extension updates the position of entities if an entity is scheduled to be removed. But if the removing fails (e.g. on foreign key constraint violence) the positions still get updated, what leads to wrong positioning.

### Example

**Table A**

ID | Name | Position [sortable]
--- | ------- | ----------------------
1 | aaa | 1
2 | bbb | 2
3 | ccc | 3

**Table B**

FK-to-table-A [onDelete `restrict`] | Name 
---------------------------------------- | --------
2 | I like bbb

Now, we tell doctrine to remove entity 2 from **Table A** with
```php
/* @var $em EntityManager */
try {
    $em->remove($entity2);
    $em->flush();
} catch (ForeignKeyConstraintViolationException $e) {
    // ...
}
```
Because of the foreign key contraint violation the entity won't be removed from database. But at the end of this action **Table A** will look like that:

ID | Name | Position [sortable]
--- | ------- | ----------------------
1   | aaa   | 1
2   | bbb   | ***1***
3   | ccc    | ***2***

So the position has been decreased.

naitsirch:
After investigeting into this issue I found out, that the decreement of the position field happens before the db transaction starts, which means that the rollback on exception does not roll back the changes of the positions.
Maybe this is the reason. I'll try to look deeper into this.

naitsirch:
The problem is, that the position relocations are processed in `onFlush` event. At this time doctrine has not started the transaction. So that the rollback after the exception does not roll back the changes of the position.

It seems to work, when the call of `processRelocations()` is moved to other events. I'll test this and create a PR if no tests will break.

leonex-cstoller:
After investigeting into this issue I found out, that the decreement of the position field happens before the db transaction starts, which means that the rollback on exception does not roll back the changes of the positions.

The problem is, that the position relocations are processed in `onFlush` event. At this time doctrine has not started the transaction.

It seems to work, when the call of `processRelocations()` is moved to other events. I'll test this and create a PR if no tests will break.

leonex-cstoller:
@l3pp4rd this issue could be closed, after the PR is merged ;-)

naitsirch:
That's great. Thanks!

PR #1214, opened 2015-01-09
Persistance of position relocations moved to other events to fix issue #1209

leonex-cstoller:
This PR fixes issue #1209. Please refer this issue to understand the background of this patch.

* I have added an unit test for issue #1209 
* Moved the persistance of position relocations from `onFlush` event, which is called before database transaction started, to the first call of `postPersist`, `preUpdate` or `preRemove`. So that the update query is executed within the transaction and if errors happen during queries the position changes will be rolled back.
* The synchronization of the objects has been separated from the database persistance, so that it will happen only if all queries were successfully executed (in `postFlush`).

The only thing I don't like with this solution is, that it will only work with entities that have identifiers. I am not sure if this is a problem.
I had to use these because the INSERTs are executed before the UPDATE query for the relocation of the positions. And that will manipulate the positions of inserted records. But I cannot use another event in this case.

Doctrine lacks an event which is fired between the beginning of a transaction and the inserts :-( This would have made the patch much simpler.

### DIFF
--- a/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
+++ b/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
@@ -67,6 +67,30 @@ final class ORM extends BaseAdapterORM implements SortableAdapter
             }
         }
 
+        // add excludes
+        if (!empty($delta['exclude'])) {
+            $meta = $this->getObjectManager()->getClassMetadata($relocation['name']);
+            if (count($meta->identifier) == 1) {
+                // if we only have one identifier, we can use IN syntax, for better performance
+                $params['excluded'] = array();
+                foreach ($delta['exclude'] as $entity) {
+                    $params['excluded'][] = $meta->getFieldValue($entity, $meta->identifier[0]);
+                }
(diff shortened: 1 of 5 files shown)


naitsirch:
Testing on a live system showed that the `shouldSyncPositionAfterDelete` did not test if the positions have been persisted correctly. So I have extended the test in 79544ff57c3243843f7969de83eb2f6fa23e9276.

Further I had to move the call to relocation persistance from `preRemove` to the `postRemove` event, because preRemove is executed right after calling `$em->remove($entity)` and not in `flush`.

I hope it's okay, that I have done that in multiple commits.

leonex-cstoller:
Testing on a live system showed that the `shouldSyncPositionAfterDelete` did not test if the positions have been persisted correctly. So I have extended the test in 79544ff57c3243843f7969de83eb2f6fa23e9276.

Further I had to move the call to relocation persistance from `preRemove` to the `postRemove` event, because preRemove is executed right after calling `$em->remove($entity)` and not in `flush`.

I hope it's okay, that I have done that in multiple commits.

l3pp4rd:
thanks, that was very useful

PR #1228, opened 2015-01-23
fix [Semantical Error] The class Entity is not annotated with @Annotatio...

vevtik:
...n. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of Entity. If it is indeed no annotation, then you need to add @IgnoreAnnotation(Entity) to the _class_ doc comment of class Gedmo\Translatable\Entity\Translation.

### DIFF
--- a/lib/Gedmo/Translatable/Entity/Translation.php
+++ b/lib/Gedmo/Translatable/Entity/Translation.php
@@ -2,24 +2,21 @@
 
 namespace Gedmo\Translatable\Entity;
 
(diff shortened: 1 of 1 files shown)


stephanvierkant:
Thanks for your PR and I'm sorry that there hasn't been any response by the maintainers.

I've closed this PR because it the branch conflicts that must be resolved. If you still want this PR to be merged, please resolve the conflicts and reopen this PR.

PR #1232, opened 2015-01-26
Fix XML tree mapping to be used with resolving target entities.

isometriks:
Similar to https://github.com/Atlantic18/DoctrineExtensions/pull/987 when using a target entity resolver it basically checks that (Acme\Model\Something)->isSubClassOf(Acme\Model\SomethingInterface) and was failing.


isometriks:
I guess this is replaced by https://github.com/Atlantic18/DoctrineExtensions/commit/5e19886a64ef7e3540bb362eacfdf7d60c5608e6, but still does not work correctly. Will make an issue.

PR #1236, opened 2015-02-03
Add enable trait

bocharsky-bw:
I think often need to enable/disable entities. What about to add a trait for it?


l3pp4rd:
it does not relate to extensions

PR #1176, opened 2014-10-19
Change translation request for postgresql platform fix #1168

xleliberty:
Hi, 
i've posted #1168 2 weeks ago, and here is a little fix for this

Use case: 
Postgresql 9.3 

Something like that: 

$qv = $this->getEntityManager()->createQuery($dqlValidities);
        $qv->setHint(
            \Doctrine\ORM\Query::HINT_CUSTOM_OUTPUT_WALKER,
            'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker'
        );
        $qv->setHint(TranslatableListener::HINT_TRANSLATABLE_LOCALE,$locale);

was failing because of type difference between foreign_key field and the generated query.

This pr solve that.

Thanks

Issue #1168, opened 2014-10-07
Translatable , TranslationWalker and Postgresql


xleliberty:
Hi, 

I've just tried to use TranslationWalker  on a join query , and i got a problem.
As long as the postgresql generated ext_translation table as a varchar foreignKey type, the generated query in translationWalker is invalid

In my cod, i've done something like that: 

        $qv = $this->getEntityManager()->createQuery($dqlValidities);
        $qv->setHint(
            \Doctrine\ORM\Query::HINT_CUSTOM_OUTPUT_WALKER,
            'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker'
        );
        $qv->setHint(TranslatableListener::HINT_TRANSLATABLE_LOCALE,$locale);

Here is the error i got: 

End of query: 
AND t9_.object_class = 'MyBundle\Entity\Productcategory' AND t9_.foreign_key = p7_.id LEFT JOIN ext_translations t10_ ON t10_.locale = 'en' AND t10_.field = 'shortdesc' AND t10_.object_class = 'MyBundle\Entity\Productcategory' AND t10_.foreign_key = p7_.id WHERE (p7_.id = ? AND v0_.visible = true) AND (v0_.deleted_at IS NULL) ORDER BY v0_.position ASC, v0_.value ASC, COALESCE(CAST(t1_.content AS VARCHAR(255)), v0_.name) ASC' with params [862]:

And the error is: 
SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer
LINE 1: ...ndle\Entity\Validitycategory' AND t1_.foreign_key = v0_.id L...
^
HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.   -

If anybody have a clue.

Thanks in advance

Xavier

edigu:
Any update on this?

edigu:
This PR can be closed, resolved & merged with #1249

vitalyiegorov:
POSTGRES
we have ```promo_side``` table and its translation table ```promo_slide_translation```, both have ```promo_slide_id:integer``` field, if i remove this ```::VARCHAR``` fix from 307 line in TranslationWalker than everything works fine.

With this fix our app is getting this error:
```
[...]

PR #1249, opened 2015-02-18
TranslationWalker patch for postgresql explicit type casting problem

edigu:
This PR aims to fix type casting problem in `TranslationWalker` which consists only with PostgreSQL platform when using composite primary key on a `Translatable` entity.

**Problem:** Since `foreignKey` property of the [Translation](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Translatable/Entity/MappedSuperclass/AbstractTranslation.php#L47) entity mapped as `string`, you can't use EQUAL operator (`=`) on this field if type of the both sides are not same. PostgreSQL is far more strict on type castings than MySQL. When you have a translatable entity with composite primary key, the final query which modified by TranslationWalker produces the error following:

> SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: character varying = integer LINE 1: ...lass = 'My\Entity' AND t1_.foreign_key = c0_.id O... ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.

Note: Also some doctrine coding standards applied (see line 307 for actual addition) and all tests are passing. I don't add any test since i can't figure out how to run platform specific unit tests for this library. Also this PR solves same issue introduced in #1168 and #1176 but runs a little more performant since we don't need to check the platform while iterating every field.

### DIFF
--- a/lib/Gedmo/Translatable/Query/TreeWalker/TranslationWalker.php
+++ b/lib/Gedmo/Translatable/Query/TreeWalker/TranslationWalker.php
@@ -11,6 +11,8 @@ use Doctrine\ORM\Query\AST\SelectStatement;
 use Doctrine\ORM\Query\Exec\SingleSelectExecutor;
 use Doctrine\ORM\Query\AST\RangeVariableDeclaration;
 use Doctrine\ORM\Query\AST\Join;
+use Doctrine\DBAL\Platforms\MySqlPlatform;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
 
 /**
  * The translation sql output walker makes it possible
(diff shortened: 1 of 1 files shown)


edigu:
Ping..

edigu:
ping..

l3pp4rd:
please do not change the CS rules. etc spacing rules are changed throughout the file

edigu:
ok, i just reverted back the CS related changes. @l3pp4rd

l3pp4rd:
thanks, by the way, concerning the foreign key you can change the mapping since it is a protected property when extending mapped superclass

PR #1259, opened 2015-02-26
Revert "Merge pull request #1212 from Preis24/embedded_documents_support"

simensen:
This reverts the merge of #1212 since it only supported EmbedOne. This is to prepare the way for merging #1069 instead as it supports both EmbedOne and EmbedMany.

PR #1260, opened 2015-02-26
EmbedOne and EmbedMany combo revert + merge

simensen:
@l3pp4rd I created this PR in order to help you avoid having to merge two other PR's. Merging this has the same effect as:

 * merging #1259 (will revert #1212)
 * merging #1069 (will enable support for EmbedMany in addition to EmbedOne)

Or you can merge each of those independently. I wanted to help make this as easy as possible for you without requiring me to spend several hours/days trying to resolve the differences in implementations between #1069 and #1212. :)


simensen:
There are two major issues I'm seeing:

 * master no longer builds successfully: [see travis](https://travis-ci.org/Atlantic18/DoctrineExtensions/builds/52189266)
 * the code from #1069 no longer runs correctly on master: [see travis](https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/52316255)

I am going to try and see if I can fix this.

l3pp4rd:
Yes indeed, this is strange, since I have tested it locally. It would be helpful if you are using ODM and could help with these issues.

simensen:
@l3pp4rd I'll see what I can do. Thanks for checking in on this.

l3pp4rd:
figured out that tests fail only on mongodb beta 12, beta 11 passes :)

simensen:
@l3pp4rd For what it is worth, I've discovered the problem. The memory leak "fix" in 4e9bbd54ebb22bb004b2ec71b1b0ce3ed5b992d0 is breaking #1069. I'm currently trying to figure out how I can work around this. I'll add the changes here.

Current ideas:

 * **Keep the cache of objects in `LoggableListener` and have them only be purged on clear().** The downside here is that memory may leak until clear is called.
 * **Rewrite the patch so that it does not rely on certain aspects of *Wrapper.** This is probably going to take the most time and I'm not sure I will be able to do this. :-/
 * **Find out if I can track whether an object has been initialized outside of the Wrapper itself.** If I can do this I *think* we can avoid the state problems we are seeing; right now, this `EmbedOne`/`EmbedMany` solution is doing something that seems to require the wrappers to be somewhat persistent.

Are any of these options completely unacceptable, @l3pp4rd? Do you have a preference for one solution over the others?

l3pp4rd:
I think in long term the second option is the way to go, it should not depend on those wrappers, it was a mistake in general.

simensen:
@l3pp4rd OK, I'll see what I can do. Do you have an idea what I should do instead? :) Any direction or suggestions at all would be helpful.

Thanks!

l3pp4rd:
Well, the test suite passes again

simensen:
I've discovered that #1069 fails due to the fixed memory leak in 4e9bbd54ebb22bb004b2ec71b1b0ce3ed5b992d0. It would appear that this implementation was relying on some state that was being kept in the wrapped objects. I wasn't able to track down where or how it was using state but by always returning a new wrapped instance this implementation no longer works.

I then tried to implement embedMany based on #1212. It was resulting in a *lot* of refactoring and was requiring making changes to more parts of the code (some of the stuff in the common Mapping namespace, for example) and ultimately I decided that I wasn't sure this was the right way for this to be handled anyway. I abandoned my work on that branch but may come back to it someday.

One of the biggest issues I was facing was the fact that it wasn't clear to me how collections should be handled, especially on revert. Is how the deltas for collections are to be handled documented anywhere? How it works now (or at least in that one branch that no longer works) seemed very complicated.

l3pp4rd:
I think orm handles collections in persisters and in higher level unit of work for cascades and such. But I think it is troublesome to do something, unless you really need it at work, then you can come back to it. Cheers, thanks for looking into it, I know it is quite complex

PR #1262, opened 2015-03-04
Yaml mapping for references

gonzalovilaseca:
This adds yaml mapping to References extension, if it's ok I will update doc also, the yaml mapping would be like this:
```
gedmo:
    reference:   
        image:        
             reference: referenceOne
             type: document
             class: Document\Product
             inversedBy: stockItems
             identifier: productId
```
Being a relationship it might be worth validating that if it's the owning side it has ``inversedBy`` and ``              identifier`` defined? And if it's the inverse side it should have just ``mappedBy`` defined?

### DIFF
--- a/lib/Gedmo/References/Mapping/Driver/Yaml.php
+++ b/lib/Gedmo/References/Mapping/Driver/Yaml.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Gedmo\References\Mapping\Driver;
+
+use Gedmo\Mapping\Driver\File;
+use Gedmo\Mapping\Driver;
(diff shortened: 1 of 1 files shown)


l3pp4rd:
Hi, looks good, also validation would be appropriate as well.

diimpp:
Why does it uses `dcm.yml` filetype? Can't I use User.mongodb.yml?

l3pp4rd:
@diimpp the extension is only a default, for example symfony2 doctrine bundle overrides this.

diimpp:
:+1:

PR #1265, opened 2015-03-13
Use a LocaleResolver to select the default locale

sroze:
Move the default locale logic inside a `LocaleResolver` class. This allow to override the default locale choice based on entities.

A usage can be to use a different default locale based on an entity field, which will improve performances if we know that a given entity will be used mostly for a given language while some other entities will be for another language.

### DIFF
--- a/lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
+++ b/lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
@@ -82,8 +82,9 @@ class TranslationRepository extends EntityRepository
                 $transMeta->getReflectionProperty('field')->setValue($trans, $field);
                 $transMeta->getReflectionProperty('locale')->setValue($trans, $locale);
             }
-            if ($listener->getDefaultLocale() != $listener->getTranslatableLocale($entity, $meta) &&
-                $locale === $listener->getDefaultLocale()) {
+
+            $defaultLocale = $listener->getLocaleResolver()->getDefaultLocale($entity, $meta);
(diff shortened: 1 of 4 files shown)


sroze:
@l3pp4rd Could you have a look here ?

l3pp4rd:
I think this adds an overhead to the library. it should be a framework related service. You can have **translatable** listener as a public service and manage its default locale based on your specific use cases.

sroze:
@l3pp4rd do you mean a listener that will be called before yours for each `TranslationRepository::translate` and `TranslationListener::postLoad` ? I think that it may be broken easily if you're adding anything else related to the `defaultLocale`. 
This PR way we can add an extension point that won't be break easily and mainly reduce the complexity of developing such a feature...

l3pp4rd:
no @sroze I mean that translatable listener is hooked to entity manager event listeners. you can use the listener instance and call setDefaultLocale whenever you like. For example a symfony framework defines a [service](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/symfony2.md#doctrine-extension-listener-services) **gedmo.listener.translatable** in this case. At any point, you can get this service from container and set locale to it. I do not mean to add yet another listener. Use the actual one

sroze:
@l3pp4rd yeah, seems better ! Btw we need to call this `setDefaultLocale` method before each case which means introducing possible bugs in the future. How much this new `LocaleResolver` is an overhead here in your opinion ?

l3pp4rd:
Well, actually you may wrap it in something like `$service->flushWithDefaultLocale('de');` which would set the locale flush and reset it back. I do not see any advantage having this abstraction it only increases complexity, extensions are already quite complex.

sroze:
So closing that PR.

PR #1270, opened 2015-03-20
Clarified grammar in tree documentation.

maxolasersquad:
There where a number of grammar issues that made reading the documentation a little bit difficult that I cleaned up here.


l3pp4rd:
great, thanks ;)

PR #1276, opened 2015-03-30
Fix translation handling when modifying documents

alcaeus:
Note: this issue is a side-effect of doctrine/mongodb-odm#999 and was found while hunting for that bug.

When handling an update for a translated object, the insertion queue of the ```UnitOfWork``` is only checked for duplicates if the object uses personal translations. This allows for duplicate key exceptions in rare cases. I think it's best to check for duplicate translation objects not only for personal translations but also for the built-in translation class.

Note 2: I've only managed to check and reproduce this issue when using the ODM, I'd appreciate it if somebody could check this against Doctrine ORM.


l3pp4rd:
thanks ;)

PR #1278, opened 2015-04-01
Added test for changing position of a record and insert a new one in front of it in one step

naitsirch:
The added should work, but it currently fails. This is why I have marked it as "incomplete".
The unit test shows that there is a bug. I couldn't find a solution, yet, but I think adding the test should be done anyway.

The problem is that during processing of the updates, the maximum position is `1`, so that the position of $item3 is set to `1`. Later, when the relocations are applied, the position of $item3 is not changed to `2` again, because it is already marked as _changed_.


l3pp4rd:
thanks, hope someday we will solve this

PR #1279, opened 2015-04-01
Fixing issue in exclusion of new entities when positions are updated.

leonex-cstoller:
In some circumstances the update of positions is executed before all new entities are persisted. In this case the new entities' identifier is `NULL`, which makes the query `AND n.{$meta->identifier[0]} NOT IN (:excluded)` failing, because _xyz NOT IN (NULL)_ is always false.

This patch is addressing the code added in #1214.

### DIFF
--- a/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
+++ b/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
@@ -72,11 +72,16 @@ final class ORM extends BaseAdapterORM implements SortableAdapter
             $meta = $this->getObjectManager()->getClassMetadata($relocation['name']);
             if (count($meta->identifier) == 1) {
                 // if we only have one identifier, we can use IN syntax, for better performance
-                $params['excluded'] = array();
+                $excludedIds = array();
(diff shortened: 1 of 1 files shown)


l3pp4rd:
thanks ;)

francescolaffi:
thanks, got bitten by the bug addressed by this patch

we downgraded to 2.3.11 while waiting this to be included in a stable release

meanwhile since we were adding new elements with position 0 this resulted in having quite a bit of mangled positions sets (duplicates, negative positions, ...), here is a query similar to what we used to bring stuff to working order again, in case anyone else is in the same situation

````SQL
set @current_group := null, @pos := 0;
update my_table
set position_column = IF(group_column = @current_group, @pos := @pos+1, @pos := 0),
    group_column = (@current_group := group_column)
order by group_column ASC, position_column ASC;
````

Its for a simple situation in mysql, but it can be adapted to other situations/dbs.

Rvanlaak:
Does `Sortable` work on Symfony `2.6.6` for you? The sorting condition is not added to my query. Reverting to `2.6.5` does solve this.

PR #1282, opened 2015-04-10
Error in documentation for Closure Method

JDruery:
The example doesn't work as given, unless you add the following annotation on the Entity:
 * @ORM\Entity(repositoryClass="Gedmo\Tree\Entity\Repository\ClosureTreeRepository")


l3pp4rd:
the master repository is in Atlantic

PR #1286, opened 2015-04-11
feat(events): invoke entity listener preSoftDelete and postSoftDelete wh...

hafuta:
Added feature to invoke entity listener's preSoftDelete/postSoftDelete methods when soft-deleting an entity.

### DIFF
--- a/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
+++ b/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
@@ -2,6 +2,8 @@
 
 namespace Gedmo\SoftDeleteable;
 
+use Doctrine\ORM\Event\LifecycleEventArgs;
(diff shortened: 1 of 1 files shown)


l3pp4rd:
hm, that looks reasonable, I think it shoots the event in the same way as $evm used dirrectly, could you add a test case?

lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Gedmo\SoftDeleteable;
 
+use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\ListenersInvoker;
```
l3pp4rd: it will not work with ODM mongodb

l3pp4rd:
`$om->getEventManager();` does not work?

PR #1288, opened 2015-04-11
Update branch alias

cevou:
2.4.0 is released now. That's why the branch alias in the master branch should be changed to 2.4.x


stof:
thanks

PR #1292, opened 2015-04-17
allow updating both group and position in sortable listener (fix #997)

### LINKED ISSUE #997: [Sortable] Updating SortableGroup and SortablePosition duplicates the position
I have an ItemPosition Entity which has 2 properties 

- Category (references a category Entity) as SortableGroup
- Position as SortablePosition

CategoryA has about 30 items. Same for Category B.

When I change ItemPosition with Category : "Category B" and Position "4" to
"Category A" and Position "7"

I get some duplicated position

Expected behavior : 

```sql
update item_position set position = position - 1 where category = 'Category B' and position > 4

update item_position set position = position + 1 where category = 'Category A' and position >= 7
````

Got : 

```sql
update item_position set position = position - 1 where position >= 8 and position < 33   and category = 'Category B'

update item_position set position = position - 1 where position >= 5 and position < 8 and category = 'Category A' 
````

This means I get 2 items with position 4 in Category A.

I've tried to write a phpunit test to reproduce this but was unable to. I'll give it another try later today.

It may be important to mention, this may be related to some APC cache issue as I was only able to reproduce this on my prod environment where doctrine cache is enabled.

brunoroux:
As pointed out in #997 Updating both group and position on an entity provoked a duplicated position. This has been solved and a test has been added to cover this case.

### DIFF
--- a/lib/Gedmo/Sortable/SortableListener.php
+++ b/lib/Gedmo/Sortable/SortableListener.php
@@ -230,6 +230,7 @@ class SortableListener extends MappedEventSubscriber
         $uow = $em->getUnitOfWork();
 
         $changed = false;
+        $groupHasChanged = false;
         $changeSet = $ea->getObjectChangeSet($uow, $object);
 
         // Get groups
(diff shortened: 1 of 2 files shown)


l3pp4rd:
thanks ;)

Issue #1283, opened 2015-04-10
[Sluggable] Unique base is not working Yaml driver


cevou:
When using `unique_base` with the YAML driver there is an exception thrown while executing `!$this->isValidField($meta, $ubase)`. The unique base field is an association. Unfortunately, `!$meta->hasAssociation($ubase)` is never checked, because the exceptino is thrown before.
In the annotation driver, there is a different check for the field: `!$meta->hasField($slug->unique_base)`
If this is also used in the YAML driver, everything works fine. Can this be easily exchanged?

PR #1298, opened 2015-04-24
Sluggable: Use unique base with association and YAML/XML driver

cevou:
Fix for #1283


l3pp4rd:
thanks

patrickascher:
Hello, i have a simular problem with travis like you had here. 

PHP Fatal error:  Call to undefined method DoctrineTest::setDisallowChangesToGlobalState() in .../phpunit/src/Framework/TestSuite.php on line 716

do you still know what was the solution for this?

B.R
Patrick

PR #1306, opened 2015-05-08
Removing refs to timestampable in blameable docs

ryancastle:
I think the old references were a mistake, left over from a copy-paste?


l3pp4rd:
thanks ;)

Issue #1144, opened 2014-08-25
[Translatable] [ODM] objectClass field name and index key are not match


vkartaviy:
Look at class index [annotation](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib%2FGedmo%2FTranslatable%2FDocument%2FTranslation.php#L15) and objectClass field [annotation](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib%2FGedmo%2FTranslatable%2FDocument%2FMappedSuperclass%2FAbstractTranslation.php#L31). As you see field is stored as is and not match index.

I think it should be annotated like below to keep BC. 

``` php
/**
 * @var string $objectClass
 * 
 * @MongoODM\String(name="object_class")
 * @MongoODM\AlsoLoad("objectClass")
 */
protected $objectClass;
```

PR #1327, opened 2015-06-05
Fix mongo field name for translations to match index name

notrix:
Fix for issue #1144 
Field name in mongodb is camel cased and indexes are with underscore. It was not an issue until database reached 100k translations. Now mongo queries take ~10 secs.
New translations will be in correct case, but migration of old documents is required to take advantage of indexes (Fix has no BC breaks).

Also some tests fixed.


l3pp4rd:
the test suite was fixed with default repository, regarding the field name. use naming strategy if possible

Issue #1333, opened 2015-06-16
scheduleExtraUpdate does not exist in Doctrine\ODM\MongoDB\UnitOfWork


pomm0:
I get the error 
```
Attempted to call method "scheduleExtraUpdate" on class "Doctrine\ODM\MongoDB\UnitOfWork"

gedmo/doctrine-extensions/lib/Gedmo/Loggable/LoggableListener.php - line 141
```

The method "scheduleExtraUpdate" used in
* lib/Gedmo/Tree/Strategy/ODM/MongoDB/MaterializedPath.php
* lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
* lib/Gedmo/Loggable/LoggableListener.php
* lib/Gedmo/Tree/Strategy/AbstractMaterializedPath.php

but it does not exist in Doctrine\ODM\MongoDB\UnitOfWork

l3pp4rd:
what version of ODM?

pomm0:
latest dev-master

pomm0:
latest dev-master

pomm0:
it seems that it has been deleted there: https://github.com/doctrine/mongodb-odm/compare/f9d26e2...master#diff-ae2c9f4b58c612dce08a0f60ff683aecL1538

l3pp4rd:
well if you have time, you can try to submit a pull request with a fix. an alternative to schedule extra update.. not sure when I have time to look into it

malarzm:
Is this possible to replace `scheduleExtraUpdate` with something different here in Extensions? Justification for removal was "this is marked as internal and we're not using it for some time"

malarzm:
If I'm not mistaken the `scheduleExtraUpdate` was used to create reference between inserted item and its log, right? In ODM ids are assigned at the time of persist so there is no need to call `scheduleExtraUpdate` just to update the field, id could be passed to log entry upfront. Also [this](https://github.com/Atlantic18/DoctrineExtensions/blob/dedfbd7bd2ffb290ea19f3b1bbae52aa36c6008a/lib/Gedmo/Loggable/Document/MappedSuperclass/AbstractLogEntry.php#L36-L49) could be changed to use `@ReferenceOne()` without targetDocument specified which allows you to link anywhere, but I guess this is bigger change and is not feasible.

l3pp4rd:
well, if someone is suffering from this one, you can open a PR? not sure when I will be able to submit a fix..

[...]

PR #1335, opened 2015-06-17
Bugfix #1333: Using doctrine/mongodb-odm with version 1.0.0-BETA13

sgasser:
With version greater than 1.0.0-BETA13 the method "scheduleExtraUpdate" has been removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "doctrine/common": "~2.4"
     },
     "require-dev": {
-        "doctrine/mongodb-odm": ">=1.0.0-BETA13",
+        "doctrine/mongodb-odm": "1.0.0-BETA13",
         "doctrine/orm": ">=2.5.0",
         "doctrine/common": ">=2.5.0",
         "symfony/yaml": "~2.6",


stof:
This is not fixing the issue at all. Fixing compatibility with current versions of the ODM is the right fix, not changing the dev dependency to force an old version (which won't affect projects using the extensions)

l3pp4rd:
yes, we would need to check version and use appropriate method in the code in that case

sgasser:
It's true.

But it's correct using the latest master?

I'm using symfony with 

```
        "doctrine/mongodb-odm": "1.0.*@dev",
        "doctrine/mongodb-odm-bundle": "3.0.*@dev",
        "stof/doctrine-extensions-bundle": "1.1.*"
```

After composer update my project is broken...

malarzm:
@sgasser @stof @l3pp4rd please see https://github.com/Atlantic18/DoctrineExtensions/issues/1333#issuecomment-112743618

l3pp4rd:
this pull request has nothing to do with production version. extensions require only **doctrine-common** library. Everything else is optional, so is **doctrine/mongodb-odm** and require-dev won't change anything for your **composer update** at least it is not meant to.. you should close the PR

PR #1340, opened 2015-06-22
Nested Set Tree enhancements and fixes

pmishev:
A number of enhancements for the nested set tree:

- Added 'base' property for tree level annotation
- Verify and recover wrong levels in nested set
- Added getPathAsString() method to entity repository
- Added option to recover tree, without verifying first
- Added option to reorder only direct children in reorder() method
- Added option to specify siblings order while recovering tree
- Fixed ordering in childrenQueryBuilder()
- Added and fixed some wrong comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,3 +126,4 @@ And especially ones who create and maintain new extensions:
 - Kudryashov Konstantin [everzet](http://github.com/everzet)
 - David Buchmann [dbu](https://github.com/dbu)
 
+.
--- a/README.md
+++ b/README.md
@@ -126,4 +126,3 @@ And especially ones who create and maintain new extensions:
 - Kudryashov Konstantin [everzet](http://github.com/everzet)
 - David Buchmann [dbu](https://github.com/dbu)
 
-.
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "gedmo/doctrine-extensions",
+    "name": "pmishev/doctrine-extensions",
     "type": "library",
     "description": "Doctrine2 behavioral extensions",
     "keywords": [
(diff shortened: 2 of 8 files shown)


l3pp4rd:
well, these kind of enhancements only confuse people. if you throw in a crazy number of options into any project, no one will use it in the end.

- all options to repository methods are not needed, you may extend the repository and override or customize it based on your needs, most of the users do not need it on day to day usage..
- option name changes are introducing backward incompatible changes, there must be a good reason in order to change something like that on minor version update.
- I do not see the point for this level base option, if it does need to be 1 for your needs only, please use your own version for your project.

pmishev:
Everything I wrote was because I did need it for the project I'm working on and I would have been very happy if the functionality was already there. I chose to fork the project, as it would have been a much bigger pain to extend it in my project.

1. Which option do you think is not needed? I can give you the real-world use for it right away and then you could judge if no one else beside me would need it.

2. I haven't made a single backwards incompatible change - I have only added parameters, nothing existing is changed. Perhaps you mean the comment change from 'childStart' to 'childOpen'? I haven't changed the option - I just fixed the comment, because it's wrong.

3. The point of the level base option is that in my case, I was working with an existing DB which levelled the root nodes from 1, so how was I supposed to use your otherwise great extension then?
I don't see how it could possibly hurt to have an option like this, but be it as you say.

4. And finally I did fix the recover() and verify() options, as they were not at all considering wrong levels of the nodes, which I'm sure would be an issue for a good number of other people.
Also the childrenQueryBuilder() method was flawed - it allowed you to order by several fields, but the direction had to be the same for all of them.

l3pp4rd:
I have missed that **childOpen** was a documentation change, that can be accepted of course if it was a mistake there.
Concerning options, I do not care about real world use case, it was not needed for most of the users for a long time (etc tree level **base**) option. And I will not accept this, since it will be completely unnecessary for the most of the users.
Regarding tree recovery, the verify check could be removed at all in next minor version change, adding an option seems cumbersome. The level recovery is OK and I would accept it. But all path as string or include options are not needed, you can easily add them to the repository of your own or modify query builder. The more options there are the more complex it is to understand, I'm sad there are already some of those options which are not necessary.

Issue #961, opened 2014-01-24
[Uploadable] multiple files in the same entity


aitboudad:
Hey,
Is it possible to upload multiple files in the same entity like:

``` php
    /** 
     * @Gedmo\UploadableFileName
     */
    private $filename;

    /** 
     * @Gedmo\UploadableFileName
     */
    private $thumb;
```

FranzBruckner:
Hi there,
i have implemented this feature rudimentary. All works fine, but there is still some work to let the fork become stable.

The main drawback is broken BC. 
See https://github.com/FranzBruckner/DoctrineExtensions for more details.

webdevilopers:
:+1:

tamcy:
Hi,

I also wrote my own version of multiple uploadables in a single entity. It's different from the one by @FranzBruckner that you don't need any modification to the annotation for single uploadable.
https://github.com/tamcy/DoctrineExtensions
https://github.com/tamcy/DoctrineExtensions/blob/master/doc/uploadable.md

I can create a PR if thing is right, but currently I only added this feature to annotation.

l3pp4rd:
@tamcy if that is backward compatible would surely be glad to merge it. otherwise it could wait till 2.5 release milestone.

diegocastro:
+1 for this.

pimolo:
@l3pp4rd Any reason why this is not merged into v2.4?
Do you want some help to release v2.5?

l3pp4rd:
well, I try to maintain only bug fixes now, I do not work with php anymore, unless someone takes over the maintenance, extensions will remain at 2.4 with minor patches

PR #1342, opened 2015-06-24
Add multiple uploadables feature

tamcy:
Ref: #961 

This changeset adds multiple uploadables support in a single entity. 

* A new `identifier` property is added to all annotations in the uploadable familty (`@Uploadable`
`@UploadableFileMimeType`, `@UploadableFileName`, `@UploadableFilePath`, and `@UploadableFileSize`), with a default value `_default`. This ensures that no code change is needed in single upload situation.
* A new `@Uploadables` annotation added to support multiple uploadable behavior.

The bad thing is I only added multiple support to annotation (I probably don't have spare time to write and test the yml and xml config, so need external help). The good thing is the original configuration won't break.

However there are cases of potential bc-break:

- A new parameter `identifier` is also added to `FilenameGeneratorInterface::generate()` as the forth argument. Currently this is used by the `FilenameGeneratorSha1` class to prevent having the same SHA1 result when the same filename and extension are passed to different file fields (with different identifiers) in the same entity. This may affect users who implements their own `FilenameGeneratorInterface`.
- Visibility of method `processFile()` and `moveFile()` in `UploadableListener`  changed from `public` to `protected` because their signature are changed and they don't look right to be public. This may affect users who override the original `UploadableListener`.

Finally, not related to this project, but still worth mentioning is that a change to stof's [DoctrineExtensionsBundle](https://github.com/stof/StofDoctrineExtensionsBundle/) is also needed to support multiple uploadables in a single entity. Modify the `markEntityToUpload()` method of  `\Stof\DoctrineExtensionsBundle\Uploadable\UploadableManager` as below:

```php
    public function markEntityToUpload($entity, $fileInfo, $identifier = '_default')
    {
        if (is_object($fileInfo) && $fileInfo instanceof UploadedFile) {
            $fileInfoClass = $this->fileInfoClass;

            $fileInfo = new $fileInfoClass($fileInfo);
        }

        $this->listener->addEntityFileInfo($entity, $fileInfo, $identifier);
    }

```

Usage of this method will become:
```php
$manager->markEntityToUpload($file, $uploadedFile, 'image_thumb');
```


l3pp4rd:
thanks it looks like a great improvement, regarding YML and XML hope people will contribute if they need it, I will merge it with minor version change, since it is worth mentioning about the possible BC if extended.

l3pp4rd:
hey @stof that may have effect on the bundle as mentioned. I think I'll stop any changes to API after 3.0.0, so probably these will be the last changes to adapt unless someone starts to maintain extensions apart from myself :) 3.0.0 won't be released soon, few months at least

PR #1344, opened 2015-06-30
Index added for listing logs of entity

ismailbaskin:


l3pp4rd:
there was a comma missing after the annotation. but in general you may customize your log entry class by extending mapped supperclass and adding this index for your use case. too many indexes slows down updates and inserts

PR #1346, opened 2015-07-04
Fix DoctrineExtensions for latest ODM

malarzm:
Since we have removed `scheduleExtraUpdate` in latest ODM few extensions stopped working. Affected are

- [x] Loggable
- [x] SoftDeleteable
- [ ] Tree

@l3pp4rd I'm not sure when I'll be able to work on Tree extension, possibly good idea would be to merge this PR and leave fixing Tree to somebody else/future me


malarzm:
Eh turns out that my fixes require latest ODM and are failing on BETA-13

l3pp4rd:
thanks for the initial fixes @malarzm but the tests would fail either way, will try to fix the tree

l3pp4rd:
had to skip materialized path locking test for odm, not sure what is missing there, but hope people will be able to resolve that in case if needed

malarzm:
:+1:

PR #1354, opened 2015-07-16
Performance fix in TranslatableListener

krispypen:
Throwing an exception is really slow compared to an if-statement.

For example: In our application this exception was throwed 300 times (loading 300 entities without locale), after this change the page loads 50ms faster.


stof:
I suggest creating a private method ``isValidLocale`` and using it in both places instead of duplication the validation logic

krispypen:
@stof true, I changed it

l3pp4rd:
thanks ;)

PR #1184, opened 2014-11-03
Restore the translatable listener's skipping of translation loading after hydration

ciprianvieru:
During the initialization phase the translatable listener is setup to not load translations (by setting the skipOnLoad property to true).Then a query is hinted to load translations.
Further queries (not hinted) should not load translations, but, during the hydration of the hinted query, the listener's skipping of translation loading is hard-reset to always load translations.

PR #1357, opened 2015-07-17
Restore TranslatableListener::skipOnLoad after hydration

mente:
Followup of #1184 . Have written own implementation + tests.

Do you support older versions of extensions? I would like to apply it to Doctrine2.4 branch as well


l3pp4rd:
thanks, well, wip 2.4 branch is dropped and the master branch should support versions from 2.3 to 2.5 so there is nothing to backport actually. thanks for the tests, good work

mente:
Then we should be safe. Thanks!

PR #1361, opened 2015-07-22
Fix notice "Undefined index: groups" in sortable

zorn-v:
This notice happens when no groups defined and throw ContextErrorException in Symfony in dev env for example.


l3pp4rd:
thanks ;)

PR #1362, opened 2015-07-22
[Reference Entity] Add a new option "pull" to pull references before delete

### LINKED ISSUE #1360: [Reference_Integrity] Nullify and ReferenceMany.
I have two documents Meeting and Guest (in Symfony).

```
class Meeting
{
    /** 
     * @MongoDB\ReferenceMany(targetDocument="Guest",inversedBy="meetings",simple=true)
     */
    private $guests;
}

class Guest
{
    /** 
    * @MongoDB\ReferenceMany(targetDocument="Meetings",mappedBy="guests",simple=true)
    * @Gedmo\ReferenceIntegrity("nullify")
    */
    private $meeting;
}
```

So for each Meeting I have an array of referenced guests.

-> When I remove a guest from the dB, I wish that it goes deleted from each meetings.

It is done but the full list of guests from each impacted meeting is setted to null.

Is it not possible to configure the listener so that it only remove the reference from the array instead of nullify the full array ?

Alsatian67:
Added a new option to the ReferenceIntegrity mapping to fix issue #1360  :

```
@Gedmo\ReferenceIntegrity("pull")
```
This option aims to remove ONLY the document which is removed from the Referenced documents collection on the owning side. (The nullify option sets the references collection on null).

+ renamed '$object' variable used twice.
+ replaced 'entity' anotations with 'document.

closes #1360 

Please review

### DIFF
--- a/lib/Gedmo/ReferenceIntegrity/Mapping/Validator.php
+++ b/lib/Gedmo/ReferenceIntegrity/Mapping/Validator.php
@@ -12,6 +12,7 @@ namespace Gedmo\ReferenceIntegrity\Mapping;
 class Validator
 {
     const NULLIFY = 'nullify';
+    const PULL = 'pull';
     const RESTRICT = 'restrict';
 
     /**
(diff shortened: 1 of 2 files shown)


l3pp4rd:
thanks, would be great if you could pull a test as well

Alsatian67:
Done in #1364

PR #1366, opened 2015-07-27
Configure versioning without fields area in Entity yaml configuration

seytar:
Problem: https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1357

using:
```yaml
UserBundle\Entity\User:
    type:  entity
    table: fos_user
    gedmo:
        loggable: true
        versioned:
            fields:
                - email
                - username
        soft_deleteable:
            field_name: deleted
            time_aware: false
    id:
        id:
            type: integer
            generator:
                strategy: AUTO
    fields:
        name:
          type: string
          length: 255
          nullable: true
          gedmo:
             - versioned
        created:
          type: datetime
          gedmo:
            timestampable:
              on: create
        updated:
          type: datetime
          gedmo:
            timestampable:
              on: update
        deleted:
            type: datetime
            nullable: true
```

### DIFF
--- a/lib/Gedmo/Loggable/Mapping/Driver/Yaml.php
+++ b/lib/Gedmo/Loggable/Mapping/Driver/Yaml.php
@@ -42,6 +42,14 @@ class Yaml extends File implements Driver
                     $config['logEntryClass'] = $cl;
                 }
             }
+
+            if (isset($classMapping['versioned'])) {
+                if (isset ($classMapping['versioned']['fields'])) {
+                    foreach ($classMapping['versioned']['fields'] as $field) {
+                        $config['versioned'][] = $field;
+                    }
+                }
+            }
         }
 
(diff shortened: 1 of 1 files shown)


l3pp4rd:
it is possible to add annotations in protected properties if you extend the base class which is explained in the PR related to your referenced issue. DoctrineExtensions is a library and it will not add options to make it convenient for integration with other frameworks. most of the extensions would need similar changes which does the same thing, but differently. all those options confuse people, there are too many already. I actually never use such frameworks like fos bundles, but that is up to you, if it does not get into your way in every step.
My point is that doctrine extensions requires you to manage your entity mapping, and that is it.
Have you tried putting these mappings into **app/Resources/FOSUserBundle/config/doctrine..orm.yml** ? I think symfony framework allows you to override resources if file locator is used.

Issue #1277, opened 2015-04-01
[Sortable] (ODM) Fatal error: Class 'Doctrine\ORM\UnitOfWork' not found


Rbn3D:
Fatal error: Class 'Doctrine\ORM\UnitOfWork' not found
in Gedmo/Sortable/SortableListener.php on line 484

if I change 
```
use Doctrine\Common\EventArgs;
use Doctrine\Common\Persistence\Mapping\ClassMetadata;
use Doctrine\Common\Util\ClassUtils;
use Doctrine\ORM\UnitOfWork;
use Gedmo\Mapping\MappedEventSubscriber;
use Doctrine\ORM\Proxy\Proxy;
use Gedmo\Sortable\Mapping\Event\SortableAdapter;
```

to 

```
use Doctrine\Common\EventArgs;
use Doctrine\Common\Persistence\Mapping\ClassMetadata;
use Doctrine\Common\Util\ClassUtils;
use Doctrine\ODM\MongoDB\UnitOfWork;
use Gedmo\Mapping\MappedEventSubscriber;
use Doctrine\ODM\MongoDB\Proxy\Proxy;
use Gedmo\Sortable\Mapping\Event\SortableAdapter;
```

In that file, it works without problems.

I think you should consider use an adapter to handle UOW stuff

Thanks in advance.

Rbn3D:
I'm using version v2.3.12

l3pp4rd:
the sortable does not support odm

rubenrua:
The sortable support ODM, this is the commit to enable it: https://github.com/Atlantic18/DoctrineExtensions/commit/3279ab7803e781f6e7ad0582b289d65e06f88325

Don't close this issue.

PR #1373, opened 2015-07-30
Hotfix sortable odm without orm

rubenrua:
Fix issue #1277.

Allow the use of sortable extension in MongoDB without installing the ORM.


l3pp4rd:
thanks, I do not use extensions much recently, forgot about sortable + mogodb support :)

rubenrua:
Could you create a new tag with the last fixes? thx.

PR #1375, opened 2015-08-03
Just make the comment even more explicit

mablae:
Maybe others then do not need 30 mins to figure out... :-)


l3pp4rd:
thanks ;)

PR #1377, opened 2015-08-04
Update for Entity not using id as column name

durich:
This change is for the people that does not use the convention of "id" as the primary key column name

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/Closure.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/Closure.php
@@ -254,6 +254,8 @@ class Closure implements Strategy
             $descendantColumnName = $this->getJoinColumnFieldName($em->getClassMetadata($config['closure'])->getAssociationMapping('descendant'));
             $depthColumnName = $em->getClassMetadata($config['closure'])->getColumnName('depth');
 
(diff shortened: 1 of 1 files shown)


l3pp4rd:
thanks ;)

lib/Gedmo/Tree/Strategy/ORM/Closure.php
```diff
@@ -254,6 +254,8 @@ public function processPostPersist($em, $entity, AdapterInterface $ea)
             $descendantColumnName = $this->getJoinColumnFieldName($em->getClassMetadata($config['closure'])->getAssociationMapping('descendant'));
             $depthColumnName = $em->getClassMetadata($config['closure'])->getColumnName('depth');
 
+            $referenceId = $em->getClassMetadata($config['closure'])->getAssociationMapping('ancestor')['sourceToTargetKeyColumns'][$ancestorColumnName];
```
stof: the min PHP version is 5.3 in the composer.json, and this syntax is not compatible with it

durich: yeah, you are right, will send a new pull request

durich:
I submitted this pull request for php 5.3 compat [#1381](https://github.com/Atlantic18/DoctrineExtensions/pull/1381)

Issue #1309, opened 2015-05-12
Embeddables support for Doctrine >= 2.5


mgrajcarek:
Hello,

Do you plan to add support for Embeddables to existing extensions?
I have recently tried to add Loggable into Embeddable class, like this

```xml
    <embeddable name="Dvs\VariantBundle\Entity\Embeddable\ConsumerPackageSize">
        <field name="quantity" column="quantity" type="decimal" precision="10" scale="3" nullable="false" >
            <gedmo:versioned/>
        </field>
        <field name="amount" column="amount" type="decimal" precision="10" scale="3" nullable="true" >
            <gedmo:versioned/>
        </field>
        <field name="unit" column="unit" type="string" nullable="true" >
            <gedmo:versioned/>
        </field>

        <gedmo:loggable log-entry-class="Gedmo\Loggable\Entity\LogEntry"/>
    </embeddable>
```
Unfortunately, it ends in schema errors:

```
 [Gedmo\Exception\InvalidMappingException]                                                                                                               
  Class must be annotated with Loggable annotation in order to track versioned fields in class - Dvs\VariantBundle\Entity\Embeddable\ConsumerPackageSize 
```
It would be great if DoctrineExtensions could be used with Embeddables feature!

l3pp4rd:
Hi, well I currently do not work with extensions and only maintain it so it could be stable for the users. if you need that kind of functionality, you could submit a pull request with a test case to prevent regression..

fabsb:
@l3pp4rd I think this issue can be closed now

PR #1379, opened 2015-08-04
Add loggable support for embedded entities introduced by doctrine >=2.5

fabsb:
relates to #1309

### DIFF
--- a/lib/Gedmo/Loggable/Mapping/Driver/Xml.php
+++ b/lib/Gedmo/Loggable/Mapping/Driver/Xml.php
@@ -60,6 +60,9 @@ class Xml extends BaseXml
         if (isset($xmlDoctrine->{'reference-one'})) {
             $this->inspectElementForVersioned($xmlDoctrine->{'reference-one'}, $config, $meta);
         }
+        if (isset($xmlDoctrine->{'embedded'})) {
+            $this->inspectElementForVersioned($xmlDoctrine->{'embedded'}, $config, $meta);
+        }
 
         if (!$meta->isMappedSuperclass && $config) {
             if (is_array($meta->identifier) && count($meta->identifier) > 1) {
--- a/tests/Gedmo/Loggable/Fixture/Entity/Address.php
+++ b/tests/Gedmo/Loggable/Fixture/Entity/Address.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace Loggable\Fixture\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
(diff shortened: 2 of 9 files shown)


l3pp4rd:
great, could you also update https://github.com/Atlantic18/DoctrineExtensions/blob/master/schemas/orm/doctrine-extensions-mapping-2-2.xsd it may not pass xsd validation now

fabsb:
@l3pp4rd I made another PR for doctrine2 as it is an issue on their side for not allowing other namespaced tags or attributes for the embedded tag. I hope the PR gets through and the schema deployed soon.

l3pp4rd:
ok, thanks ;)

Ocramius:
Just pinging back here: doctrine/doctrine2#1488 was closed - I think we should move forward to splitting mapping files, rather than mixing them together.

cordoval:
@fabsb do you think you can do the mapping file splitting as @Ocramius suggested? I am duplicating my columns just because of this:

```php
<?php

public function setPatientInformation(PatientInformation $patientInformation)
    {
        $this->patientInformation = $patientInformation;

        $this->updateAddress();

        return $this;
    }

private function updateAddress()
    {
        $this->address1 = $this->patientInformation->getAddress()->getAddress1();
        $this->address2 = $this->patientInformation->getAddress()->getAddress2();
        $this->city = $this->patientInformation->getAddress()->getCity();
        $this->state = $this->patientInformation->getAddress()->getState();
        $this->county = $this->patientInformation->getAddress()->getCounty();
        $this->zipcode = $this->patientInformation->getAddress()->getZipcode();
    }
}
```

Thanks 😊

fabsb:
@cordoval I don't see how separating the extension mappings from the doctrine mapping files would help you with your issue.

cordoval:
I doesn't help it, all i want is just to use the feature. But i can't it seems. In any case I have resolved the problem another way, different from the one above.

fabsb:
Well you could use the feature as it is now without separating the mappings for the extension as I have used it in other projects of mine.

@l3pp4rd If you want I could work on separating the mapping as suggested but note that it would break BC so maybe for 2.5 or 2.6?

cordoval:
weird, i did try, was using yaml though so not sure why it did not work for me.

l3pp4rd:
@fabsb BC break is allowed in master 3.0.0, but have to add reference in [upgrade guide](https://github.com/Atlantic18/DoctrineExtensions/blob/master/UPGRADE-3.0.md) so if you could make a PR you are welcome.

PR #1380, opened 2015-08-04
Update for Oracle Compatibility

durich:
* Change level to levelNum as level is a reserved word in Oracle Databases and cannot be used in a query without using ""

* OCI8 doesnt support PDO::FETCH_KEY_PAIR, so I used PDO:FETCH_NUM an emulated  PDO::FETCH_KEY_PAIR result


l3pp4rd:
thanks ;)

stof:
@durich I suggest opening an issue (or even a PR) on doctrine DBAL to request the implementation of the key pair fetching in their oci8 driver. It will be easy to support it in https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
I was not even aware that this fetch mode was supported in PDO actually

PR #1382, opened 2015-08-04
Let Closure strategy work better with other listeners/subscribers

### LINKED ISSUE #1363: [Tree] ORM Closure Strategy - "level" property update doesn't play well with other Doctrine listeners
I've been trying to track this issue down for a while.

Before I describe the issue, this is what I've got going on:
------------

- I have a *Page* entity that is setup to be a tree with closure strategy.
- I have my own Doctrine event subscriber that can persist a new *Route* entity whenever a *Page* is created or updated. If necessary, my subscriber flushes the EM for a second time in the *postFlush* event.
- My own subscriber is being executed **after** Gedmo's TreeListener.

Now for the problem:
-------------

1. Gedmo's Closure strategy tells Doctrine to trigger an extra update on [*line 351*](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Strategy/ORM/Closure.php#L351). That's working well.
2. I believe that [*line 357*](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Strategy/ORM/Closure.php#L357) is supposed to make Doctrine keep track of the new level.
3. However, when my subscriber flush the EM, it seems to revert *level* back to the old/invalid value (*NULL* in my case).

Potential solution:
------------------
If I edit the Closure strategy to actually change the level property value on the entity, everything appears to work well.

To do this, I've changed [*lines 351-357*](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Strategy/ORM/Closure.php#L351-L357) to: 

    $levelProp = $meta->getReflectionProperty($config['level']);
    $uow->scheduleExtraUpdate(
        $node,
        array($config['level'] => array(
            $levelProp->getValue($node), $level,
        ))
    );
    $levelProp->setValue($node, $level);
    $uow->setOriginalEntityProperty(spl_object_hash($node), $config['level'], $level);

I don't really understand all of what's going on under the hood in Doctrine to know whether what I did is a big no-no, or whether or not $uow->setOriginalEntityProperty(..) is still needed, but this does solve my problem.

Do you think it's ok to make a PR with this change? Concerns?

simshaun:
Fixes #1363

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/Closure.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/Closure.php
@@ -286,6 +286,8 @@ class Closure implements Strategy
             } elseif (isset($config['level'])) {
                 $uow->scheduleExtraUpdate($node, array($config['level'] => array(null, 1)));
                 $ea->setOriginalObjectProperty($uow, spl_object_hash($node), $config['level'], 1);
+                $levelProp = $meta->getReflectionProperty($config['level']);
+                $levelProp->setValue($node, 1);
             }
 
             foreach ($entries as $closure) {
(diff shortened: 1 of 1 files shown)


l3pp4rd:
thanks ;)

simshaun:
Can we get a new tagged version that contains this fix?

PR #1387, opened 2015-08-11
Add config.yml example

Flower7C3:
Source: http://stackoverflow.com/a/16671156/2910183 and http://stackoverflow.com/review/suggested-edits/9128708

### DIFF
--- a/doc/softdeleteable.md
+++ b/doc/softdeleteable.md
@@ -15,6 +15,7 @@ Features:
 Content:
 
 - [Including](#including-extension) the extension
+- [config.yml example](#config-example)
 - Entity [example](#entity-mapping)
 - [Yaml](#yaml-mapping) mapping example
 - [Xml](#xml-mapping) mapping example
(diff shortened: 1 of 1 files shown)


l3pp4rd:
please put it in **doc/symfony.md** frameworks are not related to library listener documentation, would clutter the content

stephanvierkant:
I agree, this library should be framework agnostic.

Maybe we should put a link to https://github.com/stof/StofDoctrineExtensionsBundle or https://symfony.com/doc/current/bundles/StofDoctrineExtensionsBundle/index.html. What do you think, @l3pp4rd?

l3pp4rd:
It is much more convenient to use this library on symfony2 by following the documentation in `doc/symfony.md`, if people find it too hard, then they may just google `symfony2 doctrine extensions` they will find bundle in the first result, for others who are not capable of following these steps, nothing will help :)
And it is true, that this is a library, not something which supports all frameworks out there. We can put links to bundles or modules in zf in doc/*.md but that is it in my opinion. This pull request is doing it wrong, it looks like extensions are working only with symfony2 in that way.

PR #1394, opened 2015-08-21
Fix typo in Blameable documentation

heitorvrb:
The @Gedmo\Blameable property "fields" does not exist. The property is "field". That had me scratching my head for a while when following the docs, until I noticed it.


l3pp4rd:
thanks ;)

PR #1401, opened 2015-09-03
Added quotation marks for Yaml validity

ThomasLandauer:
I have this line in my app/config/services.yml which gives a YML syntax error:
    calls:
      - [ setAnnotationReader, [ @annotation_reader ] ]
In the Pull Request, did I only change the docs or the very source file? If necessary, please update the source too.


l3pp4rd:
thanks ;)

PR #1402, opened 2015-09-04
 - fixed centos file order for import

Imunhatep:
ERROR: exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Cannot redeclare class Gedmo\Mapping
\Annotation\Reference' in /home/someuser/somewww/vendor/gedmo/doctrine-extensions/lib/Gedmo/Mapping/Annotation/Reference.php:22


l3pp4rd:
thanks ;)

PR #1405, opened 2015-09-08
Fixed version type casting in document repository to allow search in MongoDB

erriond:
Changed version to always type cast to integer. Otherwise it's not possible to search for a revision to revert back to (when using MongoDB).

If the version is not type casted, no log entries will be found in the MongoDB database, since the MongoDB operation $lte doesn't allow strings for comparison.


l3pp4rd:
thanks ;)

PR #1409, opened 2015-09-15
Fix the incorrect fix for 'Cannot redeclare class Gedmo\Mapping\Annotation\Reference' error

loban:
The previous fix removes the 'Cannot redeclare class Gedmo\Mapping\Annotation\Reference'
error but does not fix the real bug. Sorting the list of files does not guarantee that
the same error will not show up again when new annotations are added. Instead, using an
include_once instead of include will not only prevent this bug happening again, but also
remove the requirement that the file must be loaded with require_once. This is useful
because the autoload/files section of composer.json loads files with require instead
of require_once. Adding this file to the composer.json of a library that uses
DoctrineExtensions will allow a developer to use that library without having to worry
about manually loading this file (since composer will include it in it's autoload).

### DIFF
--- a/lib/Gedmo/Mapping/Annotation/All.php
+++ b/lib/Gedmo/Mapping/Annotation/All.php
@@ -2,16 +2,13 @@
 
 /**
 * Contains all annotations for extensions
-* NOTE: should be included with require_once
 *
 * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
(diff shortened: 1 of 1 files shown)


l3pp4rd:
the assort is needed because the file order matters. centos lists files by default not in ascending order and the bug was related to that. **include_once** is ok, but revert the asort please.

PR #1411, opened 2015-09-18
Updating file name to camel case

aqnouch:
Updating file name to camel case, otherwise, it will not be detected in a project that respect naming convention.


l3pp4rd:
these changes cannot be applied, not all OS will be backward compatible with this and I do not see it reasonable to change it all over the library..

maqnouch:
I just came across this problem wile deploying a symfony application that was developed in windows to a linux server, by making this change I was able to resolve the problem, (the other solution is to change the class name in the annotation but when I changed my ide prompted me a warning message)
Sorry for bothering you ^_^

l3pp4rd:
if you do not see, you made a class name different. probably you have made a mistake and instead of `@Timestampable` used `@TimeStampable` somewhere on your entity, search in project with case sensitive mode.

Issue #1093, opened 2014-06-08
RelativeSlugHandler doesn't work properly with TreeSlugHandler?


gremo:
When slug is updated in Category, each Product slug (with the same assigned category) is updated. For example,  category slug is /food/fruits, product slug is /food/fruits/apple. When /food/fruits slug changes to /food/good-fuits, apple becomes /food/good-fruits/apple. **This is correct**

However, if slug changes for an ancestor (i.e /food) apple slug doesn't change. You end up with /good-food/fruits category, and /food/fruits/apple product.

Is this the default behaviour? Classes (simplified):

```php
class Category
{
    /**
     * @Gedmo\Slug(handlers={
     *      @Gedmo\SlugHandler(class="Gedmo\Sluggable\Handler\TreeSlugHandler",
     *          options={
     *              @Gedmo\SlugHandlerOption(name="parentRelationField", value="parent"),
     *              @Gedmo\SlugHandlerOption(name="separator", value="/")
     *         }
     *      ),
     *      @Gedmo\SlugHandler(
     *          class="Gedmo\Sluggable\Handler\InversedRelativeSlugHandler",
     *          options={
     *              @Gedmo\SlugHandlerOption(name="relationClass", value="Product"),
     *              @Gedmo\SlugHandlerOption(name="mappedBy", value="category"),
     *              @Gedmo\SlugHandlerOption(name="inverseSlugField", value="slug")
     *          }
     *      )
     * }, fields={"name"})
     * @ORM\Column(unique=true)
     */
    protected $slug;
}
```

```php
class Product
{
    /**
     * @Gedmo\Slug(handlers={
     *      @Gedmo\SlugHandler(class="Gedmo\Sluggable\Handler\RelativeSlugHandler",
    *           options={
     *              @Gedmo\SlugHandlerOption(name="relationField", value="category"),
     *              @Gedmo\SlugHandlerOption(name="relationSlugField", value="slug"),
     *              @Gedmo\SlugHandlerOption(name="separator", value="/")
     *          }
     *    )
     * }, updatable=true, fields={"name"})
     * @ORM\Column(length=64, unique=true)
     */
    protected $slug;
}
```

l3pp4rd:
well thanks for the details, not sure when I will be able to spare time for this.

PR #1414, opened 2015-09-21
Add failing test for #1093

cmfcmf:
This adds a failing test for the problem described in #1093.

### DIFF
--- a/tests/Gedmo/Sluggable/Fixture/Handler/People/Occupation.php
+++ b/tests/Gedmo/Sluggable/Fixture/Handler/People/Occupation.php
@@ -70,7 +70,7 @@ class Occupation
      */
     private $level;
 
-    public function setParent(TreeSlug $parent = null)
+    public function setParent(Occupation $parent = null)
     {
         $this->parent = $parent;
     }
--- a/tests/Gedmo/Sluggable/Handlers/BothSlugHandlerTest.php
+++ b/tests/Gedmo/Sluggable/Handlers/BothSlugHandlerTest.php
@@ -89,9 +89,15 @@ class BothSlugHandlerTest extends BaseTestCaseORM
         $this->em->persist($developer);
         $this->em->flush();
 
-        // It works if the following line is removed.
+
+        // Works (but is not updated in the actual DB)
+        $herzult = $personRepo->findOneByName('Herzult');
+        $this->assertEquals('web/enthusiast/php/herzult', $herzult->getSlug());
+
+
         $this->em->clear();
 
+        // Does not work.
         $herzult = $personRepo->findOneByName('Herzult');
         $this->assertEquals('web/enthusiast/php/herzult', $herzult->getSlug());
     }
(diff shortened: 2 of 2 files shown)


cmfcmf:
I've updated the test slightly to clarfiy what the actual issue is:
- If both slug handlers are used in conjunction, the slug update is inconsistent between database and the in-memory objects of the EntityManager. 
- Take the following example: 
  - You have a tree entity (say "Occupation") with a unique tree-handler-based slug.
    - This generates a slug like `web/developer/php` for a tree node with the title `php` whose parent has the title `developer` whose parent has the title `web`.
    - Now if you update the `web` or `developer` node, the slugs of all child nodes are correctly changed and updated. **This works.**
  - Each occupation has many persons with names and of course a slug using the relative slug handler.
    - The goal is to prepend the slug of the tree to the person's name. If a person has `php` as occupation, it's slug will be `web/developer/php/person-name`. **This works.**
    - Now if the name of the `php` node is changed to `python`, it does update the person's slug to `web/developer/python/person-name`. **This works.**
    - Now here comes the problem: If a parent node changes it's slug, the slug of our poor PHP developer remains unchanged. I.e. if `developer` changes to `enthusiast`, the slug of the `php` *node* is correctly updated to `web/enthusiast/php`. **But the slug of the *person* is not updated and remains `web/developer/php/person-name`.** However, as evil ( :smirk: ) as this library is, it **does** correctly update the slug of the in-memory objects of the EntityManager. So the objects in-memory differ from the actual objects inside the database! Here are the corresponding code snippets:

In-memory handling (notice that it uses preg_match to change all the objects with slugs starting with the changed slug):
```php
foreach ($uow->getIdentityMap() as $className => $objects) {
    // for inheritance mapped classes, only root is always in the identity map
    if ($className !== $mappedByConfig['useObjectClass']) {
        continue;
    }
    foreach ($objects as $object) {
        if (property_exists($object, '__isInitialized__') && !$object->__isInitialized__) {
            continue;
        }
        $oid = spl_object_hash($object);
        $objectSlug = $meta->getReflectionProperty($mappedByConfig['slug'])->getValue($object);
        if (preg_match("@^{$oldSlug}@smi", $objectSlug)) {
            $objectSlug = str_replace($oldSlug, $slug, $objectSlug);
            $meta->getReflectionProperty($mappedByConfig['slug'])->setValue($object, $objectSlug);
            $ea->setOriginalObjectProperty($uow, $oid, $mappedByConfig['slug'], $objectSlug);
        }
    }
}
```
Actual things which happen to the database:
```php
public function replaceInverseRelative($object, array $config, $target, $replacement)
{
    $em = $this->getObjectManager();
    $qb = $em->createQueryBuilder();
    $qb->update($config['useObjectClass'], 'rec')
        ->set('rec.'.$config['slug'], $qb->expr()->concat(
            $qb->expr()->literal($target),
            $qb->expr()->substring('rec.'.$config['slug'], strlen($replacement)+1)
        ))
        ->where('rec.'.$config['mappedBy'].' = :object')
    ;
    $q = $qb->getQuery();
    $q->setParameters(compact('object'));

    return $q->execute();
}
```

l3pp4rd:
Well, happy that you provided these details, will have a look when possible.

l3pp4rd:
not sure whether it is the best option, but before it was checking the occupation_id which was different for the person in interest since he had an occupation assigned which was inheriting the one modified.
So now it does where based on previous slug.

cmfcmf:
Thank you, it's working :-)

PR #1418, opened 2015-09-23
Quote separator and slug when using preg_*, fixes #1240.

### LINKED ISSUE #1240: Sluggable: can't use "+" as the separator
I'm using the extensions through the StofDoctrineExtensionsBundle. I'm trying to change the default separator to "+":

    /**
     * @var string $slug
     * @Gedmo\Slug(fields={"label"}, separator="+")
     * @ORM\Column(length=130, unique=true)
     * @Expose
     */
    protected $slug;

It doesn't work, I get the following error:
                                                                             
    [Symfony\Component\Debug\Exception\ContextErrorException]
    Warning: preg_match(): Compilation failed: nothing to repeat at offset 24     

When trying to escape the "+" with "\\+", I don't have this error any-more but in the database I've got the following:

    my\+slug

Same thing with a double escape "\\\". Any idea?

cmfcmf:


l3pp4rd:
thanks ;)

PR #1420, opened 2015-09-30
Fix Timestampable triggers php notice when trying to persist and flush embedded document

munkie:
Fix TimestampableListener triggers php notice when trying to persist and flush embedded document without parent document. Yes, it is uncommon case, but happens sometimes.
ODM unit of work says that such embedded documents are scheduled for insertion, but their change set actualy is empty and they are not flushed.


l3pp4rd:
thanks ;)

PR #1422, opened 2015-09-30
Fix CustomTransliteratorTest fail on Travis CI

munkie:
Since transliteration map for Chinese language has changed in 1.1.0 version of  behat/transliteration library, i've changed the expected value in CustomTransliteratorTest.
This should make build green on Travis CI


l3pp4rd:
thanks ;)

PR #1424, opened 2015-10-08
[Translatable] Inconsistent result on inherited entity

moroine:
Hi,

I discover what I think is a bug but I am not able to fix it but I made some tests that reveals that bug. I hope someone will be able to complete my PR.

I have three entities:
- Person: an abstract entity that is translatable (in my test case there is no field translatable as it works fine)
- Staff & Student: both inherited from Person and add some translatable fields

I currently use the Query hint method to retrieve my entities in specific locale. The problem is when I query on the Person class which returns either Staff or Student, all Person fields are translated correctly but this is not the case for the subclass fields.

I found a work-around by switching listener translatable locale and not using query hint (see my test cases).

Finally while writing my unit tests, I found another issue which is if someone has already loaded an entity in some locale, the entity is in managed state on EntityManager. In that case requesting the same entity via EntityManager is not working and simply returns the entity used by this "someone".

Thank you,
Hoping this could help

### DIFF
--- a/tests/Gedmo/Translatable/Fixture/Issue/Person.php
+++ b/tests/Gedmo/Translatable/Fixture/Issue/Person.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Translatable\Fixture;
+
+use Gedmo\Mapping\Annotation as Gedmo;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity
+ * @Gedmo\TranslationEntity(class="PersonTranslation")
+ */
+class Person
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @Gedmo\Translatable
+     * @ORM\Column(name="name", type="string", length=128)
+     */
+    private $name;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setName($name)
+    {
+        $this->name = $name;
+    }
+
+    public function getName()
+    {
+        return $this->name;
+    }
+}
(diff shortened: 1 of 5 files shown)


l3pp4rd:
Hi, this week I'm busy, but probably will check it next week ;)

moroine:
@l3pp4rd Did you had time to see this case ?

Thank You

l3pp4rd:
ah, forgot about it, will have a look this week, cheers

moroine:
Thanks !

l3pp4rd:
I have analyzed the problem. when using joined table inheritance: 

- in stage where orm sql walker builds sql it transparently joins those tables
- it happens after the translatable makes joins for translations.
- though, it would be complicated to determine what table aliases there are for those joins

In result, it would require to refactor the translatable sql walker and so far it seems like it may take some time in order to understand how to structure the code. It looks like it would be possible to implement this, but currently I'm not sure when I would be able or have a will to do this.

I'm currently maintaining extensions so it could be as stable as possible without adding new features. This issue is necessary to support inheritance and probably I will need to tackle it one day. But I cannot say when.

moroine:
Thank You for your reply !

I found a work around as exposed I my unit-test. I don't really know the internal process but I would be happy to help if you need !

l3pp4rd:
it is a complex issue and the result to solve it. it's all in [translation sql walker](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Translatable/Query/TreeWalker/TranslationWalker.php)

What it does is:

- analyses the query components and looks for their metadata for translatable fields.
- if found, it then creates sql replacements with translation record join statements.

What happens when there is joined table inheritance:

- default sql walker, transparently [joins inherited tables](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query/SqlWalker.php#L342).
- we would need to know these table aliases which he joins and based on that at some point join add more translation joins based on inheritance metadata and translated fields

I think it is not an easy task and my estimate is around a day or two days of work to catch all corner cases.

moroine:
I see more or less ! 

I experienced some difficulties with table inheritance in doctrine ! I will take a look this week-end and try to find out some clues that may help you one day ;)

AkenRoberts:

Thank you!

PR #1432, opened 2015-10-19
casting identifier to string since it is a varchar database field

mariohiller:
I have had some strange performance issues with the Translatable Behavior after having arount 1 million rows in ext_translations. Queries were dropping to around 300 - 400ms. Checking the Queries I found the following:

This query, generated by the behavior, is slow ( ~300ms ):
SELECT * FROM `ext_translations` WHERE `object_class` = 'Demo\\DemoBundle\\Entity\\Demo' AND `locale` = 'en' AND `foreign_key` = 111 

This query is fast ( ~1ms ):
SELECT * FROM `ext_translations` WHERE `object_class` = 'Demo\\DemoBundle\\Entity\\Demo' AND `locale` = 'en' AND `foreign_key` = '111'

There are two posibilities to "solve" this. Changing the the foreign_key column to integer did the same performance boost. But since ids can also be alphanumeric this is probably a bad idea to change globally. Casting the identifier to string seems to be no problem, since it is a varchar field anyways. Local tests only increased performance so far.

Thanks for discussing this change with me ( maybe there is a totally easy and obvious different solution to this issue :) ).

Regards,
Mario

### DIFF
--- a/lib/Gedmo/Translatable/Mapping/Event/Adapter/ORM.php
+++ b/lib/Gedmo/Translatable/Mapping/Event/Adapter/ORM.php
@@ -81,7 +81,7 @@ final class ORM extends BaseAdapterORM implements TranslatableAdapter
             }
         } else {
             // load translated content for all translatable fields
-            $objectId = $wrapped->getIdentifier();
+            $objectId = (string)$wrapped->getIdentifier();
             // construct query
             $dql = 'SELECT t.content, t.field FROM '.$translationClass.' t';
             $dql .= ' WHERE t.foreignKey = :objectId';


l3pp4rd:
Hi, another way would be to either use personal translations or extend translation superclass and override mapping to integer type. the default string was chosen, because it would fit all cases out of the box, though sacrificing performance, but in most cases if you have a mil of records you may need to customize anyway. The default behavior should fit only basic needs.
I see that this optimization will help a lot with the default behavior, thanks!

l3pp4rd:
though, probably will still need to check metadata, whether it is integer or string..

mariohiller:
I have seen your changes in commit 8bf241d which reverts the effect of my changes :)
Doctrine\DBAL\Types\StringType::convertToPHPValue() does nothing to the value, so the resulting query is slow again. Or did I miss anything?

l3pp4rd:
maybe I've made a mistake here, will have a look later

TriAnMan:
@mariohiller I've made a workaround for preformance problems in 2.x versions of Translatable.
Could you look at #1512?

PR #1434, opened 2015-10-23
Close #1415

### LINKED ISSUE #1415: SortableRepository causes an error with order
Hi!
After updating, we had a problem with sortable extension. When we use querybuilder and write in select any aggregate function (for example select('COUNT(mqtt)') ) SortableRepository causes an error by adding orderBy.
CreateQueryBuilder was redefined in cb95546bdcad2a86b8f152784cc307084472b4b6 commit.

melnikoved:
Revert cb95546bdcad2a86b8f152784cc307084472b4b6 commit, because SortableRepository causes an error with orderBy in createQueryBuilder.(for example by using select('COUNT(alias)') )


l3pp4rd:
yes seems it introduced bad magic :)

PR #1451, opened 2015-11-10
[Timestampable] Handle microseconds with DateTime

zeliard91:
Change the way of instantiating DateTime objects so that they can have the microseconds.

### DIFF
--- a/lib/Gedmo/Timestampable/Mapping/Event/Adapter/ODM.php
+++ b/lib/Gedmo/Timestampable/Mapping/Event/Adapter/ODM.php
@@ -27,6 +27,7 @@ final class ODM extends BaseAdapterODM implements TimestampableAdapter
             return new \Zend_Date();
         }
 
-        return \DateTime::createFromFormat('U.u', microtime(true))->setTimeZone(new \DateTimeZone(date_default_timezone_get()));
+        return \DateTime::createFromFormat('U.u', number_format(microtime(true), 6, '.', ''))
+            ->setTimeZone(new \DateTimeZone(date_default_timezone_get()));
     }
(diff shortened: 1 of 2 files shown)


l3pp4rd:
hi isn't the default timezone used implicitly?

zeliard91:
Hi,

The default timezone is ignored when the second parameter of `DateTime::createFromFormat` is a timestamp ([doc](http://php.net/manual/en/datetime.createfromformat.php)), UTC timezone is used in these case.

->

```php
<?php

$time = number_format(microtime(true), 6, '.', '');

// 1447417093.236281

$date = \DateTime::createFromFormat('U.u', $time);

/*
DateTime Object
(
    [date] => 2015-11-13 12:18:13.236281
    [timezone_type] => 1
    [timezone] => +00:00
)
*/

// If the current timezone is passed in the format, the instance has an incorrect value 
// as the timestamp is based on UTC
$date = \DateTime::createFromFormat('U.u-e', $time.'-'.date('e'));

/*
DateTime Object
(
    [date] => 2015-11-13 12:18:13.236281
    [timezone_type] => 3
    [timezone] => America/St_Barthelemy
)
*/

// This looks like the best way to have a correct instance but I'm opened to a better way.
$date = \DateTime::createFromFormat('U.u', $time);
$date->setTimeZone(new \DateTimeZone(date_default_timezone_get()));

/*
DateTime Object
(
    [date] => 2015-11-13 08:18:13.236281
    [timezone_type] => 3
    [timezone] => America/St_Barthelemy
)
*/

```

l3pp4rd:
thanks for the details ;)

PR #1458, opened 2015-11-15
[Loggable] Possibility to override Mapping validation checks

sofakartoffel:


l3pp4rd:
fine, but how do you override it?

sofakartoffel:
Unfortunately it is not a straight forward or clean solution.
I override the `LoggableListener` and override the `getNamespace` method.
All mapping drivers should be available in the set namespace.

PR #1466, opened 2015-11-29
Added citext field type to valid fields for sluggable

danut007ro:
Hello,

I'm using postgresql and I'm trying to generate a slug for a case insensitive field (citext type). Is there another way to override mapping and allow citext as valid field type? Other that this pull request?


l3pp4rd:
sadly there is no simple way, will probably make this static

danut007ro:
Thanks for merging and release.

PR #1475, opened 2015-12-07
Making desactivation of soft deletable working

sylvaincombes:
This patch make sure that the filter is enabled in the onFlush event before applying the behavior.

I experienced malfunction of this behavior, when I did something like that : 

```
    $em->getFilters()->disable('softdeleteable');
    $product = $em->getRepository('AppBundle:Product')->findOneBy(['idWorldwide' => 'XSWQAZ']);
    $em->remove($product);
    $em->flush();
```

It doesn't work the soft deletable behavior is still applied, this patch correct this.

### DIFF
--- a/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
+++ b/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
@@ -55,6 +55,10 @@ class SoftDeleteableListener extends MappedEventSubscriber
         $uow = $om->getUnitOfWork();
         $evm = $om->getEventManager();
 
+        if (!$om->getFilters()->isEnabled('softdeleteable')) {
+            return;
+        }
+
         //getScheduledDocumentDeletions
         foreach ($ea->getScheduledObjectDeletions($uow) as $object) {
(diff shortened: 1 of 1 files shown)


l3pp4rd:
I think it would be easier to detach listener from event manager. it would be compatible with both, orm and odm

sylvaincombes:
Hi @l3pp4rd thanks for your work, I don't understand what you mean in your comment can you give me some hint on what you are expecting so I can try to do it ?

The patch seems to work : https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/95424505 
The failed one are likely due to misconfiguration, see https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/95424506 for example.

l3pp4rd:
I meant that to disable softdeleatable, you can create a [method like this](http://stackoverflow.com/questions/12738974/remove-doctrine-life-cycle-event-from-within-a-symfony2-controller/12747541#12747541) to switch it on or off based on your needs.

AkenRoberts:

Thank you!

Issue #1090, opened 2014-06-02
Infinite loop caused by BlameableListener


Padam87:
Apart from #1074, 352a5a87a78be9691ef53353b81866ac0151637c Causes an infinite loop in some cases.

Example:

User entity:

``` php
    /**
     * @var Picture
     *
     * @ORM\OneToOne(targetEntity="Picture", cascade={"persist", "remove"}, orphanRemoval=true)
     * @ORM\JoinColumn(name="picture_id", referencedColumnName="id")
     */
    protected $picture;
```

Picture entity:

``` php
    /**
     * @var User
     *
     * @Gedmo\Blameable(on="create")
     * @ORM\ManyToOne(targetEntity="User")
     * @ORM\JoinColumn(name="uploaded_by", referencedColumnName="id")
     */
    protected $uploadedBy;

    /**
     * @var User
     *
     * @ORM\OneToOne(targetEntity="User")
     * @ORM\JoinColumn(name="user_id", referencedColumnName="id")
     */
    protected $user;
```

PR #1480, opened 2015-12-11
[BlameableListener] Fix for infinite recursion in some cases

NielsJanssen:
Fix for the problem as stated in #1090.


NielsJanssen:
Checks are failing :cry:  but that seems to be due to a misconfiguration in the test environments for PHP 5.5 and 5.6. PHP 5.4 runs successfully.

l3pp4rd:
thanks, will need to fix this travis build script

PR #1490, opened 2015-12-17
Fix cyclical persists

adrienbrault:


l3pp4rd:
thanks ;)

adrienbrault:
Actually, a better implementation would be to improve the check bellow to also check for `UnitOfWork::STATE_NEW` and not just `UnitOfWork::STATE_MANAGED`

Issue #1493, opened 2015-12-21
Introduce TreeRepository traits and remove AbstractTreeRepository's


aramalipoor:
Sometimes in an app like Sylius they're extending their repositories from **their own** `AbstractRepository` so it is not possible to extend from `NestedTreeRepository` too.

In another hand, as far as I know creating a `AbstractTreeRepository` causes so much duplicates.

Is it a good approach to introduce a **trait** e.g. **NestedTreeRepositoryTrait** so that those systems can use its useful methods? If it is I would be happy to create a PR!

l3pp4rd:
hi, yes that would be a good idea. but should be done to all strategies, closure, path and nested. I was looking forward to remove these abstract repositories one day and now when orm 2.5 requires php 5.4 it would be good to move forward to min PHP 5.4 and bump minor version to extensions. So if you want, I'd be happy to merge it including removal of abstract tree repository nonsense ;)

PR #1496, opened 2015-12-23
[Tree] Introduce tree repository traits and remove abstract tree repositories

aramalipoor:
@l3pp4rd Regarding #1493 do you think this is a good start?


lrlopez:
Cool! I was going to propose a PR extending trait support to closure trees and just found this. :+1:

l3pp4rd:
@aramalipoor yes it is still a good idea, but we need to leave NestedTreeRepository as a class which uses NestedTreeRepositoryTrait, this would ensure backward compatibility if php 5.4 or higher is used. I will bump minor version to 2.5 but still, I do not want to break BC that much. Also lets name traits ending with **Trait** it conforms [psr](https://github.com/php-fig/fig-standards/blob/master/bylaws/002-psr-naming-conventions.md) and I do not want to fight with @stof :) even if I do not like the naming. @lrlopez your PR as merged, this one may take longer and should be rebased if needed.

aramalipoor:
@l3pp4rd You're right, I've reverted Abstract repositories so no BC ;) Anything else to correct/add?

pjedrzejewski:
:+1:

l3pp4rd:
great, thanks

PR #1499, opened 2016-01-06
Use same object class name as metadata factory in repository. Fixes #1123

### LINKED ISSUE #1123: Translation translate() persist the sublcass, findTranslations() looks for the parent class!
Inheritance, where annotations are placed only in the subclass:

```php
/**
 * @ORM\Entity()
 * @ORM\Table("holiday_package")
 */
class HolidayPackage extends Page
{
    /**
     * @ORM\Column(name="content", type="text", nullable=true)
     * @Gedmo\Translatable()
     *
     * @var null|string
     */
    private $content;
}
```

When you translate:

```php
$repository->translate($holidayPackage, 'content', 'en', 'Content...');
```
... it will insert `My\Entity\HolidayPackage` as `object_class`  in `ext_translation` table (**correct**).

However when you try to find a translation:

```php
$repository->findTranslation($holidayPackage);
```
... it will look for `My\Entity\Page` (**wrong**).

The problem is in [`TranslationRepository`](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php#L116) class (line 116) where you always look for the superclass:

```php
$entityClass = $wrapped->getMetadata()->rootEntityName;
```

Is this sort of a bug?

CraigMason:
closes #1123

This effectively changes 1 line to use `useObjectClass` from the translatable-listener configuration object for an entity that is part of an inheritance chain, and the base class does not have a translatable field.

### DIFF
--- a/lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
+++ b/lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
@@ -117,12 +117,12 @@ class TranslationRepository extends EntityRepository
         $wrapped = new EntityWrapper($entity, $this->_em);
         if ($wrapped->hasValidIdentifier()) {
             $entityId = $wrapped->getIdentifier();
-            $entityClass = $wrapped->getMetadata()->rootEntityName;
-            $translationMeta = $this->getClassMetadata(); // table inheritance support
-
             $config = $this
                 ->getTranslatableListener()
                 ->getConfiguration($this->_em, get_class($entity));
+
+            $entityClass = isset($config['useObjectClass']) ? $config['useObjectClass'] : $wrapped->getMetadata()->rootEntityName;
+            $translationMeta = $this->getClassMetadata(); // table inheritance support
 
             $translationClass = isset($config['translationClass']) ?
                 $config['translationClass'] :
--- a/tests/Gedmo/Translatable/Fixture/Issue1123/BaseEntity.php
+++ b/tests/Gedmo/Translatable/Fixture/Issue1123/BaseEntity.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Translatable\Fixture\Issue1123;
+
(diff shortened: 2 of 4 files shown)


lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
```diff
@@ -117,13 +117,13 @@ public function findTranslations($entity)
...
         if ($wrapped->hasValidIdentifier()) {
             $entityId = $wrapped->getIdentifier();
-            $entityClass = $wrapped->getMetadata()->rootEntityName;
-            $translationMeta = $this->getClassMetadata(); // table inheritance support
-
             $config = $this
                 ->getTranslatableListener()
                 ->getConfiguration($this->_em, get_class($entity));
 
+            $entityClass = isset($config['useObjectClass']) ? $config['useObjectClass'] : $wrapped->getMetadata()->rootEntityName;
```
l3pp4rd: `$config['useObjectClass']` is always set by extensions, so there is no need to check with isset. so you can change it to `$entityClass = $config['useObjectClass'];`

l3pp4rd:
thanks, if it solves the problem, that is great, thought that might be more complicated than that.. just update based on the comment

lib/Gedmo/Translatable/Entity/Repository/TranslationRepository.php
CraigMason: I did originally, but then Issue84Test.php was failing saying 'useObjectClass' is not set.

l3pp4rd: hm, I will have a look

l3pp4rd:
should be updated, there was an issue in case if proxy was passed instead of entity

CraigMason:
Thanks for the quick input. We've switched our fork back already!

PR #1506, opened 2016-01-14
Update uploadable.md

gherkins:
minor fixes in code example
defined $name property, return $path property instead of undefined var


l3pp4rd:
thanks ;)

PR #1509, opened 2016-01-20
Force column names with underscore for Translatable

TataneInYourFace:


l3pp4rd:
hi, you must use [underscore naming strategy](http://stackoverflow.com/questions/12702657/how-to-configure-naming-strategy-in-doctrine-2) in that case throughout your project

TataneInYourFace:
Oh thank you ! Very good solution which I did not know.

PR #1514, opened 2016-01-27
[Tree] Ability to create relation to root node

aramalipoor:
I'm not sure if this is the right way, but in SyliusTaxonomyBundle we need to create a relation to root node.

```xml
<many-to-one field="root" target-entity="AppBundle\Entity\Taxon">
  <join-column name="tree_root" referenced-column-name="id" nullable="true" />
  <gedmo:tree-root />
</many-to-one>
```

### DIFF
--- a/doc/tree.md
+++ b/doc/tree.md
@@ -149,7 +149,8 @@ class Category
 
     /**
      * @Gedmo\TreeRoot
-     * @ORM\Column(name="root", type="integer", nullable=true)
+     * @ORM\ManyToOne(targetEntity="Category")
+     * @ORM\JoinColumn(name="tree_root", referencedColumnName="id", onDelete="CASCADE")
      */
     private $root;
 
@@ -181,6 +182,11 @@ class Category
         return $this->title;
     }
 
+    public function getRoot()
+    {
+        return $this->root;
+    }
+
     public function setParent(Category $parent = null)
     {
         $this->parent = $parent;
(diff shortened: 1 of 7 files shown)


l3pp4rd:
well, that is a great addition, thanks. could you also mention this in **doc/tree** also if you have a little more time for a unit test case, would be good to ensure stability.

aramalipoor:
@l3pp4rd Done ;)

l3pp4rd:
everything else looks awesome, great work ;)

aramalipoor:
Oops sorry my bad :p Done ;)

aramalipoor:
@l3pp4rd When will you release 2.4.12 ?

l3pp4rd:
@aramalipoor it will go with **2.5.0** but probably it is worth to cherry pick it for 2.4 as well. I will check with a release when I can, should be tomorrow..

aramalipoor:
@l3pp4rd Awesome, thanks.

l3pp4rd:
2.4.11 was released with this change

aramalipoor:
@l3pp4rd Boom! :tada: Perfect!!

aramalipoor:
:( Seems like it wasn't that easy to do! We're seeing fails in Sylius tests, which is not covered in tests here. The problem is that Tree extension is setting `rootId` (int) on `root` property in various places (which is not right when root is an association) and this is causing Doctrine exceptions.

I'll try to understand how affected parts work and come up with a solution. Can you give me some directions around this issue?

l3pp4rd:
I think in that case, we need to go through the source, and see where the rootId (int) is used explicitly and change query builder to use a root property value instead. that should solve the problem without much of the effort. If you won't be able to find all places involving this, I will help you with that later ;)

aramalipoor:
@l3pp4rd I've prepared a PR on #1515

PR #1515, opened 2016-01-28
[Tree] Prevent setting integer value (root id) on an associated root property

aramalipoor:


l3pp4rd:
looks good, also there may be places in [repository](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Traits/Repository/ORM/NestedTreeRepositoryTrait.php)

aramalipoor:
I don't see any special case in repository ;)

l3pp4rd:
indeed, ok, will need to make another release..

aramalipoor:
@l3pp4rd Yeah, thanks, that would be great :D

PR #1519, opened 2016-02-08
Soft delete objects that are marked to expire in the future

c0nstantx:
When a time aware object with future deletion date is deleted, then the deletedAt date will change to "now". If a second delete is attempted, then it will be hard deleted.
This solves the issue when an object is marked to expire in the future, but we want to soft delete it before it expires.


l3pp4rd:
Hi, makes sense indeed, thanks ;)

PR #1522, opened 2016-02-14
micro-optimise is_null

goatherd:
Micro-optimization to replace `is_null($x)` with `null === $x`. Removed a fragment where a `if`-statement was repeated.

Not that either is slow but the function call is about 14 times slower adding usually adding up over a request.


l3pp4rd:
thanks ;)

Issue #1006, opened 2014-03-11
[Translatable][MongoDB] Error creating or updating translation when field is null


draco1989:
When a Translatable annotated field is set as empty, TranslatableListener processes it and updates the Translation Document with content property as null. When Mongo saves it, omits content field.
When trying to read that field, there is no check that content property is empty, so it drops an error.

ContextErrorException: Notice: Undefined index: content in **/vendor/gedmo/doctrine-extensions/lib/Gedmo/Translatable/TranslatableListener.php line 464

goatherd:
:+1: can reproduce this with recent 2.4.0 tag.

My Test-Case:

- config: fallback defined to store in document
- 1. add fallback-translation
- 2. add translation for language X
- 3. drop translation for language X
- now the document causes the Exception mentioned above whenever the translation for language X is requested

This however is some trouble as there should be no longer any UI accessible to recover from that - beside removing the document maybe.

It is somewhat tricky to replicate as the default / fallback-language is not effected. Same holds for required form fields as they can not be emptied once set.

goatherd:
It is possible to re-produce with persist_default_translation enabled.

If the translation value is set first and set to `null` later the object is no longer possible to load with the translation.

Translation looks like

```
{
    "_id" : ObjectId("56c0dc9bc99d8cd1118b4568"),
    "locale" : "en",
    "objectClass" : "MyBundle\Document\MyDocument",
    "field" : "description",
    "foreign_key" : "55e7faac89716ed77c8b456f"
}
```

After reviewing the listener the only case for translation deletion is when the related object is removed. This however is not covering optional fields.

This can never effect document removal or insert, only update to `null` for any previously persisted translation.

The problem lies with https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.12/lib/Gedmo/Translatable/TranslatableListener.php#L634 where `$content === null` is allowed for updating / inserting as translation.

This however is not handled by the translation retrieval causing the reported exception.

As for me it is unintuitive to set a null-record for the dropped translation I'd rather have it remove the translation completely if null'ed.

goatherd:
The current position would likely be https://github.com/Atlantic18/DoctrineExtensions/blob/39c925d393a424e5d6da5fbf2808f87657717c8b/lib/Gedmo/Translatable/TranslatableListener.php#L511

goatherd:
To summarise my opinion on this:

[...]

PR #1523, opened 2016-02-14
#1006 drop null'ed translation

goatherd:
Might require further testing (and require UnitTests).

Current state: hot-fixed null'ed translation for standard-storage type - including fallback-translation case.

Untested is the personal translation.

Likely to match all cases it would be best to abstract the removal to the adapter or maybe inject the removal at an earlier point of execution. The current approach however is minimal and does not touch fallback-translations behaviour.

May add a UnitTest in march (please remind me if needed).

### DIFF
--- a/lib/Gedmo/Translatable/TranslatableListener.php
+++ b/lib/Gedmo/Translatable/TranslatableListener.php
@@ -667,8 +667,7 @@ class TranslatableListener extends MappedEventSubscriber
                 // check if need to update in database
                 $transWrapper = AbstractWrapper::wrap($translation, $om);
                 if ($content === null && !$isInsert) {
-                    // if the $content is null'ed remove the full translation to avoid fragments
-                    $om->clear($translation);
+                    // if the $content is null'ed remove the translation to avoid fragments
                     $om->remove($translation);
                 } elseif ((is_bool($content) || is_int($content) || (is_string($content) && strlen($content) > 0) || !empty($content)) && ($isInsert || !$transWrapper->getIdentifier() || isset($changeSet[$field]))) {
                     if ($isInsert && !$objectId && !$ea->usesPersonalTranslation($translationClass)) {
--- a/lib/Gedmo/Translatable/TranslatableListener.php
+++ b/lib/Gedmo/Translatable/TranslatableListener.php
@@ -666,7 +666,11 @@ class TranslatableListener extends MappedEventSubscriber
                 $translation->setContent($content);
                 // check if need to update in database
                 $transWrapper = AbstractWrapper::wrap($translation, $om);
-                if (((is_null($content) && !$isInsert) || is_bool($content) || is_int($content) || (is_string($content) && strlen($content) > 0) || !empty($content)) && ($isInsert || !$transWrapper->getIdentifier() || isset($changeSet[$field]))) {
+                if ($content === null && !$isInsert) {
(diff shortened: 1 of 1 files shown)


goatherd:
There are two edge-cases:

- unsetting the translation through `$om->getRepository('Gedmo\\Translatable\\Entity\\Translation')->translate(..);` is **not** handled (it currently fails!)
- unsetting a translation might not trigger when the fallback-translation not set (or null) the UnitOfWork will consider the field unchanged and not trigger the listener

goatherd:
Considering the related issues it would seem that this is rather a feature-request to allow translation removal by language and the behaviour currently is not defined.

It is kind of  a problem that the troublesome exception is not documented otherwise it is not exactly documented that unsetting translations should work.

lib/Gedmo/Translatable/TranslatableListener.php
```diff
@@ -666,7 +666,11 @@ private function handleTranslatableObjectUpdate(TranslatableAdapter $ea, $object
                 $translation->setContent($content);
                 // check if need to update in database
                 $transWrapper = AbstractWrapper::wrap($translation, $om);
-                if (((is_null($content) && !$isInsert) || is_bool($content) || is_int($content) || (is_string($content) && strlen($content) > 0) || !empty($content)) && ($isInsert || !$transWrapper->getIdentifier() || isset($changeSet[$field]))) {
+                if ($content === null && !$isInsert) {
+                    // if the $content is null'ed remove the full translation to avoid fragments
+                    $om->clear($translation);
```
alcaeus: Calling `clear` here won't work. First of all (at least for ODM) it takes a document name, not a document itself. Second of all, when the document has been cleared from the object manager, the call to `remove` won't do a thing since the document in question is not managed.

goatherd: Thanks.

l3pp4rd:
removing translation should be users responsibility, it is a part of general migration. Null if field is nullable is also a valid translation. this looks like a magic feature, which would just confuse people, cannot accept that.

goatherd:
@l3pp4rd Okay.

Null is not a feasible translation with ODM and the current handler as the field is dropped from the Translation-Collection (#1006).

Got confused when a customer made his app unreachable this way for one language and only realised removal (null-setting) is not intended after the pull-request was completed.
#1007 tries to implement the same logic by the way and is pending for some time.

Would a proper feature-pull-request for configurable on-null-remove be acceptable? It is somewhat hard to patch the listener without a full copy due its many private methods.

Or would a fix for the missing translation content field in ODM be better? Maybe I did look at it the wrong way altogether. 

An undefined behaviour remains for the null-translation in regard to fallback-translations: should the value be 'null' or the fallback?

l3pp4rd:
In this case add a listener which ensures that field is not nullable or removed. Extensions are mainly focused on ORM and different changes to ODM behavior won't be applied to satisfy 5% of users, they have enough options to prevent that. 

- create a listener and hook it before the translatable
- remove every translation which is null. (you can read listener configuration easily)

Though, ODM does not support php 7 still, which also says the community is very fragile, I would not invest much into it. Using a tool like ODM in general looks a little off topic to me, sqlite is a way better database, which probably you do not yet scale anyway..

goatherd:
That is a definite response. Thanks!

I understand your point too well, if only I could decide the storage more often.

PR #1525, opened 2016-02-15
handle content the same for ODM and ORM

goatherd:
Sorry to bother again regarding #1006.

I deeply analysed the differences in ODM and ORM behaviour on the `null` content and only found a single position where the key does not exist for ODM.

Thus a fix not effecting ORM-behaviour but avoiding the exception on ODM. Verified as sufficient for my use-case but to trivial to provide a unit-test.

Kindly asking you to consider this please on behalf of my customer..


goatherd:
Thank you!

disrupticons:
@l3pp4rd When will this fix be released?

Issue #1501, opened 2016-01-11
Combination of Sortable and Soft-deletable


pix-art:
Hi all,

I'm having an issue with a combination of Soft-Deletable and Sortable extensions. Can these 2 be used together? And how do i resolve this issue? 

Example:

Item 1 - Position 0
Item 2 - Position 1
Item 3 - Position 2
Item 4 - Position 3

**Delete Item 2**

Item 1 - Position 0
*Item 2 - Position 1 - Deleted at xxx*
Item 3 - Position 2
Item 4 - Position 3

List of item positions shows: 0 - 2 - 3 

Greets,
Joeri

pix-art:
While writing my issue, I've just realised i shouldn't use the position value on my frontend and just print the index of the item on the frontend. 

For example:

What i used to do:

````
{% for item in items %}
    Position: {{ item.position }}
{% endfor %}
````

What i should do:
````
{% for item in items %}
    Position: {{ loop.index }}
{% endfor %}
````

So this does work correctly and the sortable recalculation keeps working even with soft-deleted items.

creatiombe:
The class Gedmo\Sortable\Mapping\Event\Adapter\ORM should detect if there is a SoftDeleteable filter attached to the same entity to solve this problem.  The queries should be adjusted here for softdeleteable.

pix-art:
I seem to have been a bit to hasty thinking this was a solution to my problem, yes this solves the frontend part but backend this can still provide us with some issues.

Example:

Item 1 - Position DB: 0 Front: 1
Item 2 - Position DB: 1 Front: 2
Item 3 - Position DB: 2 Front: 3
Item 4 - Position DB: 3 Front: 4

**Delete Item 2**

Item 1 - Position DB: 0 Front: 1 
Item 2 - Position DB: 1 Deleted at xxx
Item 3 - Position DB: 2 Front: 2
Item 4 - Position DB: 3 Front: 3

**Move Item 2 to position 2 (Front)**

Item 1 - Position DB: 1 Front: 1 
Item 2 - Position DB: 0 Deleted at xxx
Item 3 - Position DB: 2 Front: 2
Item 4 - Position DB: 3 Front: 3

As you can see putting it on position 2 makes it stay on position 1 because there is a hidden item it has to jump over in the database.

pix-art:
I was wondering, if a combination of these 2 extensions would need a listener to see if an entity also has a position, and if so on delete move it to the bottom of the list?

[...]

PR #1526, opened 2016-02-16
#1501 : Docs updated with example of how to solve sortable/softdelete…

pix-art:
https://github.com/Atlantic18/DoctrineExtensions/issues/1501

/cc @l3pp4rd


l3pp4rd:
thanks ;)

PR #1531, opened 2016-02-22
accept any phpunit release; don't require mock-objects

narcoticfresh:
Hi guys..

Currently, we cannot switch to a recent `phpunit/phpunit` release because of the notation of `phpunit/phpunit-mock-objects` as a dependency in *this* repo blocks us.

**Important:** It is *not* necessary to specify `phpunit/phpunit-mock-objects` as a dependency as it's already a requirement of phpunit itself!


l3pp4rd:
thanks ;)

Issue #1137, opened 2014-08-11
[Blameable] Integer not a valid type mapping


Maff-:
I'm trying to use the Blameable behavior in a Symfony2 application (via Stof..Bundle). But because my User object is not an entity managed by Doctrine I'v implemented my own listener that set the user value with the current user id (int).

Now the problem occurs when `Gedmo\Blameable\Mapping\Driver\Annotation::isValidField()` is called. As `$validTypes` = `['one', 'string', 'int']`. This seems strange, as the type doctrine uses for integers is **integer** not **int** [ [1](http://doctrine-dbal.readthedocs.org/en/latest/reference/types.html#integer) ].

Should we update the Blameable mapping drivers to use `['one', 'string', 'integer']` as validTypes?

``` php
trait UserBlameableTrait
{
    /**
     * @var int
     *
     * @Gedmo\Blameable(on="create")
     * @ORM\Column(name="entered_by_user_id", type="integer", nullable=true)
     */
    private $enteredByUserId;
}
```

AkenRoberts:
Closing due to age and lack of activity. If you are interested in this issue or are still experiencing any problems described here, please add a comment!

PR #1533, opened 2016-02-22
Added blameable integer field support

Padam87:
- [x] Code
- [x] Tests
- [x] Docs

Related issue: #1137

Use case:
- User is an entity
- Comment is a document

Mapping: 
```php
    /**
     * @var int
     *
     * @Gedmo\Blameable(on="create")
     * @MongoDB\Int()
     */
    protected $createdById;
```

It works conviniently with references as well:

```php
    /**
     * @var User
     *
     * @Gedmo\ReferenceOne(type="entity", class=User::class, identifier="createdById")
     */
    protected $createdBy;

    /**
     * @var int
     *
     * @Gedmo\Blameable(on="create")
     * @MongoDB\Int()
     */
    protected $createdById;
```

**Will this be merged if I work on it?**

### DIFF
--- a/doc/blameable.md
+++ b/doc/blameable.md
@@ -643,3 +643,36 @@ class UsingTrait
 The Traits are very simplistic - if you use different field names it is recommended to simply create your
 own Traits specific to your project. The ones provided by this bundle can be used as example.
 
+## ORM - ODM Hybrid databases
+
+You can use blameable with hybrid databases. If your user is an entity, but your blameable is a document, you can use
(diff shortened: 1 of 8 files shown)


Padam87:
@l3pp4rd What do you think?

l3pp4rd:
hi @Padam87 yes it is fine, but please add some tests for this method which gets the user value, it grows complex :)

Padam87:
Great! I will add the tests and docs tomorrow.

PR #1535, opened 2016-02-24
#1501 : Docs updated with working example

pix-art:
I did a PR with updated documentation but made a mistake in the example.


l3pp4rd:
by the way is the flush needed? because this event already runs inside the flush

pix-art:
If i don't flush it it doesn't seem to send the position through, so i flush it before i "flush" the remove event.

l3pp4rd:
OK, if so, figured it was not necessary ;)

PR #1537, opened 2016-02-25
Remove excessive fields

aivus:
Relations for `root` field was introduced in https://github.com/Atlantic18/DoctrineExtensions/pull/1514, but scalar field was not removed.


l3pp4rd:
indeed, thanks ;)

Issue #1530, opened 2016-02-21
[Uploadable] UploadableFileName is not correct when used with appendNumber: true


vrbata:
Hi,

in my yaml file I use Uploadable like this:

```
gedmo:
    uploadable:
      allowOverwrite: false
      appendNumber: true
      pathMethod: path
      filenameGenerator: ALPHANUMERIC

fields:
    name:
      type: string
      gedmo:
        - uploadableFileName
```

When I upload file test.jpg for the first time I have in path specified by pathMethod file called test.jpg. But when I try to upload same file again, path is correctly set  as "path/to/file/test-2.jpg" but name stored in database by uploadableFileName is still "test.jpg". Shouldnt it be "test-2.jpg"?

I think problem is in Gedmo\Uploadable\UploadableListener in moveFile method when  generated filename is set to $info['fileName'] but before $appendNumber logic is used.

When used like this I cant predict correct filename because i have different filename in file system and in database.

If I use uploadableFilePath instead of uploadableFileName everything is ok, as in $info['filePath'] the filename is generated correctly.

Ronan-Lenor:
+1

i have the same problem.
and it became a VERY BIG problem in my devellopment.

vrbata:
I am working on it. I will hopefully send PR by end of this week.

l3pp4rd:
that would be great @vrbata since the original maintainer of uploadable is not reachable ;)

vrbata:
@l3pp4rd [PR #1540](https://github.com/Atlantic18/DoctrineExtensions/pull/1540) sent as promised

l3pp4rd:
great, thanks, was merged a possible BC break will be mentioned in readme, but it is a good timing since the 2.5.0 release is going to be the next, with some BC breaks.

alcaeus:
> the 2.5.0 release is going to be the next, with some BC breaks.

Might I suggest adhering to [Semantic Versioning](http://semver.org/spec/v2.0.0.html) and making the next release 3.0.0 if the release contains BC breaks? That would prevent lots of people using this library some headache. For example: The `StofDoctrineExtensionsBundle` (which I assume many people using this library in a Symfony project use) specifies the requirement as `^2.3.1`, which makes a 2.5 release eligible when running `composer update`. Since it contains BC breaks, people might be upgrading and finding their projects broken, which is exactly what SemVer is trying to solve.

Padam87:
[...]

PR #1540, opened 2016-02-28
[Uploadable] UploadableFileName is not correct when used with appendNumber: true

vrbata:
Solving issue: [#1530](https://github.com/Atlantic18/DoctrineExtensions/issues/1530)

$info from **UploadableListener::moveFile** contains wrong data about file if used with appendNumber option.

Data before update:
```
$info['fileName'] == 'test.txt';
$info['fileWithoutExt'] == 'SOME/PATH/test';
$info['filePath'] == 'SOME/PATH/test-2.txt';
```

As you can see only filePath contains correct filename, when used with UploadableFileName annotation file is stored as SOME/PATH/test-2.txt in file system but as test.txt to database.
Same data are used for [callback option](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/uploadable.md#uploadable-annotations)

I changed this information so they reflect correct generated filename like this:
```
$info['fileName'] == 'test-2.txt';
$info['fileWithoutExt'] == 'SOME/PATH/test-2';
$info['filePath'] == 'SOME/PATH/test-2.txt';
```
So data received by callback method and then returned from UploadableListener::moveFile contains correct info.

**This can be possible BC break but only if used with appendNumber option** but without it you cant use UploadableFileName mapping with appendNumber option.

### DIFF
--- a/lib/Gedmo/Uploadable/UploadableListener.php
+++ b/lib/Gedmo/Uploadable/UploadableListener.php
@@ -553,12 +553,8 @@ class UploadableListener extends MappedEventSubscriber
             );
             $info['fileName'] = $filename;
 
-            if ($pos = strrpos($info['filePath'], '.')) {
-                // ignores positions like "./file" at 0 see #915
-                $info['fileWithoutExt'] = substr($info['filePath'], 0, $pos);
-            } else {
-                $info['fileWithoutExt'] = $info['filePath'];
-            }
+            $pathInfo = pathinfo($info['filePath']);
+            $info['fileWithoutExt'] = $pathInfo['dirname'] . '/' . $pathInfo['filename'];
         }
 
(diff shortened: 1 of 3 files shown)


l3pp4rd:
thanks ;)

mulraj208:
array (size=7)
  'fileName' => string 'faq.png' (length=7)
  'fileExtension' => string '.png' (length=4)
  'fileWithoutExt' => string '../uploads/themes/backgrounds/faq' (length=33)
  'origFileName' => string 'faq.png' (length=7)
  'filePath' => string '../uploads/themes/backgrounds/faq-5.png' (length=39)
  'fileMimeType' => string 'image/png' (length=9)
  'fileSize' => int 890261

File Name and File path are different.
File name stored in database is faq,png where as path is different

PR #1543, opened 2016-03-01
[Uploadable Events] getAction() method returns INSERT and not CREATE.

Alsatian67:
Corrected the description to be compatible with the listener :

https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Uploadable/UploadableListener.php#L39


l3pp4rd:
thanks ;)

PR #1544, opened 2016-03-02
Hot fix for issue #370

ksn135:
It works in my case, tests required


l3pp4rd:
this branch is not maintained

PR #1546, opened 2016-03-03
 [Sortable] ORM Custom Types 

### LINKED ISSUE #1545: [Sortable] ORM Custom Types
We have a one to many assoctions.
One Person has many Cars.
We want to sort Cars by Person.
Everything works fine.
Now, the Person became a custom doctrine type as id.
The get MaxPosition query is broken, cause it doesn't resolve the type.

Gladhon:
We have a one to many assoctions.
One Person has many Cars.
We want to sort Cars by Person ( by group ).
Everything works fine.
Now, the Person became a custom doctrine type as id.
The get MaxPosition query is broken, cause it doesn't resolve the custom type.

Fix #1545

### DIFF
--- a/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
+++ b/lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
@@ -2,9 +2,10 @@
 
 namespace Gedmo\Sortable\Mapping\Event\Adapter;
 
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ORM\QueryBuilder;
 use Gedmo\Mapping\Event\Adapter\ORM as BaseAdapterORM;
 use Gedmo\Sortable\Mapping\Event\SortableAdapter;
-use Doctrine\ORM\QueryBuilder;
 
 /**
  * Doctrine event adapter for ORM adapted
@@ -20,8 +21,8 @@ final class ORM extends BaseAdapterORM implements SortableAdapter
         $em = $this->getObjectManager();
 
         $qb = $em->createQueryBuilder();
-        $qb->select('MAX(n.'.$config['position'].')')
-           ->from($config['useObjectClass'], 'n');
+        $qb->select('MAX(n.' . $config['position'] . ')')
+            ->from($config['useObjectClass'], 'n');
         $this->addGroupWhere($qb, $groups);
         $query = $qb->getQuery();
         $query->useQueryCache(false);
@@ -36,13 +37,22 @@ final class ORM extends BaseAdapterORM implements SortableAdapter
         $i = 1;
         foreach ($groups as $group => $value) {
             if (null === $value) {
-                $qb->andWhere($qb->expr()->isNull('n.'.$group));
+                $qb->andWhere($qb->expr()->isNull('n.' . $group));
             } else {
-                $qb->andWhere('n.'.$group.' = :group__'.$i);
-                $qb->setParameter('group__'.$i, $this->getGroupValue($value), $this->getGroupType($value));
+                $qb->andWhere('n.' . $group . ' = :group__' . $i);
+                $qb->setParameter('group__' . $i, $this->getGroupValue($value), $this->getGroupType($value));
             }
             $i++;
         }
+    }
+
+    /**
+     * @param $value
+     * @return bool
+     */
+    private function isEntity($value)
+    {
+        return (is_object($value) && $this->getObjectManager()->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value)));
     }
 
     /**
@@ -51,7 +61,7 @@ final class ORM extends BaseAdapterORM implements SortableAdapter
      */
     private function getGroupValue($value)
     {
-        if (!is_object($value)) {
+        if (!$this->isEntity($value)) {
             return $value;
         }
         return $this->getObjectManager()->getUnitOfWork()->getSingleIdentifierValue($value);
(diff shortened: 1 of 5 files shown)


Padam87:
"Now, the Person became a custom doctrine type as id."

If I understand it correctly your ID is a custom type?

Is that a good idea? I would image that doctrine has a lot a problems resolving the id. You would have to specify the type in every query builder... you will have a lot of problems with 3rd party libs.

Gladhon:
Yes and No. Doctrine it self should have problems with that. 3rd party libs could have problems with it, like this one, but is there any reason to not support doctrine custom types ?

Padam87:
In general, custom, and composit primary keys are a bad practice, but let's put that aside for now.

Your PR assumes that every object that is passed as a group is an entity, with a single primary key, and that is why the test are failing... Objects can be \DateTime, or even a custom value object for a custom type, so you must handle those cases as well.

lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
```diff
@@ -39,12 +39,38 @@ private function addGroupWhere(QueryBuilder $qb, $groups)
...
         }
     }
 
+    /**
+     * @param $value
+     * @return mixed
+     */
+    private function getGroupValue($value)
+    {
+        if (!is_object($value)) {
```
Padam87: is_object check is not enough, you will have to create an `isEntity` function to check that the object is a mapped entity.

lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
```diff
@@ -39,12 +39,38 @@ private function addGroupWhere(QueryBuilder $qb, $groups)
...
+            return $value;
+        }
+        return $this->getObjectManager()->getUnitOfWork()->getSingleIdentifierValue($value);
+    }
+
+    /**
+     * @param $value
+     * @return \Doctrine\DBAL\Types\Type|null|string
+     */
+    private function getGroupType($value)
```
Padam87: `$group` should be passed as well, and the metadata should be fetched like this:
- if the object is a simple value object, `$metaData->getTypeOfField($group)`
- if the object is an entity, your code is ok

l3pp4rd:
Doctrine is automatically parsing the **identifier** if you pass in the entity in query. It also manages scalar type transformations. What you need is to register a DBAL type for your custom value, because extensions will handle it only the standard way (otherwise there would be many corner cases). Custom types - custom transformations which you need to handle yourself in this case, or open an issue in doctrine ORM because it should extract id from an entity (but though you still need a DBAL type)

Gladhon:
@l3pp4rd 
the custom dbal type is of course registered, but this will not work on query builder.

l3pp4rd:
have you analyzed why exactly it does not work? I think first we need to know the reason

Gladhon:
well i can not see where in doctrine the type should be handled autoamtically - can you show me that

l3pp4rd:
look how it maps [query parameters](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query.php#L321) to dbal types

Gladhon:
I see, it use ParameterTypeInferer::inferType($value) to get undefinied types , which guess type by value, and not support any custom types. Do you think there have to  custom types supported also ?

l3pp4rd:
well, you could ask this question in ORM issue tracker, because we just can't make our own type guessing implementation. it should be compatible with ORM. you could debug more how your custom parameter is processed and see exactly whether you can adapt it or not

Gladhon:
I would ask that.
On the other hand, setParameter is designed to became the type of the parameter. And i think reading type from classMetadata is not incompatible and also no guessing.

l3pp4rd:
if it reads successfully from the metadata, then you should not have this problem, or am I missing something?

Gladhon:
Yes, if it reads successfully from the metadata, then i should not have this problem.

Gladhon:
@Padam87 
I have no a better check if it's an entity.
I'im not using $metaData->getTypeOfField($group) right now, cause it doesn't have anything to do with my problem. And i dont  likte to change the bahavior from bevore accidentally. But if you wish, i can add that additional

l3pp4rd:
@Gladhon please create a test case first, without it, we cannot merge and find the best solution.

Gladhon:
i dont know where and how to write test the Gedmo\Sortable\Mapping\Event\Adapter\ORM class. I don't find any tests for that adapter. 
I could just write a unit test, that mocked doctrine classes will be called, but it don't seems to make sense

l3pp4rd:
create functional test where fixtures has custom type **Id** like [SortableGroupTest](https://github.com/Atlantic18/DoctrineExtensions/blob/master/tests/Gedmo/Sortable/SortableGroupTest.php)

lib/Gedmo/Sortable/Mapping/Event/Adapter/ORM.php
```diff
@@ -36,15 +37,54 @@ private function addGroupWhere(QueryBuilder $qb, $groups)
...
+    {
+        if (!$this->isEntity($value)) {
+            return null;
```
Padam87: Please return $metaData->getTypeOfField($group) here ;)

Like I said, custom type value objects are also a possibility. Infer type will not be able to resolve those, and the sortable won't work.

Gladhon: added

l3pp4rd:
thanks @Gladhon I'm very busy next week, but I'll find some time to verify and check whether there are better ways to fix the issue or not.

Gladhon:
@l3pp4rd  no problem, best would be a doctrine ORM internal solution, or a helper from doctrine ORM. 

On the other hand, merge this is in, would not prevent anybody to implement a better way later, it should be easier, cause you have at least a test for it.

l3pp4rd:
I'm not sure whether it will be backward compatible, that is mainly why I need to have a look

Gladhon:
Mainly $metaData->getTypeOfField would be used instadof ParameterTypeInferer::inferType($value) for group-id's. I would say, this is not a BC, as nobody should count on that internals. But I will wait for your opinion.

PR #1547, opened 2016-03-04
[Tree] Closure cannot handle different EntityManagers

Slamdunk:
Hi, I think the test is pretty self-explanatory about the bug.

The real use case is everywhere you work with multiple EntityManagers, like UnitTest, where the [EntityManager::create](https://github.com/doctrine/doctrine2/blob/v2.5.4/lib/Doctrine/ORM/EntityManager.php#L827) function is heavily used to ensure isolation between EntityManager instances with a good performance.

I don't know if the other 2 variables, [$pendingNodeUpdates](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.13/lib/Gedmo/Tree/Strategy/ORM/Closure.php#L49) and [$pendingNodesLevelProcess](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.13/lib/Gedmo/Tree/Strategy/ORM/Closure.php#L57) should need the same fix, but I was not able to create a test to find it out, so I prefered to leave them untouched.

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/Closure.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/Closure.php
@@ -170,7 +170,7 @@ class Closure implements Strategy
      */
     public function processPrePersist($em, $node)
     {
-        $this->pendingChildNodeInserts[spl_object_hash($node)] = $node;
+        $this->pendingChildNodeInserts[spl_object_hash($em)][spl_object_hash($node)] = $node;
     }
 
     /**
@@ -237,8 +237,9 @@ class Closure implements Strategy
     public function processPostPersist($em, $entity, AdapterInterface $ea)
     {
         $uow = $em->getUnitOfWork();
+        $emHash = spl_object_hash($em);
 
-        while ($node = array_shift($this->pendingChildNodeInserts)) {
+        while ($node = array_shift($this->pendingChildNodeInserts[$emHash])) {
             $meta = $em->getClassMetadata(get_class($node));
             $config = $this->listener->getConfiguration($em, $meta->name);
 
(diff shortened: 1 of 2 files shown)


l3pp4rd:
I think extensions at all do not support multiple entity managers which has the same listener instance in event manager. You should use different listener instances for them. Otherwise it would be much more code to adapt which I do not see the point in doing. so create a different instance of listener for each entity manager. it could mention it in documentation though

Slamdunk:
@l3pp4rd Doctrine made the the Observer pattern for listeners to be used in a stateless way. Some of this project lines are an example of this: https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.13/lib/Gedmo/Tree/TreeListener.php#L102-L104
In the idea of the ORM, only the UOW should be statefull. EntityManagers, Connections, Configuration should all be stateless when the work with the domain Entities, and so should be the EventManager. A proof is that in the entire doctrine projects, only the UOW implements a `clear` method. The fact that the TreeListener does not subscribe the `onClear`, indicates that there is no state, and in fact I think there could be a bug about this.

So, I feel confident that, for the same domain entities, the same metadatas, the same connection, and the same configuration, to use a different EntityManager instantiated with the same arguments.
The performances thank this: if in every `MyTest::setUp` I should have to inizialite from scratch the EntityManager, my tests would be ~10 times slower.

Theory apart, if I provide to the project a test, a bugfix, a use-case and nothing else is broke, why reject a contribution?

l3pp4rd:
it won't be slower if you create new event listeners. I agree it should follow ORM ideology, but currently it does not, the simplest option is to make new listener instances. all extensions should be adapted, not only a closure tree in order to close this issue.
the reason why no one has opened this issue before, is probably that they just created new listeners.

PR #1559, opened 2016-03-30
[git] Add .gitattributes to remove unneeded files

Slamdunk:
Copying what was done in all ZF modules, there are asome files not needed in the package built and downloaded from composer.


l3pp4rd:
thanks ;)

garak:
It looks like `.gitattributes` was eventually deleted, I can't find it anymore in this repo :disappointed:

PR #1570, opened 2016-04-19
Fixed listener name collision

### LINKED ISSUE #1569: [Bug] Missing parent::__construct on listeners
Some listeners, like the `uploadable` and `references` listener miss calling the parent constructor. This will cause the `$name` property to remain `null`, and these listeners essentially overwrite each other's config in the cache.

Padam87:
closes #1569


l3pp4rd:
indeed, good catch

PR #1572, opened 2016-04-20
Update Closure.php

durich:
I was having an error when creating a new entity, it's was because the insert part is asumming the column_name is the same name of the entity field/property.

This extra get's the correct name


l3pp4rd:
but you should probably remove the previous line, since it overwrites the variable.

durich:
Hi, 

I didn't remove the previous because it is needed so my new line of code uses it for the last mapping.

I just used another variable name to avoid the override.




Ricardo

l3pp4rd:
ah indeed, missed it :)

l3pp4rd:
thanks

PR #1576, opened 2016-04-25
[Sortable] Fixed DQL error on update

Padam87:
https://github.com/Atlantic18/DoctrineExtensions/issues/370#issuecomment-214493463


Padam87:
Come to think of it, this solution may not be good enough. Sure it solves the problem, but people will have a hard time debugging why the order changed... maybe an exception would be more appropriate here?

l3pp4rd:
looks reasonable. if they do not set position it gets inserted | updated as last. In general they should solve this problem in their code, to prevent sending nulls.

PR #1577, opened 2016-04-26
Translatable: Fix PHP7 error "Cannot use 'String' as class name".

veego:
There is deprecate field type annotation in ODM in favor of Field.

Old:

    /**
    * @MongoODM\String
    */

New:

    /**
    * @MongoODM\Field(type="string")
    */

### DIFF
--- a/lib/Gedmo/Translatable/Document/MappedSuperclass/AbstractPersonalTranslation.php
+++ b/lib/Gedmo/Translatable/Document/MappedSuperclass/AbstractPersonalTranslation.php
@@ -21,7 +21,7 @@ abstract class AbstractPersonalTranslation
     /**
      * @var string $locale
      *
-     * @MongoODM\String
+     * @MongoODM\Field(type="string")
      */
     protected $locale;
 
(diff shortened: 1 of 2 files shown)


l3pp4rd:
thanks ;)

nicodmf:
Very thanks for this modification. @l3pp4rd can we have, please, a version more than use the dev version for this modification?

webdevilopers:
:+1: for merging in **2.4** as suggested in this issue too: https://github.com/stof/StofDoctrineExtensionsBundle/issues/332

spiroski:
Can we have this merged in 2.4.x please? I'm using this extension with php7 and the AbstractTranslation class still uses MongoODM\String

l3pp4rd:
it is in v2.4.23 tag. see the commits and changes update your extensions

spiroski:
I think there's some mistake and it isn't or maybe I'm super confused but:

https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.23/lib/Gedmo/Translatable/Document/MappedSuperclass/AbstractTranslation.php

l3pp4rd:
hm, indeed. could you open a pull request with the Field annotation usage instead?

spiroski:
It has already been merged into master. Should there be new pull request to merge it into v2.4.x branch or how should this be done?

spiroski:
#1683 
This should be it

PR #1583, opened 2016-05-06
Make behat/transliterator optional as needed only with Sluggable

Slamdunk:
behat/transliterator becomes mandatory only with the Sluggable extension.

In all other contexts is just an unused installed component.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,10 +39,10 @@
     },
     "require": {
         "php": ">=5.4",
-        "behat/transliterator": "~1.0",
         "doctrine/common": "~2.4"
     },
     "require-dev": {
+        "behat/transliterator": "~1.0",
         "doctrine/mongodb-odm": ">=1.0.2",
         "doctrine/orm": ">=2.5.0",
         "doctrine/common": ">=2.5.0",
@@ -50,6 +50,7 @@
         "phpunit/phpunit": "*"
     },
     "suggest": {
+        "behat/transliterator": "(~1.0) to use the Sluggable extension",
         "doctrine/mongodb-odm": "to use the extensions with the MongoDB ODM",
         "doctrine/orm": "to use the extensions with the ORM"
     },


l3pp4rd:
yes, but its a library as a whole, it is a hard dependency for sluggable and this cannot apply. sluggable does not ship in a separate repository.

Slamdunk:
This is exactly was I thought, then I grepped `use Doctrine\ORM` into the library and saw 13 extentions and 40 source files with explicit dependancy on an optional component, and then grepped `use Doctrine\ODM\MongoDB` and saw 11 extentions and 25 source files with another explicit dependancy on another optional component.

It's a bit unbalanced, isn't it?

A little thought about composer dependancies: if a library provides more than one functionality, I think every dependancy that is not used across ~~all functionalities~~ many functionalities should be optional and suggested.
Otherwise a single `"require"` key in a user composer.json may explode with tens of depenancies. This is the way many frameowkrs are going, to reduce bloatware while retaining simple component structure and maintainance.
There are not much disadvantages: as soon as a user find a Fatal error due to missing class declaration, he searches through composer's suggetions and explicits the missing component in its composer.json, like we did with this component and doctrine-orm/doctrine-mongodb-odm.

l3pp4rd:
I do not have time to maintain this library. so I'm just trying to keep it stable as much as I can. I'm sure that splitting the library into separate components won't be managed by me and is completely out of the question. unless someone else really starts to maintain this library well.

Slamdunk:
I did not intend to ask you to split the components, "_ain't nobody got time for that_" for the purpose of this library.

I'm asking you: how will a user be affected by setting the transliterator optional, in a different way of a user is already affected by having mongodb-odm optional?

stof:
Moving this as a suggestion will break any project using sluggable (as they will have to require the package themselves).
So this cannot be done in a minor or patch release. anyway

Slamdunk:
@stof this is why this PR is against master

stof:
@Slamdunk as this package provides extensions for the ORM and the ODM, the user will already be using one of them anyway (otherwise it simply does not make sense to use this library). This is why it is not an issue to suggest them only (actually, we could even remove the suggestions as saying "install the ODM to be able to use the extension in your ODM-based code" is useless). And requiring both would be very wrong (it would require ORM users to have the mongo extension installed to be able to install the ODM)

Slamdunk:
Composer was and is the savior for PHP community, the only fault in its adoption was not spreading the habit of making non-mandatory dependancies as optional.

Now to use 10 packages, I have 100 packages installed, and only 20 of them are really in use.

I hope you can understand the burden, and still with the fewer dependancies possible on the upcoming v3.

Just ask, and I'll PR a migration guide.

Slamdunk:
What I'm going to link is not an evidence on who is right or the right practice to follow, but is just a note that someone considers valuable to not require a non-mandatory dependancy, and considers the change a patch (so not a minor nor a major release needed):
https://github.com/zendframework/zend-serializer/pull/16

l3pp4rd:
there is nothing to consider related to transliteration library. most of the users were happy, because this behat transliterator was extracted from extensions and they could use it as a separate component. Since extensions have a hard dependency on it, it must be required, unless sluggable gets out as a separate component - and it will not. ORM and ODM is obvious, you must choose at least one of them to use the library, otherwise it is **useless**. Again, there is nothing to consider.

Slamdunk:
Sluggable has a hard dependency on it. Extensions hah a soft dependency on it.

We are going nowhere.

PR #1584, opened 2016-05-07
[SoftDelete] Use correct column name in MultiTableDeleteExecutor

rodrigm:
I've found an issue in `MultiTableDeleteExecutor` when there is a difference between field and columnName (column name in database) for example with a naming strategy underscore. 

This is because the field is being used instead of the column name. (https://github.com/Atlantic18/DoctrineExtensions/blob/8eae0903abfe8cfac2715018d807637bb92bbfc5/lib/Gedmo/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php#L39)

In detail:

```yml
doctrine:
    orm:
        naming_strategy: doctrine.orm.naming_strategy.underscore

Entity:
    type: entity
    gedmo:
        soft_deleteable:
            field_name: deletedAt
     fields:
     …
     deletedAt:
            type: datetime
            nullable: true
    inheritanceType: JOINED
    discriminatorColumn:
        name: type
```

Thanks!


l3pp4rd:
thanks ;)

PR #1594, opened 2016-05-18
Blameable yaml driver: Add support for mongodb odm association mapping

lopsided:
Current implementation does not work if using yml mapping files with mongodb since the driver is looking for `manyToOne` mapping information. In mongodb these are `referenceOne`.

### DIFF
--- a/lib/Gedmo/Blameable/Mapping/Driver/Yaml.php
+++ b/lib/Gedmo/Blameable/Mapping/Driver/Yaml.php
@@ -73,8 +73,9 @@ class Yaml extends File implements Driver
             }
         }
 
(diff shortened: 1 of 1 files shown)


Padam87:
https://github.com/Atlantic18/DoctrineExtensions/pull/1533

I did a PR to fix this issue, but this is way more easy to use. Please add tests :)

lopsided:
I did want to add tests but I couldn't see anything to base it from - there are no yml tests at the moment. I've had another look but with the amount of work required to add them and with upcoming deadlines I won't be able to add them for a while. Sorry

l3pp4rd:
yes since there aren't many ODM driver tests, it is fine, thanks

PR #1602, opened 2016-05-31
Fix the part about reloading in a different locale ; fix #1600

### LINKED ISSUE #1600: How to define the locale in which to retrieve entities?
I'm trying to do a simple thing, I guess, but I didn't find any help in the documentation. I really wished there was one more paragraph in the section [Basic usage examples](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/translatable.md#basic-usage-examples), telling us how to retrieve the entity in `de_de` locale, because that's exactly what I want to do. This section explains how to set the fields in `en_us`, then how to set the fiels in `de_de`, then how to retrieve the fields in `en_us`. The omission of "how to retrieve the fields in `de_de`" is glaring (even if it is to say that the translations cannot be retrieved via the standard repository methods and give pointers).

I have my entities correctly translated in the ext_translation table. I can use the query hints from [Using ORM query hint](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/translatable.md#using-orm-query-hint) to get the entities in any locale. But when I try to use the `find()` method of my repository, I get the entity in the default locale. I don't know how to change that. I tried setting the locale of my request, but it didn't help.

Alsciende:


l3pp4rd:
thanks ;)

PR #1609, opened 2016-06-07
Fix 1605

### LINKED ISSUE #1605: [Sluggable] Incompatibility with php 7.0.7
PHP 7.0.7 introduced this change: https://bugs.php.net/bug.php?id=72209

This has unveiled a peculiar bug in `Gedmo\Sluggable\SluggableListener::makeUniqueSlug()`: when there are multiple entities with the same `slug` property, at some point in [this cycle](https://github.com/Atlantic18/DoctrineExtensions/blob/f3c474dd9c3dbcf2fcd86075961a4ffad88fd20e/lib/Gedmo/Sluggable/SluggableListener.php#L436-L444) the wrong metadata is used, so a reflection of a different entity object is used, and this now results in an error.

I was able to write a patch to fix the issue:

```diff
diff --git a/lib/Gedmo/Sluggable/SluggableListener.php b/lib/Gedmo/Sluggable/SluggableListener.php
index e4af702..7858617 100644
--- a/lib/Gedmo/Sluggable/SluggableListener.php
+++ b/lib/Gedmo/Sluggable/SluggableListener.php
@@ -434,10 +434,12 @@ class SluggableListener extends MappedEventSubscriber
         // collect similar persisted slugs during this flush
         if (isset($this->persisted[$class = $ea->getRootObjectClass($meta)])) {
             foreach ($this->persisted[$class] as $obj) {
-                if ($base !== false && $meta->getReflectionProperty($config['unique_base'])->getValue($obj) !== $base) {
+                $similarMeta = $om->getClassMetadata(get_class($obj));
+                if ($base !== false && $similarMeta->getReflectionProperty($config['unique_base'])->getValue($obj) !== $base) {
                     continue; // if unique_base field is not the same, do not take slug as similar
                 }
-                $slug = $meta->getReflectionProperty($config['slug'])->getValue($obj);
+                $slugRefl = $similarMeta->getReflectionProperty($config['slug']);
+                $slug = $slugRefl->getValue($obj);
                 $quotedPreferredSlug = preg_quote($preferredSlug);
                 if (preg_match("@^{$quotedPreferredSlug}.*@smi", $slug)) {
                     $similarPersisted[] = array($config['slug'] => $slug);
```

but honestly I'm a bit lost on writing a unit test for this; I'm not familiar with the codebase, there are no contributing guidelines, and as far as I've seen there are mostly functional test, so I could use some pointers to put together a proper PR.

stefanotorresi:
This PR fixes #1605.

As explained by the test, the bug is reproduced by having two different entities using joined inheritance and both map the slug under the same property (but not inheriting it).

### DIFF
--- a/lib/Gedmo/Sluggable/SluggableListener.php
+++ b/lib/Gedmo/Sluggable/SluggableListener.php
@@ -434,10 +434,12 @@ class SluggableListener extends MappedEventSubscriber
         // collect similar persisted slugs during this flush
         if (isset($this->persisted[$class = $ea->getRootObjectClass($meta)])) {
             foreach ($this->persisted[$class] as $obj) {
-                if ($base !== false && $meta->getReflectionProperty($config['unique_base'])->getValue($obj) !== $base) {
+                $similarMeta = $om->getClassMetadata(get_class($obj));
+                if ($base !== false && $similarMeta->getReflectionProperty($config['unique_base'])->getValue($obj) !== $base) {
                     continue; // if unique_base field is not the same, do not take slug as similar
                 }
-                $slug = $meta->getReflectionProperty($config['slug'])->getValue($obj);
+                $slugRefl = $similarMeta->getReflectionProperty($config['slug']);
+                $slug = $slugRefl->getValue($obj);
                 $quotedPreferredSlug = preg_quote($preferredSlug);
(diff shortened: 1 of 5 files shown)


l3pp4rd:
thanks ;) there were no cases covered concerning joined inheritance

PR #1611, opened 2016-06-09
Add methods verify() and recover() for ClosureTree

### LINKED ISSUE #1430: how to update category_closure table if we have existing data ?
Hi,
I have existing category hierarchy with parent_id column. I have configured Gedmo Tree extension and update schema. I can see 'level' column is added to category table and category_closure table is created.

How do I populate the data in category_closure table.  There are no 
    $repo->verify();
    $repo->recover();

methods for ClosureTreeRepository. I have tried updating category records so that to trigger TreeListener but still there is no data in category_closure.

How do I populate hierarchy in category_closure table ?

aminin:
The solution for #1430

closes #1430


l3pp4rd:
great, thanks ;) eh, will need to manage phpunit deprecations

l3pp4rd:
would be good to have a test cases too..

PR #1619, opened 2016-06-27
change odm string annotation to field with type string

fredeastside:
This changes are fixed error, which appears in php v. > 7 - [RuntimeException]                                                                                                                                                       
  An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                
  PHP Fatal error:  Cannot use 'String' as class name as it is reserved in

### DIFF
--- a/doc/blameable.md
+++ b/doc/blameable.md
@@ -280,14 +280,14 @@ class Article
     private $id;
 
     /**
-     * @ODM\String
+     * @ODM\Field(type="string")
      */
     private $title;
 
     /**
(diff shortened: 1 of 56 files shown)


l3pp4rd:
thanks ;)

webdevilopers:
Hey @fredeastside and @l3pp4rd ! Coming from this issue:
* https://github.com/stof/StofDoctrineExtensionsBundle/issues/332#issuecomment-235898693

Dou you think there is a chance to push the fix into a **2.4.x** release?

PR #1623, opened 2016-07-07
Add closure tree operations in 2.4.x

aminin:
verify() and recover()


l3pp4rd:
thanks ;)

aminin:
It would be great to add a new tag

PR #1626, opened 2016-07-08
Support for Symfony 2.6 and above

stefano-93:
Modified DoctrineExtensionListener to support both Symfony < 2.6 and Symfony >= 2.6 and to support blameable extension too.


l3pp4rd:
thanks ;)

PR #1627, opened 2016-07-09
Ensure children elements in Gedmo namespace are found

holtkamp:
Prevents warnings like "Node no longer exists" (triggered when wrong namespace is used) by making sure child elements exist before processing it. 

Related:
https://github.com/Atlantic18/DoctrineExtensions/issues/1598
https://github.com/Atlantic18/DoctrineExtensions/issues/1625


l3pp4rd:
thanks ;)

PR #1629, opened 2016-07-11
Fix typo

nacholibre:
I've spent probably an hour wondering why this example doesn't work, and finally I saw that `fields` should be `field` without the `s`.


l3pp4rd:
thanks ;)

PR #1630, opened 2016-07-13
Type-hinted parameters in TimestampableListener

Brunty:
They were in the docblock, but not in type hinted as parameters.


Brunty:
Will close this PR, just noticed the typehinting when a friend linked it and edited through web editor for speed thinking it should be a quick fix, will look at fixing properly when I get a chance.

l3pp4rd:
blemeable and iptraceable extends timestampable listener. so those should be updated too.

Brunty:
@l3pp4rd yeah, I'll look into updating all the right things when I've got the chance to get it properly pulled down and fixed locally, initially it looked like a quick fix but tests broke, never a simple thing, eh? ;)

l3pp4rd:
yes, extensions is a complex abstraction ;)

Issue #1634, opened 2016-07-18
BC break for tilde fields


kix:
After [this commit](https://github.com/Atlantic18/DoctrineExtensions/commit/ba17ee7ee7b460f7c47652712bb50e10f4353ada), having a tilde (`~`) in an entity's field configuration leads to a fatal error, as the extracted method has an `array` typehint, yet the tilde translates to `null`.

kix:
Versions after 2.4.16 seem to be affected by this.

vbartusevicius:
@kix - please give some more information, i.e. exact fiend name and the error you're getting.

kix:
@vbartusevicius, here's the YAML config:
```yml
UserBundle\Entity\OAuth:
    type: entity
    table: user_oauth
    id:
        id:
            type: integer
            generator:
                strategy: AUTO
    fields:
        provider:
            column: provider
        identifier:
            column: identifier
        accessToken:
            column: access_token
            nullable: true
        refreshToken:
            column: refresh_token
            nullable: true
    manyToOne:
        user:
            targetEntity: UserBundle\Entity\User
            joinColumn:
                nullable: false
            inversedBy: oauthAccounts
    uniqueConstraints:
        -
            columns: [user_id, provider]
        -
            columns: [provider, identifier]
```

And the error I get is:

```
  [RuntimeException]                                                                                                                                                                                                                                                                                                         
  An error occurred when executing the "'cache:clear --no-warmup'" command:                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                             
    [Symfony\Component\Debug\Exception\ContextErrorException]                                                                                                                                                                                                                                                                
    Catchable Fatal Error: Argument 2 passed to Gedmo\Translatable\Mapping\Driver\Yaml::buildFieldConfiguration() must be of the type array, null given, called in /drone/src/github.com/vivaster/api.vivaster.com/vendor/gedmo/doctrine-extensions/lib/Gedmo/Translatable/Mapping/Driver/Yaml.php on line 51 and defined    
```

vbartusevicius:
Added PRs to `2.4.x` and `master` branches. @l3pp4rd - please add `patch` tag to propagate the changes to packagist.

kix:
@vbartusevicius, thanks for fixing this so quick :)

vbartusevicius:
Well, I left a major bug there... :)

filipe-x:
Hello There

[...]

PR #1635, opened 2016-07-18
fix for #1634

vbartusevicius:
Do not typehint `array`, but check internally. Fixes https://github.com/Atlantic18/DoctrineExtensions/issues/1634


l3pp4rd:
thanks ;)

Issue #1638, opened 2016-07-25
 Incorrect syntax for arrays in PHP 5.3 for version 2.4.x


Loscillo:
Hi,

In the commit 7c85797d06ded72d3e768e34c41ab0bf383fc699, the short syntax to create array was used in line 407 : 

`$errors = [];`

However, if [we look](https://packagist.org/packages/gedmo/doctrine-extensions#v2.4.18) at the version of PHP required to run the extension in version 2.4.x, this is the version 5.3.2. I know that PHP 5.3 has been abandoned, but it would be good to update the dependencies or maintain compatibility.

l3pp4rd:
indeed, could you submit a pull request?

PR #1639, opened 2016-07-25
Fix incorrect syntax for arrays in PHP 5.3

Loscillo:
Pull request for issue #1638 

Replace short syntax for arrays in PHP 5.3


l3pp4rd:
thanks

PR #1646, opened 2016-08-11
Fix for nested set issue in php7 (see https://3v4l.org/IARp7)

kbond:
I'm not sure if this is the best solution but it fixed my problem.

In php 7.0, `ReflectionProperty::getValue()` creates a notice if the property does not exist. See https://3v4l.org/IARp7.  In php 7.1 it looks like it is going to throw an exception because the type doesn't match.

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/Nested.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/Nested.php
@@ -534,6 +534,9 @@ class Nested implements Strategy
                 if ($node instanceof Proxy && !$node->__isInitialized__) {
                     continue;
                 }
+                if (get_class($node) !== $class) {
+                    continue;
+                }
                 $oid = spl_object_hash($node);
                 $left = $meta->getReflectionProperty($config['left'])->getValue($node);
                 $currentRoot = isset($config['root']) ? $meta->getReflectionProperty($config['root'])->getValue($node) : null;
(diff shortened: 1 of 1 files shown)


lib/Gedmo/Tree/Strategy/ORM/Nested.php
```diff
@@ -534,6 +534,9 @@ public function shiftRL(EntityManager $em, $class, $first, $delta, $root = null)
                 if ($node instanceof Proxy && !$node->__isInitialized__) {
                     continue;
                 }
+                if (get_class($node) !== $class) {
```
l3pp4rd: tests are failing, maybe you could use `$meta->getReflectionClass()->hasProperty($config['left'])` ?

kbond: This won't work as the problem is the `$meta` is always the correct NestedSet entity but the `$node` can be a non-nested-set entity.

kbond:
@l3pp4rd using the metadata of `$node` worked. The tests are passing.

l3pp4rd:
thanks ;)

kbond:
Can this be backported to the 2.4 branch?

l3pp4rd:
hm, guess need another PR in 2.4.x branch.

kbond:
Ok, see #1647

PR #1647, opened 2016-08-17
PHP 7 nested set fix for 2.4

kbond:


kbond:
Thanks, I take it that failing test was expected?

l3pp4rd:
yes, phpunit deprecations. fixed in master

Issue #1604, opened 2016-06-01
Slug validation on duplicate entry


ducho:
Hi. 
Is it possible to validate slug (title) after form submit? UniqueEntity() not working becouse slug is generated with onFlush() handler. 

Now, after flush() is calling exception: 

> SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'spolocnost/pracovne-prilezitosti' for key 'UNIQ_140AB620989D9B62'
500 Internal Server Error - UniqueConstraintViolationException

I need to validate slug before persist and if exists duplicate slug the return error message to form. 

Thanks.
Ducho.

l3pp4rd:
Hi, if you set unique option for slug field it will prefix non unique slug with index. if you do not wish to have an index, I suggest you simply use [sluggable transliterator](https://github.com/Behat/Transliterator/blob/master/src/Behat/Transliterator/Transliterator.php) in your form data transformer and then form is validated in controller confirm that it is unique by querying a database.

ducho:
Thanks for quick reply. That transliterator is compatible with $sluggableListener->generateSlug() ?

l3pp4rd:
yes, it uses it

ducho:
This annotation not working with unique index.
   
    /**
     * @Gedmo\Slug(handlers={
     *      @Gedmo\SlugHandler(class="Gedmo\Sluggable\Handler\TreeSlugHandler", options={
     *          @Gedmo\SlugHandlerOption(name="parentRelationField", value="parent"),
     *          @Gedmo\SlugHandlerOption(name="separator", value="/")
     *      })
     * }, fields={"title"}, unique=true)
     * @ORM\Column(type="string", length=255, unique=true)
     */
    private $slug;

ducho:
The sluggable transliterator is not possible to use with:
Sluggable + NestedTree + TreeSlugHandler

@ORM\PreFlush() not working too... 

No exists solution to handle slug duplicity with Forms...

l3pp4rd:
slug handlers do not check the uniqueness, the library needs to be improved concerning this case. but probably there was a reason, why it is so, just can't remember now.

ducho:
Sorry, but probably I do not understand enough... What is the purpose function makeUniqueSlug() in the SluggableListener?

l3pp4rd:
as far as I remember slug handlers are not checked for uniqueness

ducho:
[...]

Issue #1429, opened 2015-10-15
[Sluggable] unique check after running TreeSlugHandler


vlcekmi3:
Hi, is there any possibility to generate unique slug after the slug is handled by TreeSlugHandler?

Example:
I have two entitties
```
1. entity --> slug = "test"
2. entity --> slug = "entity_with_sub-entity"

3. entity (before persist) --> slug = "test"
```
and then i wanna persist 3rd entity, which is nested to 2nd entity (so resulted slug should be "test/xxx")

expected result: `3. entity --> slug = "entity_with_sub-entity/test"`
received result: `3. entity --> slug = "entity_with_sub-entity/test-1"` because unique slug is generated before TreeSlugHandler do anything

lashae:
@vlcekmi3 I have the same issue. Since you had closed the issue, i suppose you fixed it. How did you manaage to fix?

PR #1648, opened 2016-08-17
Fix unique slugs with the TreeSlugHandler

cmfcmf:
- Added test case for #1429 and #1604
- Added new callback before the slug's uniqueness is ensured while maintaining backward compatibility.


cmfcmf:
Maybe I should've targeted the 2.4 branch. Hmm.

l3pp4rd:
I will review that when I can, currently have a busy schedule ;)

PR #1650, opened 2016-08-22
[2.4] Fix unique slugs with the TreeSlugHandler

cmfcmf:
_This is the same as #1648 but for the 2.4 branch._
- Added test case for #1429 and #1604
- Added new callback before the slug's uniqueness is ensured while maintaining backward compatibility.


cmfcmf:
The test failure is not related to this change and has already been there in older builds, for example: https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/153029559.

l3pp4rd:
thanks ;)

PR #1659, opened 2016-09-08
Fix node deletion for similar paths with materialized path strategy

urbicon:
When deleting a node with a path that is contained in other paths, those nodes will be deleted as well.

Example:
node1: /one/two
node2: /one/two/three
node3: /one/two-test

**Problem**:
Deleting node1 will delete all three nodes.

**Expectation**:

Deleting node1 will delete nodes 1 and 2.

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/MaterializedPath.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/MaterializedPath.php
@@ -23,12 +23,18 @@ class MaterializedPath extends AbstractMaterializedPath
         $wrapped = AbstractWrapper::wrap($node, $om);
 
         $path = addcslashes($wrapped->getPropertyValue($config['path']), '%');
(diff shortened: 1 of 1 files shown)


l3pp4rd:
this is an expected behavior. when you delete a directory in your system, what happens? Move all node1 children nodes before removing if you wish to have this behavior.

urbicon:
I guess I didn't make it quite clear, the problem is not the deletion of children, but of sibling nodes.
Let's take the example of an url tree: When I delete 'home', it is expected behaviour to also delete 'home/subpage'. It is not expected to also delete 'homepage', which it will do right now, because it looks for paths LIKE 'home%'.

l3pp4rd:
ah, yes make sense, thanks for explaining. the **master** branch is not released yet, if you want this to go to 2.4.x releases, then submit same PR to 2.4.x branch

PR #1665, opened 2016-09-21
Add example to loggable: Loggable Entity example with your custom logEntryClass

nutii:
Add little example to docs which shows how create own "logEntryClass".


l3pp4rd:
thanks ;)

Issue #1666, opened 2016-09-21
QueryAnalyzer fails when parameters contains array


x86demon:
QueryAnalyzer trigger notice in method generateSql on line 199 because not all parameter types are converted to string/number in getConvertedParams.
Conversion is failed for type == 101 (Connection::PARAM_INT_ARRAY) and type == 102 (Connection::PARAM_STR_ARRAY). For this types value returned as is (array)

Steps to reproduce.
We have some entity.

```php
/**
 * @ORM\Entity
 */
class DummyEntity
{
   /**
     * @var integer
     *
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    public $id;
}
```

Data is fetched next way:

```php
$repository->findBy(['id'=> [1, 2, 3]]);
```

Issue Fix PR https://github.com/Atlantic18/DoctrineExtensions/pull/1667

PR #1667, opened 2016-09-21
Fix issue #1666 QueryAnalyzer fails when parameters contains array

x86demon:
Fixed Issue #1666 
Added conversion of parameters with type: Connection::PARAM_INT_ARRAY and Connection::PARAM_STR_ARRAY which fix Array to string conversion notice

Please, include this fix to version 2.3.x


l3pp4rd:
thanks ;)

PR #1668, opened 2016-09-22
Fix sortable doc annotation names

virtualize:
doc was wrong, referring to an (probaly old) @Gedmo\Sortable annotation


l3pp4rd:
Hi, you should look at the doc on the release version you are using, master branch has backward incompatible changes and though annotation is correct

virtualize:
you're right by bad.

PR #1669, opened 2016-09-22
Fix node deletion for similar paths with materialized path strategy (2.4.x)

BowlingX:
See pull-request #1659 in master. Fix for 2.4.x Branch.

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/MaterializedPath.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/MaterializedPath.php
@@ -23,18 +23,21 @@ class MaterializedPath extends AbstractMaterializedPath
         $wrapped = AbstractWrapper::wrap($node, $om);
 
         $path = addcslashes($wrapped->getPropertyValue($config['path']), '%');
-        $lvl = $wrapped->getPropertyValue($config['level']);
 
         // Remove node's children
         $qb = $om->createQueryBuilder();
         $qb->select('e')
             ->from($config['useObjectClass'], 'e')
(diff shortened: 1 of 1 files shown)


lib/Gedmo/Tree/Strategy/ORM/MaterializedPath.php
```diff
@@ -23,12 +23,18 @@ public function removeNode($om, $meta, $config, $node)
         $wrapped = AbstractWrapper::wrap($node, $om);
 
         $path = addcslashes($wrapped->getPropertyValue($config['path']), '%');
+        $lvl = $wrapped->getPropertyValue($config['level']);
```
l3pp4rd: actually there might be an issue if level is not configured. you should first check it with `isset($config['level'])` like https://github.com/freshcells/DoctrineExtensions/blob/dc0c702cc7b3b7376f9761099f209d8c864c1a1f/lib/Gedmo/Tree/Strategy/AbstractMaterializedPath.php#L316 because it is optional. should be also fixed in master, haven't noticed it before

BowlingX: I added a check

BowlingX: I guess the check `$meta->hasField($lvlField)` is not needed.

l3pp4rd: yes, if config has it, then there is no need to check field

PR #1671, opened 2016-09-23
Fix node deletion for similar paths with materialized path strategy

BowlingX:
Same fix for master, see #1669.


BowlingX:
Sorry for the reverts, I was testing out another problem and committed it.

PR #1674, opened 2016-09-25
Add support for empty root nodes in MaterializedPath strategy

BowlingX:
Right now, `MaterializedPath` does not support empty root nodes.
I added support for empty `titles` that also play correctly with other configuration options like `path_starts_with_separator` etc.

This is work in progress, I will add more tests to cover other configurations.


l3pp4rd:
this is a big change, what do you mean by empty root nodes, if a node path element is empty, that does not make any sense. the path would simply not be valid, for example a child node could match the same root/child node.

BowlingX:
Right now it is not possible / buggy to have a node that has an empty name and has no parent id (what yields to an empty path). I also introduced support for `path_starts_with_separator`, so we can have a real root node in this strategy. (/).

l3pp4rd:
No, no, I cannot accept such feature, extensions are currently locked for new features (including new options) unless they are really necessary. But in your case, empty path nodes could lead to much more trouble and confusion for every user who may interact with it. I suggest you to prevent such tree paths or use NestedSet instead, you can also base your path on Node ID, and build path in separate repository method based on titles (which may be empty in your case). But this is not something I would want to include in the library at all.

BowlingX:
I understand. But `NestedSet` supports the same behaviour and MaterializedPath is broken in this case. This is not really a new feature but a bugfix. I will provide an example. This should also not break existing implementations.

l3pp4rd:
nested set does not depend on titles and for that reason it cannot be broken. if you use empty titles for a tree path I think that is your problem, unless I do not understand something. Please explain how you are trying to fix it?

AkenRoberts:

Thank you!

PR #1675, opened 2016-09-30
Fix url for xsd schemas

Shine-neko:
http://gediminasm.org/schemas/orm/doctrine-extensions-mapping now return a 404

### DIFF
--- a/lib/Gedmo/Mapping/Driver/Xml.php
+++ b/lib/Gedmo/Mapping/Driver/Xml.php
@@ -18,7 +18,7 @@ use SimpleXMLElement;
  */
 abstract class Xml extends File
 {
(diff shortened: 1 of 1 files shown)


l3pp4rd:
that would start to fail for everyone. the same way as you see it in tests. This xsd schema url is not required to be available and I cannot change the url because it will not be compatible anymore. this change would break BC.

l3pp4rd:
use master branch if you want it to validate through correct xsd schema url.

Shine-neko:
@l3pp4rd It is possible to recover http://gediminasm.org/schemas/orm/doctrine-extensions-mapping ?

l3pp4rd:
it is not important to have it recovered. because you only need schema for validation. you may get it from the repository and validate if you need to.

PR #1677, opened 2016-10-04
Use now expression for timeAware filter constraint to prevent cached times from being used for filtering

benjiw87:
As suggested by @l3pp4rd at https://github.com/Atlantic18/DoctrineExtensions/issues/1356


l3pp4rd:
hope it will be compatible with the databases. the master branch is not released yet, so it probably is a good place to look if there are any issues with that change. won't patch 2.4.x versions since it may involve BC breaks probably

stephanvierkant:
Any idea when the fix will be available in a tagged version? It's only merged into the master branch right now.

stephanvierkant:
@l3pp4rd Is this something that could be cherry-picked for `v2.4.x`?

PR #1680, opened 2016-10-07
Fix hardcoded field names in ClosureTreeRepository

teohhanhui:
Also fixed wrong node level value (should be closure depth + 1).


l3pp4rd:
Hi, is it backward compatible? if level is changed might have impact on upgrade

teohhanhui:
@l3pp4rd This does not get called normally. The `updateLevelValues` function was doing the wrong thing anyway. At least after my fix `recover` + `updateLevelValues` works properly for initializing the closure tree (from a flat structure in my case). And I see that it has been removed in `master`...

l3pp4rd:
ah, if thats in recover, then it is fine, thanks ;)

PR #1683, opened 2016-10-11
Fix PHP7 error "Cannot use 'String' as a class name".

spiroski:
This enables the 2.4.x version to run on PHP7 with MongoDB


spiroski:
Actually

PR #1692, opened 2016-10-19
[Tree] Fix positioning for rootless trees

iluuu1994:
Here's the same fix for the `master` branch. It required a small adjustment concerning siblings but that was it. Thanks for merging!


l3pp4rd:
thanks ;)

PR #1697, opened 2016-10-26
fix translation when the value is '0' or empty string

Uplink03:
# Description

I got a bug report that although unpublished the content would still show and during investigation I found that the "isPublished" field was still true. I found out that it was falling back to the base language for every PHP-falsy value; i.e. 0, "0", "".

This patch fixes that.

# Explanation of the patch

It starts with $translated set to a hard false value, and it expects that the translation field is always a string. This allows for "" and "0" to be valid translations for situations where such is desirable. If $translated is different from PHP boolean false then it is considered intentional.

# Point of thought

I haven't explored nullable columns.

### DIFF
--- a/lib/Gedmo/Translatable/TranslatableListener.php
+++ b/lib/Gedmo/Translatable/TranslatableListener.php
@@ -470,7 +470,7 @@ class TranslatableListener extends MappedEventSubscriber
             );
             // translate object's translatable properties
             foreach ($config['fields'] as $field) {
-                $translated = '';
+                $translated = false;
                 foreach ((array) $result as $entry) {
                     if ($entry['field'] == $field) {
                         $translated = $entry['content'];
@@ -478,7 +478,7 @@ class TranslatableListener extends MappedEventSubscriber
                     }
                 }
                 // update translation
-                if ($translated
+                if ($translated !== false
                     || (!$this->translationFallback && (!isset($config['fallback'][$field]) || !$config['fallback'][$field]))
                     || ($this->translationFallback && isset($config['fallback'][$field]) && !$config['fallback'][$field])
                 ) {


l3pp4rd:
Hi, thanks for getting back to this. I think we should use a different value to identify if field was not translated. Something what cannot come up as a translation.

``` php
const UNTRANSLATED = '__$untranslated$__';

$translated = self::UNTRANSLATED;
foreach ((array) $result as $entry) {
    if ($entry['field'] == $field) {
        $translated = $entry['content'];
        break;
    }
}
```

Because `false` or `null`, might be a valid translation.

Uplink03:
While `'__$untranslated$__'` is very unlikely to happen by itself in a translation, it's still not impossible.

But I like the general idea.

The things that you can never get are: an object or an array. The solutions I have in mind now aren't all pretty :)

Option 1, using an array placeholder:
```php
const UNTRANSLATED = array();

$translated = self::UNTRANSLATED;
// ...
if (!is_array($translated)) { //...
```

Option 2, using an object placeholder:
```php
$translated = new \stdClass();

if (!is_object($translated)) { //...
```

Option 3, using a separate flag:
```php
$is_translated = false;
foreach ((array) $result as $entry) {
    if ($entry['field'] == $field) {
        $translated = $entry['content'];
        $is_translated = true;
        break;
    }
}

// ...

if ($is_translated) { //...
```

At this point I'd refactor my proposed solution to use Option 3. It doesn't do any acrobatics with variable types and it's the cleanest of them all.

I'll make that change later today and make another PR.

l3pp4rd:
lets go with `$translated = new \stdClass();` since array constants are only supported in PHP 7

Uplink03:
discussed changes are in PR #1698

PR #1698, opened 2016-10-28
fix translation when the value is falsy (e.g empty string, '0', null)

Uplink03:
# Description

I got a bug report that although unpublished the content would still show and during investigation I found that the "isPublished" field was still true. I found out that it was falling back to the base language for every PHP-falsy value; i.e. 0, "0", "", null, false.

This patch fixes that.
# Explanation of the patch

It starts with $translated set to an impossible translation value (`new \stdClass`). This allows for "" and "0", false and null to be valid translations for situations where such is desirable.

### DIFF
--- a/lib/Gedmo/Translatable/TranslatableListener.php
+++ b/lib/Gedmo/Translatable/TranslatableListener.php
@@ -470,7 +470,10 @@ class TranslatableListener extends MappedEventSubscriber
             );
             // translate object's translatable properties
             foreach ($config['fields'] as $field) {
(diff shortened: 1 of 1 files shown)


l3pp4rd:
many tests have failed, guess it somehow tried to use stdClass as a value the condition succeeded

Uplink03:
OK... I'll check this out when I get the time to understand what's happening. It looks like I have a busy week ahead. I'll return with a new PR when I have a solution that passes the tests.

PR #1700, opened 2016-11-08
Add support for embeddable, add test

kvdnberg:
I wanted to have a Timestampable changed date that would update when embeddable properties were updated, this did not seem to work because the code is counting on the changeSet containing objects. However for embeddables, the changed property is included like "author.name". 

This PR should allow embeddable properties to be set as the fields to check for changes. I also updated the test to check this, but I couldn't get the tests running locally (PHP7) so I'm checking here if they're ok... fingers crossed.

### DIFF
--- a/lib/Gedmo/AbstractTrackingListener.php
+++ b/lib/Gedmo/AbstractTrackingListener.php
@@ -106,12 +106,17 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
                         $trackedFields = $options['trackedField'];
                     }
 
-                    foreach ($trackedFields as $tracked) {
+                    foreach ($trackedFields as $trackedField) {
                         $trackedChild = null;
-                        $parts = explode('.', $tracked);
+                        $parts = explode('.', $trackedField);
                         if (isset($parts[1])) {
                             $tracked = $parts[0];
                             $trackedChild = $parts[1];
+                        }
+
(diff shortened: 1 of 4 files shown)


l3pp4rd:
yes, the tests for php 7 fixed only in master branch. seems like there are issues.

kvdnberg:
Managed to get the tests working locally by skipping the mongodb dependency. Tests should pass now.

kvdnberg:
Only PHP 5.6 failed, ironically, because that's the version I used locally. Not sure what the problem is? Can anyone help figure it out?

l3pp4rd:
but this change will react to any property of embeded object or am I wrong? etc if you change Author.email instead in the test, on update it will update the timestamp correct?

l3pp4rd:
concerning php 5.6 failure, I think it may be due to phpunit version. 2.4.x is not adapted for this sadly :)

kvdnberg:
The changeset from Doctrine will contain "author.email" as one of the changed fields (if that has in fact changed) in case of an embedded object. Before the change, if "author.email" was in one of the tracked fields, it would be split in parts and then checked as though "author" in the changeset should be an array (as I believe it would be in one-to-many or many-to-many relationships) but it's not. The change ensures that if there is a "author.email" in the changeset as well as in the tracked fields, the "changed at" date is updated.

l3pp4rd:
thanks ;)

PR #1709, opened 2016-11-28
fix translation when the value is falsy (e.g empty string, '0', null)

Uplink03:
# Description

I got a bug report that although unpublished the content would still show and during investigation I found that the "isPublished" field was still true. I found out that it was falling back to the base language for every PHP-falsy value; i.e. 0, "0", "", null, false.

This patch fixes that.
# Explanation of the patch

I'm using a flag variable `$is_translated` to mark if the value of `$translated` is the result of a translation. This allows `$translated` to be any value, including the falsy ones.


Uplink03:
A test failed for a reason beyond my control in PHP 5.6. The tests were passed on the other platforms and on my machine.

l3pp4rd:
thanks, this one is simple and good solution

PR #1723, opened 2016-12-13
Use mb_strlen to get string length

mcorteel:
I use the sluggable extension to get a user-friendly path generated for my tree nodes.
For that, the name used to generate the slug is the unaltered name of the node (which can contain accentuated characters). These characters are counted as more than 1 character by `strlen`, which messes up the result of `replaceRelative`. Using `mb_strlen` instead solves the problem!


mcorteel:
I realize that there are more changes to do if this pull request is accepted. But I wanted to discuss the idea first.

l3pp4rd:
Hi, thanks can you open PR to 2.4.x branch. master is not the main branch and it might be that 2.4 version will stay for long

mcorteel:
I'm closing this pull request to open a clean one.

PR #1724, opened 2016-12-13
Use mb_strlen to get string length to avoid unicode characters problems

mcorteel:
Re-creating this pull request (including another use of `strlen`).
So, what do you think, should it be merged?


l3pp4rd:
thanks ;)

PR #1727, opened 2016-12-21
Allow Sortable objects to be compared by custom method

yellow1912:
Sometimes it is necessary to customize the way objects should be compared.
This will help to fix issues such as deep nesting reference when we
compare objects that link to themselves.

This commit fixes Atlantic18/DoctrineExtensions#1726


lib/Gedmo/Sortable/Comparable.php
```diff
@@ -0,0 +1,17 @@
...
+ * Entities which has custom compare methods
+ * Comparable
+ *
+ * @link https://wiki.php.net/rfc/comparable
+ * @author Raine Ng <yellow1912@gmail.com>
+ * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
+ */
+interface Comparable
+{
+    public function compareTo($other);
```
l3pp4rd: please specify that it should return boolean if it matches. because in most languages compareTo returns 0 if equals -1 if less and 1 if greater..

l3pp4rd:
this can be accepted, but please add a section in documentation explaining it. thanks

yellow1912:
Since doctrine already has the Comparable interface, I decided to drop the Comparable interface in Sortable. The code thus becomes much more minor now.

I also added some more documentation for Sortable to cover this use case.

l3pp4rd:
yes, this is much better, thanks ;)

l3pp4rd:
there is no need for it in master, I think we will not make a release from master, 2.4 will may be the last version of extensions.

yellow1912:
Thank you for your time. I appreciate your time spent on this project even when you don't have much use for it anymore. This is a wonderful project and hopefully someone can help to pick it up in the future :).

l3pp4rd:
even if no one picks it up, it will be backward compatible which is also good ;)

PR #1728, opened 2016-12-21
Fix linking error for Custom Comparison in document

yellow1912:
My bad, I have just realized that I made a mistake in the document internal linking. Sorry for the trouble.


l3pp4rd:
thanks for fixing it :)

PR #1733, opened 2016-12-28
Add PHPDoc for soft deleteable filter.

Aliance:
This allows IDE the correct highlighting works.


Aliance:
cc @l3pp4rd 
Do not know why test are broken on 5.6

l3pp4rd:
phunit version is used which has deprecations, and it fails. nevermind. thanks ;)

PR #1734, opened 2016-12-28
[Sluggable] Add callback support to generate slug

michaelperrin:
Hello!

This PR introduces a new `callback` option for the Sluggable behavior. It looks quite big in the code, but is actually quite simple.

The purpose of this addition is two-fold:

* It allows more than one relative handlers.
* It is simpler to use than the `RelativeSlugHandler` if there is no need for the `InversedRelativeSlugHandler`.

It can be used instead of the usual `fields` option to define a callback on the current object that will return the slug tokens.
The `fields` option is therefore not required anymore (either `fields` or `callback` is now required).

It is currently not possible to define multiple relative handlers, for generating a slug with a structure like `<category>-<brand>-<model>` (where the category and the brand are relationships to the current entity).

A question was raised on StackOverflow about it: http://stackoverflow.com/questions/37397975/multiple-relativeslughandler

The feature introduced in this PR can handle this, like in this example:

```php
<?php
class Product
{
    /**
     * @ORM\ManyToOne(targetEntity="Category")
     */
    protected $category;
    
    /**
     * @ORM\ManyToOne(targetEntity="Brand")
     */
    protected $brand;
    
    /**
     * @ORM\Column(name="model", type="string", length=128)
     */
    protected $model;

    /**
     * @Gedmo\Slug(callback="getSlugTokens")
     */
    protected $slug;

    public function getBrand()
    {
        return $this->getBrand();
    }

    public function getCategory()
    {
        return $this->getCategory();
    }

    public function getModel()
    {
        return $this->getModel();
    }

    public function getSlugTokens()
    {
        return [
            $this->getCategory()->getName(),
            $this->getBrand()->getName(),
            $this->getModel(),
        ];
    }
}
```

This doesn't replace the `RelativeSlugHandler` as the there is no counterpart for the
`InversedRelativeSlugHandler`.
But it is really useful for getting tokens to generate a slug, be it for relationships
or other kind of data.

I hope you like this idea! I'm open to any suggestion for improving it :)

### DIFF
--- a/doc/sluggable.md
+++ b/doc/sluggable.md
@@ -12,6 +12,10 @@ Features:
 - Annotation, Yaml and Xml mapping support for extensions
 - Multiple slugs, different slugs can link to same fields
 
+Update **2016-12-28**
+
+- Callback support
+
 Update **2013-10-26**
 
 - Datetime support with default dateFormat Y-m-d-H:i
@@ -328,7 +332,8 @@ echo $article->getSlug();
 
 ### Some other configuration options for **slug** annotation:
 
-- **fields** (required, default=[]) - list of fields for slug
+- **fields** (required if **callback** is not defined, default=[]) - list of fields for slug
+- **callback** (required if **fields** is not defined, default=null) - callback method name to get slug tokens
 - **updatable** (optional, default=true) - **true** to update the slug on sluggable field changes, **false** - otherwise
 - **unique** (optional, default=true) - **true** if slug should be unique and if identical it will be prefixed, **false** - otherwise
 - **unique_base** (optional, default=null) - used in conjunction with **unique**. The name of the entity property that should be used as a key when doing a uniqueness check.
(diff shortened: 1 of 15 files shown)


l3pp4rd:
there won't be new features, you can extend sluggable with handlers or transliterators or simply by extending listener

michaelperrin:
Aw, too bad. Implementing this feature would be much harder if is not native.
Why no new feature can be merged, if tests are passing? (I also added the necessary tests for this feature.)

l3pp4rd:
too many options confuse people, especially when the same thing can be done in many ways. currently I will only maintain extensions at it's current feature set, nothing new will be added. green tests does not mean that it is fully compatible with all possible scenarios. I simply do not have time for extension maintenance as I do not develop with PHP anymore. the project is at stable point and I'll leave it there. If someone wants **doctrine-extensions2** they can go this path and create the repository.

michaelperrin:
Ok, thank you for this clarification.

Would more new features be needed, a fork may be the way to go, but for now there are not enough in the pipeline to make one.

AkenRoberts:

Thank you!

PR #1735, opened 2016-12-31
Type-hint DateTimeInterface rather than DateTime

foaly-nr1:
This establishes compatibility with DateTimeImmutables.


l3pp4rd:
hi, doctrine extensions will remain to be compatible with PHP 5.4.

foaly-nr1:
😢

Issue #1512, opened 2016-01-25
[Translatable] Performance issues workaround


TriAnMan:
Hey guys!

I've found a workaround for performance issues of Translatable. E.g. #860

Currently it applies for MySQL DB, PostgreSQL users should make little more things to make indexes working. I.e. use Gedmo v2.4.1 or less in case of b90d088a64715015720d669050ad0eabde169903 (see #1249).

So the main problem is type mismatch of a `foreignKey` field in a translations table (which is a varchar) and a primary key of translated entities tables which in most cases is integer. In this case MySQL silently converts one type to another and obviously not uses indexes. Postrges has more reliable behavior and warns the user about a type mismatch if there is no explicit typecasting. See #1249 (also #1168, #1142 and #1383)

Current behavior is satisfactory if there is no many translations (but for my mind such assumption is too naive). If there is more than 1 thousand of translations then a lot of performance problems arise, especially for queries with joins.

To workaround this problem we should harmonize types of columns. Obviously we can't force an autoincremental ID to be a string. So the only option left is to change a type of the `foreignKey` field. Here is a way to do this.
1. Override `\Doctrine\ORM\Mapping\ClassMetadataFactory`

```
<?php

namespace YourNameSpace;

use Doctrine\ORM\EntityManager;

class ClassMetadataFactory extends \Doctrine\ORM\Mapping\ClassMetadataFactory
{
    /**
     * @var EntityManager
     */
    private $em;

    /**
     * Workaround for private EntityManager field in \Doctrine\ORM\Mapping\ClassMetadataFactory::newClassMetadataInstance()
     * @inheritdoc
     */
    public function setEntityManager(EntityManager $em)
    {
        $this->em = $em;
        parent::setEntityManager($em);
    }

    /**
     * @inheritDoc
     */
    protected function newClassMetadataInstance($className)
    {
        return new \YourNameSpace\ClassMetadata($className, $this->em->getConfiguration()->getNamingStrategy());
    }
}
```
1. Override `\Doctrine\ORM\Mapping\ClassMetadata`

```
<?php

namespace YourNameSpace;

class ClassMetadata extends \Doctrine\ORM\Mapping\ClassMetadata
{
    /**
     * @inheritdoc
     */
    public function mapField(array $mapping)
    {
        // Fix performance issue with column types mismatch and lack of indexes optimization
        if (
             $this->name === 'Gedmo\Translatable\Entity\Translation'
[...]

PR #1742, opened 2017-01-24
Minor update of documentation for Translatable

svgrafov:
Update Translatable documentation to warn users of performance cap.
https://github.com/Atlantic18/DoctrineExtensions/issues/1502#issuecomment-171011176
https://github.com/Atlantic18/DoctrineExtensions/issues/860#issuecomment-26154908

### DIFF
--- a/doc/translatable.md
+++ b/doc/translatable.md
@@ -3,7 +3,9 @@
 **Translatable** behavior offers a very handy solution for translating specific record fields
 in different languages. Further more, it loads the translations automatically for a locale
 currently used, which can be set to **Translatable Listener** on it`s initialization or later
-for other cases through the **Entity** itself
+for other cases through the **Entity** itself.
+
+Please note that **Translatable** behaviour is meant to be used for small projects and prototypes because of performance limitations.
 
 Features:
 


l3pp4rd:
well, there are many users who use it in production for larger projects. the point is that by default translatable casts foreign key to string (which is not optimal if you have integer). in that case optimization is to override the mapping of and change the field to integer.

It could mention instead that it may handle optimally up to million row tables with certain optimizations like using personal translations or overriding mapping of foreign key to prevent casting in database.

TriAnMan:
@l3pp4rd not millions. 50k of translation rows is enough for performance cap mentioned in #1512

l3pp4rd:
yes, but it mentions exactly that, you need to modify the type of **foreignKey** to improve the performance.

> Have fun with blazing speed of your DB queries. (In some cases we have accelerated our queries in more than 500 times: 130 milliseconds instead of 72 seconds)

So it is capable to handle big row size if the foreignKey is of correct type in order to prevent casting. That should mention it and point to that issue in order to update metadata for this column.

If we would say that it is capable to handle only small sets of data, that would not be **true**, because you can apply this optimization.

doc/translatable.md
```diff
@@ -3,7 +3,9 @@
 **Translatable** behavior offers a very handy solution for translating specific record fields
 in different languages. Further more, it loads the translations automatically for a locale
 currently used, which can be set to **Translatable Listener** on it`s initialization or later
-for other cases through the **Entity** itself
+for other cases through the **Entity** itself.
+
+Please note that **Translatable** behaviour is meant to be used for small projects and prototypes because of performance limitations.
```
l3pp4rd: **NOTE:** that **Translatable** uses string type as foreignKey in order to be compatible with more types of columns.
When your database grows over 10K translated fields, if your **foreignKey** actually an **integer** for better performance update this column type, see #1512 for more details. If the performance gains are not well enough, consider a custom implementation for your translations.

AkenRoberts:

Thank you!

PR #1747, opened 2017-02-09
[Loggable] Float comparison for decimal type

martijnhartlief:
Decimal fields were always marked as changed

### DIFF
--- a/lib/Gedmo/Loggable/LoggableListener.php
+++ b/lib/Gedmo/Loggable/LoggableListener.php
@@ -232,6 +232,18 @@ class LoggableListener extends MappedEventSubscriber
                 continue;
             }
             $value = $changes[1];
+            
+            if ($meta->getTypeOfField($field) == 'decimal') {
+                $mapping = $meta->getFieldMapping($field);
+                $scale = isset($mapping['scale']) ? $mapping['scale'] : 2;
+                // Base $epsilon on decimal scale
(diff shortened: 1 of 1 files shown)


l3pp4rd:
My advice is to google for `floating point madness` better use integers, more even if you deal with money. This comparison you provided cannot be correct. I will just let users to manage it for themselves, if this would be merged, then it would have to be maintained, I do not want such burden for anyone to have

martijnhartlief:
I don't mind writing an alternative solution if that your main problem against merging. 

Just ignoring the field type means it's basically impossible to maintain an application that requires different levels of significance (199.99, 20800.272). Using the decimal ensures the right significance is saved, **especially** for money.

But currently all decimal fields will report as changed if they're not null. What I found out is the value from doctrine is i.e. string "119.00" while the value that is saved is int 119, and somewhere that gets marked as changed.

The placement inside the LoggableListener was to not intervene with any parts that are depending on this behaviour.

martijnhartlief:
After reflection I think I realise what you mean; That my implementation may only be good for X percentage of the user base. 

This is a side effect of Doctrine and not Loggable per se. So the way to deal with this could be different for everyone. 

Is that what you're trying to make clear?

l3pp4rd:
yes exactly, floating point values in PHP and doctrine is a mess. and is handled differently by many users. You could try to implement a custom database type. I think loggable stores it based on the value database type converts it to.

PR #1749, opened 2017-02-13
[Sluggable] Fix query when identifier is a embedded value holder

cmodijk:
We use a value holder class as one of our entities `identifiers` and this generates a problem when finding the unique value of a slug. The ORM class generates a SQL statement to find al Similar slugs. But then it try's to generate a named variable with a . inside of it and that is not allowed by doctrine. 

```
SELECT rec.slug FROM CoreDomain\Category\Category rec WHERE rec.slug LIKE :slug AND rec.id.uuid <> :id.uuid
``` 

```php
class Category
{
    /**
     * @var CategoryId
     *
     * @ORM\Embedded(class="CoreDomain\Category\CategoryId", columnPrefix=false)
     */
    private $id;
}
```

```php
/**
 * @ORM\Embeddable
 */
final class CategoryId
{
    /**
     * @var UuidInterface
     *
     * @ORM\Id
     * @ORM\Column(type="uuid")
     */
    private $uuid;
}
```

error


l3pp4rd:
thanks ;)

PR #1750, opened 2017-02-13
[Translatable] Allow to save an empty string translation

TriAnMan:
Fixes Atlantic18/DoctrineExtensions#788


l3pp4rd:
well, this is an ugly if statement already..

PR #1751, opened 2017-02-15
Remove unnecessary variable from example

oconnedk:
Undeclared variable `$translatable;` in the code does nothing.


l3pp4rd:
thanks ;)

PR #1753, opened 2017-02-20
Update ClosureTreeRepositoryTrait.php

pavelgajdos:
Bug fix - use name for parent column from config instead of fixed name 'parent'


l3pp4rd:
thanks, but master branch may not be released, I suggest to use 2.4.x

pavelgajdos:
Ok, can you suggest a way of extending a custom entity repository and having your closure tree repository functionality as the same time? The custom repository is a base repository developed by a third party, so there's no way to use it as a trait. Your library is used only for one entity, maybe a few more in the future, but I need to keep the consistency throughout all repositories that all inherit from the mentioned custom repository class. Or are you going to implement traits in 2.x branch? Thank you.

l3pp4rd:
copy the source code in repository. it will not change much in the future.

PR #1754, opened 2017-02-24
Add TreeRoot support to MaterializedPath trees

BPScott:
This commit allows the TreeRoot() annotation to be used with trees using
the MaterializedPath type. It stores the first value in the ancestry
field into the root field. The field decorated with TreeRoot() may be
used as an association and thus act as a foreign key if it is configured
to be a reference rather than an integer field.


l3pp4rd:
that is great, thanks for contributing ;)

PR #1761, opened 2017-03-09
[SoftDeleteable] Easily allow users to perform selects of ONLY soft deleted entities.

malc0mn:
Added a **MySQL** filter `SoftDeleted` that does the exact opposite of the `SoftDeleteable` filter: it only selects soft deleted entries!

This is useful to build admin only views in applications so that admins can actually see all data ;-)

Needed this for an app I'm building and it seems generic enough to add as a feature to the SoftDeleable behavior.

This one is only for SQL though... Should be easy enough to port to MongoDB

### DIFF
--- a/doc/softdeleteable.md
+++ b/doc/softdeleteable.md
@@ -11,6 +11,7 @@ Features:
 - Can be nested with other behaviors
 - Annotation, Yaml and Xml mapping support for extensions
 - Support for 'timeAware' option: When creating an entity set a date of deletion in the future and never worry about cleaning up at expiration time.
+- Support for **only** selecting **soft deleted** entries by swapping from the `SoftDeleteable` filter to the `SoftDeleted` filter.
 
 Content:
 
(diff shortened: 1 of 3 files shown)


l3pp4rd:
Hi, this is over engineered, just disable the filter and query objects where deleted at column is not null.  in admin views, disabling a filter would query everything, which is also a good option. all you haveto do, is to mark softdeleted items in the lists..

malc0mn:
I beg to differ as the whole point of the doctrine soft delete extension is **not** to have to care about creating special queries: 

> I'ts not what you can do for Doctrine, but what Doctrine can do for you :-p

If you have 2 filters (one to hide deleted stuff and one to show only deleted), you can create a **central** flip switch in your application to instantly toggle between 'modes' without having to care about any query at all...

Don't want soft delete anymore? Drop the bundle, drop the deleted_at cols and all is done: no need to care about adjusting any queries...

**But by all means:** if nobody cares for this functionality then let's deny the pull request. Anyone can easily create such a filter themselves ;)

mlc.

l3pp4rd:
exactly, anyone can create such a filter if they need to. extensions now do not accept new features in the codebase, the reason is because the main maintainer does not work with php anymore and wishes it to remain as stable as it can be, unless there are new maintainers. So I would rather skip such extra features if it is possible to integrate them in different ways

AkenRoberts:

Thank you!

PR #1764, opened 2017-03-13
fix missing blameable service definition in symfony2 doc file

nnmer:


l3pp4rd:
It was not necessary, the example is just an example on how to configure some of extensions, not all are needed in most cases. but fine if that is more obvious

l3pp4rd:
though, master branch is not supported and will not see the day I think. so make a pull request against v2.4.x which should be suggested by default

nnmer:
ok, added PR #1765

PR #1765, opened 2017-03-14
fix missing blameable service definition and it’s usage in DoctrineEx…

nnmer:
fix missing blameable service definition and it’s usage in DoctrineExtensionListener in symfony2 doc file


l3pp4rd:
thanks ;)

PR #1766, opened 2017-03-17
bug with entity column naming

breathman:
An exception occurred while executing 'INSERT INTO application (id, identifier, status, type, createdAt, updatedAt) VALUES (?, ?, ?, ?, ?, ?)' with params [9, "2212", "\u041d\u043e\u0432\u044b\u0439", "\u041a\u0440\u0435\u0434\u0438\u0442", "2017-03-17 18:48:18", "2017-03-17 18:48:18"]:
ORA-00904: "CREATEDAT": invalid identifier"

### DIFF
--- a/lib/Gedmo/Timestampable/Traits/TimestampableEntity.php
+++ b/lib/Gedmo/Timestampable/Traits/TimestampableEntity.php
@@ -16,14 +16,14 @@ trait TimestampableEntity
     /**
      * @var \DateTime
      * @Gedmo\Timestampable(on="create")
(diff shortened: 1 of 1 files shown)


l3pp4rd:
Hi, it is not a bug. The column name is not defined for a good reason, because people might want it in different representation. you have to configure ORM to use an **underscore** naming strategy in order to have it by default the way you like for all your entities. Google orm configuration to find more details how to configure it properly.

breathman:
Hello. Thank you for your comment. I am wrong. I forgot for setup this configure.

PR #1767, opened 2017-03-21
Closure tree - fix incorrect delete

Koc:
Sometimes there is no any ids and error occures - 'DELETE FROM closure WHERE id IN ()' - wrong query


l3pp4rd:
thanks ;)

PR #1768, opened 2017-03-21
nullable slug must not be null if base field has "0"

vetalt:
nullable slug must not be null if base field has "0"

### DIFF
--- a/lib/Gedmo/Sluggable/SluggableListener.php
+++ b/lib/Gedmo/Sluggable/SluggableListener.php
@@ -381,7 +381,7 @@ class SluggableListener extends MappedEventSubscriber
                     $slug = substr($slug, 0, $mapping['length']);
                 }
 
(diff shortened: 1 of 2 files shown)


lib/Gedmo/Sluggable/SluggableListener.php
```diff
@@ -381,7 +381,7 @@ private function generateSlug(SluggableAdapter $ea, $object)
                     $slug = substr($slug, 0, $mapping['length']);
                 }
 
-                if (isset($mapping['nullable']) && $mapping['nullable'] && !$slug) {
+                if (isset($mapping['nullable']) && $mapping['nullable'] && strlen($slug) === 0) {
```
l3pp4rd: ah yes, good old PHP tricks ;)

l3pp4rd:
thanks ;)

umpirsky:
@l3pp4rd Can you tag maintenance release including this fix please?

l3pp4rd:
should be available ;)

PR #1772, opened 2017-03-30
Update Sluggable.php

victorvassilev:
changing description of prefix / suffix, they were incorrectly written


l3pp4rd:
thanks ;)

PR #1776, opened 2017-04-04
Fix string length for unique indexes on utf8mb4 MySQL collation

GavChap:
Updated the string length to make it compatible with utf8mb4 which is now Symfony best practice as per the best practice guide. [Friends Of Symfony](https://github.com/FriendsOfSymfony/FOSUserBundle/issues/2182) are also updating their extensions for compatibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
     },
     "require": {
         "php": ">=5.3.2",
-        "behat/transliterator": "~1.0",
+        "behat/transliterator": "1.1",
         "doctrine/common": "~2.4"
     },
     "require-dev": {
--- a/lib/Gedmo/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
+++ b/lib/Gedmo/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
@@ -44,7 +44,7 @@ abstract class AbstractLogEntry
     /**
      * @var string $objectClass
      *
-     * @ORM\Column(name="object_class", type="string", length=255)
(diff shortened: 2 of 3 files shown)


schodemeiss:
Friends of Symfony UserBundle fixed in this commit:

https://github.com/FriendsOfSymfony/FOSUserBundle/pull/2446/commits

l3pp4rd:
this is a backward incompatible change and most likely won't be available in extensions at all, since I doubt there will be any BC related changes anymore.

### option 1

It is possible to extend the AbstractTranslation class or AbstractLogEntry. in order to override this annotation, since these properties are protected. 

### option 2

since the fields in this example most likely will not have 4 byte unicode characters, you can just use utf-8 collation or even ASCII based one.

GavChap:
Could you explain how Option 1 would work please?

l3pp4rd:
@GavChap see [mapping](https://symfony.com/doc/master/bundles/StofDoctrineExtensionsBundle/index.html#step-3-add-the-extensions-to-your-mapping) instead of registering all entity mapping in Gedmo, you can register only mapped superclasses and create the Translation or LogEntry entity with your own indexes and mapping for columns, all you need is to extend those super classes. There is [symfony](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/doc/symfony2.md#ext-mapping) documentation examples in extensions covering mapping configuration in details. In general bundle just limits access to listeners, which in many cases might be very convenient.

AkenRoberts:
Fixed in `master` for upcoming 3.0 release. Thank you!

PR #1777, opened 2017-04-07
Removed from duplicate root declaration.

franciscomemoli:
Related with https://github.com/Atlantic18/DoctrineExtensions/issues/1710#issuecomment-292451603 issue.
The root field is declared twice, on fields and manyToOne secction.


l3pp4rd:
thanks ;)

PR #1778, opened 2017-04-18
Fix tree root setter for roots with association mapping

### LINKED ISSUE #1571: [Tree] Not correct setter for root node
For  [this ](https://github.com/Atlantic18/DoctrineExtensions/blob/master/doc/transaction-safety.md) example setter of root element works not correctly. Now in the current implementation, Nested set root to self but need to set root to Shop entity.
I think need replace [this string](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Strategy/ORM/Nested.php#L400) to
 `if ($root = $meta->getFieldValue($node, $config['root'])) {
                $newRoot = $root;
            } else {
                if ($meta->isSingleValuedAssociation($config['root'])) {
                    $newRoot = $node;
                } else {
                    $newRoot = $wrapped->getIdentifier();
                }
            }`

BoShurik:
Fix #1571

### DIFF
--- a/lib/Gedmo/Tree/Strategy/ORM/Nested.php
+++ b/lib/Gedmo/Tree/Strategy/ORM/Nested.php
@@ -388,10 +388,11 @@ class Nested implements Strategy
                 $wrapped->setPropertyValue($config['right'], $right);
             }
             $newRoot = $parentRoot;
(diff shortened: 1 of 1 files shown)


l3pp4rd:
thanks ;)

PR #1781, opened 2017-04-23
Possibility to use sluggable with embedded entites

dgafka:
This allows to use of embedded classes with sluggable annotations


l3pp4rd:
thanks ;)

PR #1783, opened 2017-04-26
[sluggable] updates transliterator version

garak:


l3pp4rd:
sadly it is not backward compatible. may end up very badly for legacy projects if they decide to upgrade. though, it would be great if that would be possible to override it somehow for new projects. but I'm not aware how that could be done right

garak:
I can't find any evidence of BC in [changelog](https://github.com/Behat/Transliterator/blob/master/CHANGELOG.md).
Maybe `^1.1 || ^1.2` could be better?

l3pp4rd:
tests are failing for that reason https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/225940911
if we make it `^1.1 || ^1.2` composer will likely choose 1.2.

garak:
Since that failing test is depending on the library, could we alter such test to be flexible and expects different results? For example, if library is version <= `1.1`, expect _this_, otherwise expect _that_.

l3pp4rd:
If we change this to `1.2` and update the test case, some users may still get some BC breaks, depending on how they were using this library.

But, maybe we can try to leave this change without tagging release for a while. Could you update the test case in another commit?

PR #1790, opened 2017-05-15
fix tree hydrator proxy

mssimi:
Hey, when calling hydrator multiple times, sometimes there r proxy classes in [data](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Tree/Hydrator/ORM/TreeObjectHydrator.php#L56). When this happens [getConfiguration](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Tree/Hydrator/ORM/TreeObjectHydrator.php#L57) returns empty array and hydrator throw 'The `parent` property is required for the TreeHydrator to work'. I made minor fix for this issue. Please let me know if u have any questions.


l3pp4rd:
makes sense, thanks!

PR #1791, opened 2017-05-26
[Loggable] Allow for Embedded documents to also have child documents

cmodijk:
In the current implementation if a embedded class has  child embedded class the fields for this child's do not get logged. This is solving that problem. I'm not sure how to update and test the other mappings here.


l3pp4rd:
thanks ;)

PR #1795, opened 2017-06-14
Add method to revert Loggable entity to previous version(s)

Fuco1:
Do you think such a feature is reasonable?

I will add tests if I get a good-to-go so that I won't waste effort if the patch won't be accepted.

### DIFF
--- a/lib/Gedmo/Loggable/Entity/Repository/LogEntryRepository.php
+++ b/lib/Gedmo/Loggable/Entity/Repository/LogEntryRepository.php
@@ -64,6 +64,34 @@ class LogEntryRepository extends EntityRepository
     }
 
     /**
+     * Reverts given $entity to $n-th previous revision.
+     *
+     * @param object $entity
+     * @param integer $n
+     *
+     * @throws \Gedmo\Exception\UnexpectedValueException
+     *
+     * @return void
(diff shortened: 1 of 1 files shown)


l3pp4rd:
Hi, I think people can add similar behavior to their repositories. it already provides a method for reverting that is enough for general purpose, this is just for convenience, if we add a bunch of that kind of methods it will be confusing to find and understand generic ones

Fuco1:
Thanks for the reply, I tend to agree, here's my reasoning for this change:

I expected the `revert` method to go to the previous version by default but it goes to the first one instead.  That is why I created this as a separate method.  I think it would be best to change the default to go -1 but that will break BC if someone relies on that (but I can't imagine who would ever want to revert to the first revision, that is very very unlikely).

Anyhow, we don't need to accept this if you think it will make the API less clear.  Feel free to close if this is the final decision.

AkenRoberts:
@Fuco1 Thank you for the pull request, and apologies for it being ignored for a while.

I like this idea for the future. I agree that I would assume "revert" meant go back one version, not to the beginning (which might be "reset").

Going to close this specific PR, though, until we can get this library back on track and deal with some of the larger looming issues first.

PR #1799, opened 2017-06-29
Ignore tests on git export

guiwoda:
This prevents composer downloading the lib's test suite on dependent projects.


l3pp4rd:
thanks ;)

PR #1801, opened 2017-07-03
Mongodb tree nested strategy

alitvinenko:
I added MongoDB Nested set strategy and appropriate Repository class.
As a basis, I took the appropriate classes for ORM.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,12 @@ php:
   - 5.6
   - 7.0
 
-matrix:
-  allow_failures:
-    - php: 7.0
-
 services: mongodb
 
 before_install:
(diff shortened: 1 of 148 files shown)


l3pp4rd:
Hi, I do not think that nested set strategy is feasible for mongoDB. first you should open an issue to discuss the big changes whether these can be accepted. Master branch is also obsolete.

The reason why nested set is not a good fit for mongo db, is that it does not support ACID transactions, you can safely update only one document at the time in mongo db or most key value databases. Nested set requires to update left right values for separate documents and there may be thousands of affected tree nodes by a single update, mongodb will update each of them without a global lock and during that time if a concurrent request would be made, the tree would be broken.

I cannot accept a change which cannot ensure consistency, people would suffer from such tree implementation and open issues which I will not be able to resolve.

PR #1803, opened 2017-07-03
Substitute Date annotation with Field(type="date")

jeremyjumeau:
User Deprecated: Doctrine\ODM\MongoDB\Mapping\Annotations\Date will be removed in ODM 2.0. Use `@ODM\Field(type="date")` instead.


l3pp4rd:
thanks ;)

PR #1812, opened 2017-07-29
[Loggable] ODM query performance fix using compound index

robholmes:
Adds a compound index to ODM `LogEntry`, to optimise query performance, especially for `getNewVersion` query.

Fixes https://github.com/Atlantic18/DoctrineExtensions/issues/1811

### DIFF
--- a/lib/Gedmo/Loggable/Document/LogEntry.php
+++ b/lib/Gedmo/Loggable/Document/LogEntry.php
@@ -2,12 +2,20 @@
 
 namespace Gedmo\Loggable\Document;
 
-use Doctrine\ODM\MongoDB\Mapping\Annotations\Document;
+use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoODM;
 
 /**
  * Gedmo\Loggable\Document\LogEntry
  *
- * @Document(repositoryClass="Gedmo\Loggable\Document\Repository\LogEntryRepository")
+ * @MongoODM\Document(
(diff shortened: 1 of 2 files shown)


l3pp4rd:
Hi please add index not to the mapped superclass but to the LogEntry class, because the superclass is meant to be used with customizations including indexes.

robholmes:
Hi @l3pp4rd, I am unsure about the reason for moving it, as for ODM all other indexes are clearly defined in the `AbstractLogEntry` class (`MappedSuperclass`).

I'm not sure it makes sense to put just one of the 
core indexes in the `LogEntry` class, doesn't it make sense to keep these together?

l3pp4rd:
https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Loggable/Document/LogEntry.php

has no indexes, so instead of superclass, place index to this class. the point is to have it overridable. same as for example translatable https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Translatable/Document/Translation.php has indexes in this instead of mapped superclass.

robholmes:
@l3pp4rd I see what you mean now, I have fixed this and moved all index definitions from the MappedSuperclass to the `LogEntry` class.

l3pp4rd:
thanks ;)

robholmes:
@l3pp4rd no problem, just glad I could contribute :+1:

PR #1819, opened 2017-08-16
Translatable Mapping Adapter fix to be able to use second level cache for translations

AleksSem:
There is one small problem with ORM Adapters under
lib/Gedmo/Translatable/Mapping/Event/Adapter
Custom queries does not support doctrine second level cache (needs object instead of array) 

http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/second-level-cache.html

With this fix you can create custom TranslationEntity which extends AbstractTranslation class add annotation and cache will work
>     cache:
>         usage : READ_ONLY
>         region : region_transaltions

**_My solution is not the best but it helps._**

### DIFF
--- a/lib/Gedmo/Translatable/Entity/MappedSuperclass/AbstractPersonalTranslation.php
+++ b/lib/Gedmo/Translatable/Entity/MappedSuperclass/AbstractPersonalTranslation.php
@@ -9,7 +9,7 @@ use Doctrine\ORM\Mapping as ORM;
  *
  * @ORM\MappedSuperclass
  */
-abstract class AbstractPersonalTranslation
+abstract class AbstractPersonalTranslation implements TranslationInterface
 {
     /**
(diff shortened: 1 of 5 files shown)


l3pp4rd:
Hi, I do not think it is the right time to implement this caching feature in your case, it is experimental and may be removed. The reason why an array result was used for translations, is due to performance gains, translating to objects takes reasonably longer. Since there aren't many users using experimental cache features and instead all other users will have reduced performance after this change.
In result, it does not seem to me like a good change. I suggest you to fork it, if you have performance gains due to this change.

AleksSem:
Hi, 
Maybe there is another way how to implement a caching? 
Add ability to inject some kind of provider (Doctrine\Common\Cache\CacheProvider)?

AkenRoberts:

Thank you!

PR #1823, opened 2017-08-29
[fix/question] Add the call to the Doctrine UnitOfWork class 

clementtalleu:
Why does we not have an `use Doctrine\ORM\UnitOfWork;` call in this class ? 

The UnitOfWork object is used as parameter in many methods of this class


l3pp4rd:
thanks ;)

clementtalleu:
thanks @l3pp4rd ! do you have an idea in how many times we could have a release with this fix ?  

no pressure it's just a question 😄 . Thanks for the bundle, have a good day !

PR #1825, opened 2017-09-06
Tests: remove manual env checks, update PHPUnit

Slamdunk:
Currently manual checks are creating false positive:
https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/269621530#L306

Because `die()` exists with zero status code, which is considerer a passing build.

We could use `exit(1)`, but I consider harmful to manualy check environment: is is much safer to let the CI fail, otherwise false positive is much easier to occur, like this one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ matrix:
         - php: 7.1
         - php: 7.2
 
+cache:
+    directories:
+        - $HOME/.composer/cache
+
 services: mongodb
 
 before_install:
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "doctrine/orm": ">=2.5.0",
         "doctrine/common": ">=2.5.0",
         "symfony/yaml": "~2.6|~3.0|~4.0",
-        "phpunit/phpunit": "*"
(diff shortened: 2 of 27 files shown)


l3pp4rd:
can you rebase and try again, travis should indicate success, otherwise something is wrong here..

Slamdunk:
Rebased. Now there is no more fail positive, and in fact some test fails because PHPUnit is not aligned. If you merge this PR I can create a new one that fixes this (I prefer to do this in separate PR).

l3pp4rd:
I would prefer to have it fixed in the same PR, otherwise other pull requests may be false positive for php 7

Slamdunk:
Ok. Give me some time though.

l3pp4rd:
all the time you need ;)

Slamdunk:
@l3pp4rd here you are

- [x] All tests run in all PHP versions
- [x] PHPUnit runs in 3 versions: ^4.8, ^5.7, ^6.5
- [x] All tests pass

l3pp4rd:
amazing, thanks a lot, was struggling with phpunit maintenance ;)

PR #1826, opened 2017-09-11
Fix phpunit version

enumag:


l3pp4rd:
```
Problem 1
    - phpunit/phpunit 5.7.9 requires php ^5.6 || ^7.0 -> your PHP version (5.4.45) does not satisfy that requirement.
```
We support older versions at the moment. for PHP 7, travis uses a different `composer7.json` file for dependencies

l3pp4rd:
guess this was fixed

PR #1827, opened 2017-09-11
Allow to configure hard deletion

becoded:
Allow to configure hard deletion.
By default, hard deletion is active to be backwards compatible.

### DIFF
--- a/doc/softdeleteable.md
+++ b/doc/softdeleteable.md
@@ -79,6 +79,8 @@ must be of any of the date types.
 Available configuration options:
 - **fieldName** - The name of the field that will be used to determine if the object is removed or not (NULL means
 it's not removed. A date value means it was removed). NOTE: The field MUST be nullable.
+- **timeAware** A boolean making soft-deleted entities time aware (date value for the _fieldName_ in the future won't mark the entity as deleted). Default: false
+- **allowHardDelete** - A boolean specifying if deleting a soft-deleted entity will hard delete it. Default: true.
 
 **Note:** that SoftDeleteable interface is not necessary, except in cases where
 you need to identify entity as being SoftDeleteable. The metadata is loaded only once then
--- a/lib/Gedmo/Mapping/Annotation/SoftDeleteable.php
+++ b/lib/Gedmo/Mapping/Annotation/SoftDeleteable.php
@@ -20,4 +20,7 @@ final class SoftDeleteable extends Annotation
 
     /** @var bool */
     public $timeAware = false;
+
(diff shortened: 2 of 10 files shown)


l3pp4rd:
Hi, what is exactly about? can you explain what this feature adds and why it is needed?

becoded:
Hi,
This is related with the current behaviour of softDeletable entities.
The first time you delete an entity, it will be soft deleted, meaning, the _delete_at_ will be filled in.
If you have a soft deleted entity and you remove it again, it will be hard deleted.

Or in code: 
```
$user = $repo->findOneBy(array('username' => $username));
$this->em->remove($user);
$this->em->flush();
// Entity is soft deleted at this point. (deleted_at field is filled in)
$this->em->remove($user);
$this->em->flush();
// Entity is hard deleted => removed from the database at this point.
```

The added options makes it possible to disable hard deletion so you are sure your entities are never deleted from the database. 
```
// with allowHardDelete=false
$user = $repo->findOneBy(array('username' => $username));
$this->em->remove($user);
$this->em->flush();
// Entity is soft deleted at this point. (deleted_at field is filled in)
$this->em->remove($user);
$this->em->flush();
// Entity is still soft deleted
```

l3pp4rd:
well, I do not want to add extra options just in order to add them. it confuses people. because in order to prevent hard deletion, check the `deletedAt` field whether it is null or not. if you do want to auto check that, create an event listener which would not propagate deletion at all, if that field is not null. If you can do it one or another way without that option, do not create such an option.

becoded:
Thanks @l3pp4rd for your answer.
I didn't add this option, just to add it of course. A customer has a huge codebase and for certain flows, soft-deleted entities needs to be included.
To avoid losing any data, they forked this repo and removed the lines where the hard-deletion is done.
I added this flag in an effort to use the main-repos instead of some outdated forks and I also wanted to share it with the community.

It made sense to me to have soft-deletion doing just soft-deletion and not switch suddenly to hard-deletion (as far as I can see, this behaviour isn't documented btw).
As it wasn't the case and I didn't want to break implementations that relies on this, I added the option.

Thanks for the suggestions btw, but I believe that you will have issues with cascades.

l3pp4rd:
I think hardDelete option was added to replace this, should be resolved ;)

becoded:
I'm glad you changed your mind in the end!

l3pp4rd:
well, indeed, made a mistake, anyway glad someone else went the same way to revisit it ;) and it made me think twice, since more users needed such behavior

PR #1828, opened 2017-09-12
Allow immutable date times

TNAJanssen:


l3pp4rd:
I do not mind adding more types if that is necessary, but this looks just crazy to me :) anyway cheers

PR #1835, opened 2017-10-05
Fix #1173: allow personal translations for entities with composite foreign keys (ORM only, not ODM)

### LINKED ISSUE #1173: foreign_key of two column
I have an entity which consists of a primary key on two columns.
It would have to be able to translate this entity but when I use the command "doctrine: schema: update" I have an error message:
Translatable does not support composite identifiers in class.

do you know the solution to use "gemo translatable" with an entity that has a primary key of two columns?
thank you

NicolaF:
Basic support for personal translations for entities with composite foreign keys. Tell me if I missed something.

### DIFF
--- a/lib/Gedmo/Translatable/Mapping/Driver/Annotation.php
+++ b/lib/Gedmo/Translatable/Mapping/Driver/Annotation.php
@@ -4,6 +4,7 @@ namespace Gedmo\Translatable\Mapping\Driver;
 
 use Gedmo\Mapping\Driver\AbstractAnnotationDriver;
 use Gedmo\Exception\InvalidMappingException;
+use Gedmo\Translatable\Entity\MappedSuperclass\AbstractPersonalTranslation;
 
 /**
  * This is an annotation mapping driver for Translatable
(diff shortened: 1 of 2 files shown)


l3pp4rd:
Hi, I do not think we can integrate this, having composed foreign key support brings much complexity in the library. Not all SQL drivers support it too, plus I doubt that it will work without issues. and since I only maintain v2.4 branch of this repository (I do not wish any new features added here in this library). If some new maintainers comes around, they can do new feature releases. I simply do not work with PHP anymore and having new features with new bugs is the least I'd like to have. That is the sad truth about it.

NicolaF:
Hi,

I hear your point, but it is a really limited change IMHO. It only concerns personal translations, and all that was needed was to help doctrine a bit by doing the joins manually.

It would be a nice enhancement because the idea behind personal translations is to have more control on the mapping.

Moreover, It looks like all platforms currently supported by doctrine definitely handle composite primary keys.

Best regards,
NicolaF_

l3pp4rd:
I cannot add support for something which cannot be fully supported by maintainers, if you need it badly, fork extensions, that is the only option, composite keys will not be supported in this library

NicolaF:
Hi @l3pp4rd,
That's what we've done :(

PR #1838, opened 2017-10-12
More effective way to get column values

Cosmologist:
More effective way to load column values for Gedmo\Tree\Strategy\ORM\Closure::updateNode


l3pp4rd:
thanks ;)

PR #1847, opened 2017-11-15
[Loggable][Bug] Avoid performing a type conversion

loalf:
`object_id` is of type `varchar`, whereas the parameter is passed as an integer. This forces MySQL to perform a type conversion from integer to string which has a performance impact in the SQL query.

Please, take a look at the following example where the `ext_log_entries` table had 37410 records.

captura de pantalla 2017-11-15 a las 11 12 53

captura de pantalla 2017-11-15 a las 11 12 21

In the first image (with the changes proposed in this PR), the query is blazing fast whereas in the second (the code with no changes) it takes longer because of the performance impact to convert the integer into a string.


l3pp4rd:
thanks ;)

loalf:
my pleasure, thank you for making this library awesome!

PR #1861, opened 2017-12-20
Add an option to enable/disable hard delete

babeou:
Soft delete works like this:
- first call: it sets the fieldname with a new DateTime
- second call: it hard deletes your record

## Problem
I'd like to be able to disable the hard delete but there's no configuration to do it.

## Solution
I suggest to add a parameter in order to disable the hard delete.
By default I let the hardDelete parameter to true in order to keep the default behaviour.

Hope you'll like it.

### DIFF
--- a/doc/softdeleteable.md
+++ b/doc/softdeleteable.md
@@ -11,6 +11,7 @@ Features:
 - Can be nested with other behaviors
 - Annotation, Yaml and Xml mapping support for extensions
 - Support for 'timeAware' option: When creating an entity set a date of deletion in the future and never worry about cleaning up at expiration time.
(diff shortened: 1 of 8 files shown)


l3pp4rd:
Hello, thank you, would you be able to add the [documentation](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/doc/softdeleteable.md) for this new option?

babeou:
You've right, that's done!

l3pp4rd:
thanks, that is a nice fluent contribution ;)

enricog84:
@babeou it seems that this PR introduced a bc break, see #1864

PR #1865, opened 2018-01-08
Changed default for hardDelete option to 'true'

### LINKED ISSUE #1864: [SoftDeleteable][Bug] BC break introduced with hard_delete option
In contrary to the docs the default for the **hardDelete** option is set to **false** instead of **true**.   
This leads to backwards incompatibility, because it the option is not explicitly set in the config, no hard delete is ever done.   

I am using the YAML configuration, after adding the **hard_delete: true** it did remove it properly again.

BC break was introduced with @ #1861 .
 
My guess is inside the **SoftDeleteable/Mapping/Driver/\*.php** files the line 

` $config['hardDelete'] = false;`

should be changed to 

` $config['hardDelete'] = true;`

enricog84:
Fixes #1864 by setting the hardDelete option to true by default.


l3pp4rd:
thanks ;) will make a release

PR #1868, opened 2018-01-11
[SoftDeleteable] Detach soft-deleted objects from entity manager after flush

natepage:
When soft-deleting an object, it is still available in the entity manager when retrieving by id.

Example:
```php
// ...

// Retrieving user
$user = $this->entityManager->getRepository(User::class)->find($userId);

// Soft-deleting user
$this->entityManager->remove($user);
$this->entityManager->flush();

// Retrieving soft-deleted user by id
$user = $this->entityManager->getRepository(User::class)->find($userId); // User object

// You need to clear the entity manager cache to force it to go back to the database and use the filter SQL
$this->entityManager->clear();

// Retrieving soft-deleted user by id again
$user = $this->entityManager->getRepository(User::class)->find($userId); // null
```

This PR fixes the issue by detaching soft-deleted objects using the `postFlush` Doctrine event.

To maintain backwards compatibility a config option `detachOnDelete` has been added to SoftDeleteable which by default won't detach the soft-deleted objects from the entity manager.


l3pp4rd:
Hi, since this is more like a bug, I do not see valid reasons to have such option. could you remove it?

sjdaws:
Which option should be removed @l3pp4rd?

l3pp4rd:
detachOnDelete, the one created in pull request

sjdaws:
Even though it’s a bug the option allows the change to be backwards compatible in the off chance anyone is expecting the current behaviour. Are you sure you want it removed?

l3pp4rd:
in my opinion, bugs should not be backwards compatible. I would prefer to not have this option, since it will just confuse users, if the entity is softdeleted, it should not find it cached in entity manager and the code should be able to cope with not finding the entity. I do not think that may cause issues to users.

natepage:
I've removed the option so now all soft-deleted objects are detached from the entity manager.

lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -29,6 +29,13 @@ class SoftDeleteableListener extends MappedEventSubscriber
      */
     const POST_SOFT_DELETE = "postSoftDelete";
 
+    /**
+     * Objects soft-deleted on flush.
+     *
+     * @var array
+     */
+    private $softDeletedObjects = [];
```
l3pp4rd: please use basic array here, since extension compatibility is with php 5.3, everything else seems OK

l3pp4rd: `= array();` I mean ;)

natepage:
Will do when I have a second :) Thank you

l3pp4rd:
thanks ;)

oojacoboo:
@l3pp4rd (@AkenRoberts) it's saying this was merged into 2.4.x, however I'm not showing this code.  Was this merge never published as a new version?

AkenRoberts:
Looks like it was reverted in #1917.

PR #1870, opened 2018-01-24
[Translatable] Fix replacement of last group by element

iluuu1994:
Some Sql vendors (like Sql Server) require every selected element to be in
the group by clause. Because the translation walker adds additional
columns, those columns must be added to the group by as well. The regex was incorrect and did not replace the very
last element in the group by clause.


l3pp4rd:
Hi, this is a risky change, may break something for other vendors, not sure whether it will not introduce a BC break. but well, cannot know until it happens :)

iluuu1994:
Thanks for merging this. Is there a specific reason you find this dangerous? Sql Vendors that don't require the field to be in the group by should behave the same. And those who do require it would fail in the current version anyway.

PR #1872, opened 2018-01-26
Add support for `DateTimeInterface`

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |no    |
|New feature? |yes   |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

**TODO:**
- [x] Add tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,26 +2,37 @@ language: php
 
 sudo: false
 
-php:
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
+matrix:
+    include:
+        - php: 5.4
+          env: phpunit_exclude_groups=datetimeinterface
+        - php: 5.5
+        - php: 5.6
+        - php: 7.0
+        - php: 7.1
+        - php: 7.2
 
 services: mongodb
 
 before_install:
-  - if [[ "$TRAVIS_PHP_VERSION" = 5.* ]]; then echo 'extension=mongo.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi
-  - if [[ "$TRAVIS_PHP_VERSION" != 5.* ]]; then echo 'extension=mongodb.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi
-  - if [[ "$TRAVIS_PHP_VERSION" != 5.* ]]; then cp composer7.json composer.json; fi
+    - if [[ "$TRAVIS_PHP_VERSION" = 5.* ]]; then echo 'extension=mongo.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi
+    - if [[ "$TRAVIS_PHP_VERSION" != 5.* ]]; then echo 'extension=mongodb.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi
+    - if [[ "$TRAVIS_PHP_VERSION" != 5.* ]]; then cp composer7.json composer.json; fi
 
 install:
-  - composer install --prefer-dist
+    - composer install --prefer-dist
 
 script:
-  - bin/phpunit -c tests/
+    - |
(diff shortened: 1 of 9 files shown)


l3pp4rd:
Hi, open a pull request to v2.4.x branch, the master is not maintained

phansys:
@l3pp4rd, do you know how can I exclude a test from the suite? My attempt doesn't seem to be working: https://github.com/Atlantic18/DoctrineExtensions/pull/1872/files#diff-354f30a63fb0907d4ad57269548329e3R30

.travis.yml
```diff
@@ -2,26 +2,35 @@ language: php
...
+    - if [[ "$TRAVIS_PHP_VERSION" != 5.* ]]; then cp composer7.json composer.json; fi
 
 install:
-  - composer install --prefer-dist
+    - composer install --prefer-dist
 
 script:
-  - bin/phpunit -c tests/
+    - |
+        if [[ ! $test-exclude-groups ]]; then
```
l3pp4rd: try `if [[ ! -z "$test-exclude-groups" ]]; then`

phansys: I'm pretty sure the problem isn't in the condition check, since the `else` condition is being executed for php 5.4. I'll push a commit proving this ASAP.

l3pp4rd: I think `\DateTimeInterface::class` syntax is not supported in php 5.4

l3pp4rd: because php fails simply with parse error, use full class name as string and check whether is sublass or whatever ;)

phansys: Yes, it isn't supported. But this test is intended to be run with a version of PHP with support for `\DateTimeInterface`, so `::class` resolver must be available. Do you think we should change the test group name to something pointing to PHP version ("php55-or-greater" by instance) instead?

phansys: I see with the latest commit that the executed command is `Debug: 'bin/phpunit -c tests/ --exclude-group -exclude-groups'`. I really don't know why Travis is stripping the "test" string from `$test-exclude-groups` environment variable and using the name instead of resolving it AS IS. I'll try to use another name for this variable in order to check if the result changes.

l3pp4rd: use `${phpunit-exclude-groups}` a dash in variable name is not something you should usually use, since it works as a separator, better use underscores, dealing with bash is not that simple ;)

l3pp4rd: but better use underscores instead of dash, will prevent more issues

phansys: Great @l3pp4rd! Thank you. Good to know, issue seems to be solved. I'll squash my commits now.

lib/Gedmo/Sluggable/SluggableListener.php
```diff
@@ -298,7 +298,8 @@ private function generateSlug(SluggableAdapter $ea, $object)
                         $needToChangeSlug = true;
                     }
                     $value = $meta->getReflectionProperty($sluggableField)->getValue($object);
-                    $slug .= ($value instanceof \DateTime) ? $value->format($options['dateFormat']) : $value;
+                    // Remove `$value instanceof \DateTime` check when PHP version is bumped to >=5.5
+                    $slug .= ($value instanceof \DateTime || $value instanceof \DateTimeInterface) ? $value->format($options['dateFormat']) : $value;
```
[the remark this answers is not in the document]
phansys: Your concern is about using `instanceof` operator with a non existing class? https://3v4l.org/RFfXE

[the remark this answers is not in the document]
l3pp4rd: ah, OK then guess can be merged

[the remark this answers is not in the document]
phansys: Thank you @l3pp4rd!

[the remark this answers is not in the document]
l3pp4rd: cheers ;)

PR #1875, opened 2018-02-07
added missing getAncestors method

max107:


l3pp4rd:
master will not be supported, can be closed, only v2.4.x branch

max107:
Okay

PR #1877, opened 2018-02-10
Feature/tree root

delboy1978uk:
image

Now I have set `@Gedmo\TreeRoot(identifierMethod="getMenu")`, so it can restrict and renumber by the rows with a menu_id of x (the real tree root, a row in another table).

After running the repository method `moveUp(7)` (the action movie category, in the above example, third out of 3 movie categories), instead of an exception being thrown, now we get:

image


delboy1978uk:
Oh yeah, i also changed to PSR-4, and moved the source code to the `src/` folder standard, which won't affect anything :-)

delboy1978uk:
I tidied up the pull request, code is back in the lib folder so you should be able to review changes easily now :-)

l3pp4rd:
yes, I would not accept the `lib` to `src` change since that would involve BC breaks, since some mapping was directly referenced to directories on configuration for some projects.

could you also add a unit test for this new feature?

delboy1978uk:
Yeah sure, I can't do it today, but I'll work on it and give you another shout. Regarding the lib folder, we are now on PSR-4 anyway 😃 (just not in `src/`)

delboy1978uk:
Hi @l3pp4rd I have tested the new functionality and updated the `tree.md` docs :-)

l3pp4rd:
thanks for all the effort, hope it will be useful for all users ;)

PR #1889, opened 2018-03-09
Change wording to be a bit more explicit with @Versioned

cedricziel:
This change simply improves the error message in the event that a `Versioned` behaviour cannot be applied so it's understood a little easier.


l3pp4rd:
thanks ;)

cedricziel:
Thx for the quick merge!

PR #1905, opened 2018-04-06
Add row_format=dynamic to ext_log_entries and ext_translations 

### LINKED ISSUE #1904: Index column size too large for Translatable bundle
When trying to use the Translatable bundle with MariaDB, I hit the following error:

Fatal error: Uncaught PDOException: SQLSTATE[HY000]: General error: 1709 Index column size too large. The maximum column size is 767 bytes. in /var/www/myapp/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/ToolsException.php on line 34

This relates to this:

https://github.com/Atlantic18/DoctrineExtensions/blob/a43f57c4a3d0ae4241096b69f8f57ef98bc6a8e9/lib/Gedmo/Translatable/Entity/Translation.php#L13-L21

And this issue with MariaDB < 10.2.2 or MySQL < 5.7:
https://stackoverflow.com/questions/42043205/how-to-fix-mysql-index-column-size-too-large-laravel-migrate

The corresponding documentation is available here https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_large_prefix and here https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_large_prefix and, basically, for it too work in versions lower than 10.2.2 and 5.7, the create table needs to be defined as DYNAMIC or COMPRESSED row format.

Something that would translate to:
```
ALTER TABLE tbl ROW_FORMAT=DYNAMIC;
```

Would it be possible to ask for this change to be added, so Translatable also works for these previous versions of MySQL and MariaDB? Basically, considering your bundle can be loaded in many different environments, it would be really cool to support these, and as far as I understand this could be done relatively easily (although it is mysql-specific)

ywarnier:
...to avoid Index column size too large on MySQL < 5.7 (must be combined with innodb_large_prefix=on option in the MySQL server config) - fixes #1904

### DIFF
--- a/lib/Gedmo/Loggable/Entity/LogEntry.php
+++ b/lib/Gedmo/Loggable/Entity/LogEntry.php
@@ -9,6 +9,7 @@ use Doctrine\ORM\Mapping as ORM;
  *
  * @ORM\Table(
  *     name="ext_log_entries",
+ *     options={"row_format":"DYNAMIC"},
  *  indexes={
  *      @ORM\Index(name="log_class_lookup_idx", columns={"object_class"}),
  *      @ORM\Index(name="log_date_lookup_idx", columns={"logged_at"}),
--- a/lib/Gedmo/Translatable/Entity/Translation.php
+++ b/lib/Gedmo/Translatable/Entity/Translation.php
@@ -12,6 +12,7 @@ use Doctrine\ORM\Mapping\Entity;
  *
  * @Table(
  *         name="ext_translations",
+ *         options={"row_format":"DYNAMIC"},
  *         indexes={@Index(name="translations_lookup_idx", columns={
  *             "locale", "object_class", "foreign_key"
  *         })},


l3pp4rd:
Hi, is this annotation supported since doctrine 2.4 ? just concerned about BC

ywarnier:
Hi @l3pp4rd I'm not sure. Not a Doctrine specialist, but see:

https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L497

and the blame view here:

https://github.com/doctrine/dbal/blame/f8e884c628da185b692240e3b03d285ecce054c6/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L497

That seems to say that the function to manage options in general has been there since 5 years ago.
Considering Doctrine 2.4 has been released in September 2013, and that the commit adding the function to support it dates from 21 Feb 2013, it seems that yes, this was added to 2.4.0.

And yes, in the last commit of the v2.4.0 release, the corresponding code exists:
https://github.com/doctrine/dbal/blob/3eb557f144ec41e6e84997e43acb1946c92fbc88/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L443

Or did you mean something else?

l3pp4rd:
thanks ;)

PR #1906, opened 2018-04-11
Show the status for the same branch as the build

tomasliubinas:
Currently the build status image is shown as failed even if the build passes. This happens as image source points to different branch than the build (namely master)


l3pp4rd:
thanks ;)

PR #1911, opened 2018-04-19
[tree] Test for property existence

nclavaud:
ref https://github.com/Atlantic18/DoctrineExtensions/issues/1910

Instead of testing the property value, test its existence.


nclavaud:
@l3pp4rd any thought/feedback on this?

blackandred:
When there will be a tag with this change?

maxolasersquad:
This fix has been in for about 2 months. Is a new tagged version coming soon?

PR #1917, opened 2018-05-08
Revert "[SoftDeleteable] Detach soft-deleted objects from entity manager after flush"

### LINKED ISSUE #1912: Entity persistance error
Hello,

When I update DoctrineExtensions from `v2.4.33` to `v2.4.34` I have this error:

```
Doctrine\ORM\ORMInvalidArgumentException: Entity #1 is not managed. An entity is managed if its fetched from the database or registered as new through EntityManager#persist
```

My code looks like:

```php
$entity = $repository->findOneById(1);
$entityManager->remove($entitty);
$entityManager->flush();
$entityManager->refresh($entity);
```

The entity is loggable, soft deleteable with some versionned fields.

l3pp4rd:
Reverts Atlantic18/DoctrineExtensions#1868

causes internal failures to proxies and other unexpected behavior, closes #1912

### DIFF
--- a/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
+++ b/lib/Gedmo/SoftDeleteable/SoftDeleteableListener.php
@@ -30,13 +30,6 @@ class SoftDeleteableListener extends MappedEventSubscriber
     const POST_SOFT_DELETE = "postSoftDelete";
 
     /**
-     * Objects soft-deleted on flush.
-     *
-     * @var array
-     */
-    private $softDeletedObjects = array();
-
-    /**
      * {@inheritdoc}
      */
     public function getSubscribedEvents()
@@ -44,7 +37,6 @@ class SoftDeleteableListener extends MappedEventSubscriber
         return array(
             'loadClassMetadata',
             'onFlush',
-            'postFlush'
         );
     }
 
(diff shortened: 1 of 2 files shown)


sjdaws:
@l3pp4rd can we put this back [with the option](https://github.com/Atlantic18/DoctrineExtensions/pull/1868#issuecomment-356861253) as we originally had it in #1868? Reverting this breaks `find()` functionality when using soft deletes.

l3pp4rd:
this causes unexpected behavior with proxies sadly, see the issue. If I add this option, then people who will be using it, will go through the same issues, that is not acceptable. Filter your `find()` queries by checking whether it is softdeleted or fork extensions. I will not add any other feature in extensions, unless someone else is wishing to maintain it, it is too hard for me now, since I haven't used php for more than 3 years now :)

sjdaws:
@l3pp4rd thanks for the reply. This is the reason we initially added an option to preserve the current state in the original pull request, by default we set the option to `false` which enforced the status quo allowing people who want reverse functionality to change this setting. Can we merge the option which has the setting as we originally intended?

l3pp4rd:
my point is, that if there is an option, someone will use it, and since there are issues with that option, which I can't and will not resolve, then it will result in general extension library issue. I will not accept any features, which cannot be stable in no matter what way people will be using it. Currently extensions already has enough bugs creeping and I wish it to get into the stable version in the end without new features being added.

sjdaws:
Ok, thanks @l3pp4rd.

PR #1924, opened 2018-06-28
Fix calculating the change set data for embedded collections

carlos-granados:
Setting the `$wrapped` variable within the loop is not necessary to calculate the change set and setting it will make the code not work properly for the next properties


carlos-granados:
Some tests are failing but those failures seem completely unrelated to this PR

l3pp4rd:
note, the master branch is not maintained, I suggest to migrate patches needed to v2.4.x branch

PR #1926, opened 2018-07-04
[Sortable] Fix issue with position synchronization after multiple deletes

naitsirch:
In the `postFlush` method of the `SortableListener` the position of objects get updated via `$this->setFieldValue()` after applying the relocation. The problem is that `$this->setFieldValue()` creates a change set for the passed object and that will prevent from other relocations being executed on this object.

I have added a unit test that tests this case and extended the behaviour of `postFlush` so that all relocations for one object will get applied.

### DIFF
--- a/lib/Gedmo/Sortable/SortableListener.php
+++ b/lib/Gedmo/Sortable/SortableListener.php
@@ -413,6 +413,9 @@ class SortableListener extends MappedEventSubscriber
     {
         $ea = $this->getEventAdapter($args);
         $em = $ea->getObjectManager();
+
+        $updatedObjects = [];
+
         foreach ($this->relocations as $hash => $relocation) {
             $config = $this->getConfiguration($em, $relocation['name']);
             foreach ($relocation['deltas'] as $delta) {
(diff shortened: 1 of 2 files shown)


l3pp4rd:
thanks, I thing that was bugging some users for a while

jkabat:
@naitsirch Im using sortable for many2many relation and I have exactly the same problem. However your fix is not properly working for me - maybe you can shed a light on what can be improved to fix it completely.

When I delete and add items in one operation (select2 form element), order becomes corrupted, because for every removed item update position query is executed. For 3 removed items and 2 added items these queries are used:

```sql
INSERTs... (2x)
UPDATE `table` SET `position` = `position` - 1 WHERE `position` >= 1 AND `group` = 1234;
UPDATE `table` SET `position` = `position` - 1 WHERE `position` >= 2 AND `group` = 1234;
UPDATE `table` SET `position` = `position` - 1 WHERE `position` >= 3 AND `group` = 1234;
DELETEs... (3x)
```

Is it possible to fix it, or this sortable extension is not intended for such usage?
(drag'n'drop approach to sort the set is used, once relations are added using select2)

jkabat:
After some digging it seems to be resolvable. Deletion relocations just have to be aware of newly created entities.

Dirty hack would be adding following lines:
```php
} elseif (-1 == $val['delta'] && 1 == $needle['delta']) {
    $val['exclude'] = array_merge($val['exclude'], $needle['exclude']);
}
```

at https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Sortable/SortableListener.php#L582

@l3pp4rd  I'm not sure this would be accepted as general solution, so I guess anyone with extension knowledge can come up with something more accurate.

l3pp4rd:
Hi I think you can open PR @jkabat if it passes tests and users do not spot issues it is good enough, either way they will submit another fix if there is something wrong. I was not working with extensions like few years now, just trying to get it to the stable point, so that would definitely help

PR #1932, opened 2018-07-26
[Sortable] Fix issue with position synchronization when items are add…

jkabat:
…ed and deleted in one operation

Deletion relocations must be aware of newly created objects in order to generate correct position update queries.

However I kind of don't understand original intention of collecting newly created objects in "excludes" for add relocation. There is most probably something missing and I was not able to unravel it correctly. I would really appreciate if anyone fix this piece the way author has intended to.


l3pp4rd:
thanks @jkabat guess you will make many users happy :)

PR #1936, opened 2018-08-21
Use CURRENT_TIMESTAMP instead of date to allow caching queries in Softdeletable

supersmile2009:
Using CURRENT_TIMESTAMP instead of current datetime value allows queries to be cached.
`$platform->getCurrentTimestampSQL()` provides platform-specific SQL implementation.
This feature is available in doctrine/dbal v2.2, so this PR is backwards-compatible.

```
# Cachable
SELECT *
FROM foo f
WHERE f.deleted_at > CURRENT_TIMESTAMP

# Non-cachable
SELECT *
FROM foo f
WHERE f.deleted_at > ?
["2018-08-19 10:11:12"]
```


l3pp4rd:
thanks ;)

PR #1939, opened 2018-08-23
[RFC] Require and configure PHP CS Fixer

AkenRoberts:
[PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) is a tool for automatically adjusting code styles based on a set of configured rules. I started with a basic configuration:

* `@Symfony` ruleset, which includes PSR-1 & 2 and other rules they prefer. I consider their standards a good starting point, especially considering Doctrine's heavy ties to the framework.
* Alphabetical ordering of `use` namespace statements.
* Force long array syntax (`array()` vs `[]`). This would be reversed when the minimum PHP version is bumped.

There are others that I would personally include, however this is a community project, so I'm PRing it for comments.

A Composer script is there to make running the fixer easier: `composer fix-cs`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ tests/temp/*.log
 /composer.lock
 /composer.phar
 .idea
+.php_cs.cache
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,13 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__.'/lib')
+    ->in(__DIR__.'/example');
+
+return PhpCsFixer\Config::create()
+    ->setRules(array(
+        '@Symfony' => true,
+        'array_syntax' => array('syntax' => 'long'),
+        'ordered_imports' => array('sort_algorithm' => 'alpha'),
+    ))
+    ->setFinder($finder);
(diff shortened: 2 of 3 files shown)


AkenRoberts:
Failing builds should've been expected - PHP CS Fixer has a minimum PHP requirement of 5.6.

andrewtch:
In fact I've never seen cs-fixer as a part of the composer ) I suppose we should add it to Travis and you can set env variable like

```
matrix:
    include:
        - php: 5.4
          env: phpunit_exclude_groups=datetimeinterface dependencies=lowest disable_phpcs=1
        - php: 5.5
          env: disable_phpcs=1
```

and then

```
script:
    - |
        if [[ ! $disable_phpcs ]]; then
            ...
```

AkenRoberts:
It isn't the fixer itself running, it's the `composer install` command failing at the PHP version check. I'm not sure if there's a way to conditionally ignore an individual Composer package.

AkenRoberts:
I think I'll shelve this for 2.x branches, and target the next major release when we bump the PHP version. I don't want to introduce something that can break tests and hide actual regression issues. Unless someone has an idea for how to implement and skip running on PHP < 5.6.

andrewtch:
Actually, noone _requires_ csfixer, see symfony: https://github.com/symfony/symfony/blob/master/composer.json

We can install it conditionally. I will complete one big launch by monday and will update this PR; currently a bit busy.

stephanvierkant:
Good idea to include a coding standard. I personally like the Doctrine coding standard more.

AkenRoberts:
Closing and re-implementing on `master` manually.

PR #1951, opened 2018-09-18
Change \DateTime into \DateTimeInterface in traits

ksowa:
Changed `\DateTime` to wider `\DateTimeInterface`, in order to allow `\DateTimeImmutable` usage, e.g. in tests.

Since traits are available from PHP 5.4, but `\DateTimeInterface` from PHP 5.5, this change implies incompatibility with older PHP versions.

### DIFF
--- a/lib/Gedmo/Timestampable/Traits/Timestampable.php
+++ b/lib/Gedmo/Timestampable/Traits/Timestampable.php
@@ -11,22 +11,22 @@ namespace Gedmo\Timestampable\Traits;
 trait Timestampable
 {
     /**
-     * @var \DateTime
+     * @var \DateTimeInterface
      */
     protected $createdAt;
 
     /**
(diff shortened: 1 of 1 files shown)


AkenRoberts:
Hi @ksowa! Thank you for your first contribution.

Because of the backwards incompatibility, this isn't something we can add to any 2.x release (in respect of semantic versioning). As such, this would target the next major release, 3.0, which at this point is very much up in the air as far as the direction Doctrine Extensions will take. Myself and other new maintainers are currently trying to determine the best path.

Until that can be decided, pull requests like these will have to be closed or held. Targeting `DateTimeInterface` to add further compatibility will definitely be part of that strategy, though!

bt-nn:
Hi @AkenRoberts 
3.0 is out and min php version now is 7.2 - so maybe it could be merged now?

PR #1954, opened 2018-09-26
Small improvements and micro optimizations

supersmile2009:
- `isset` instead of `array_key_exists`
- Removed use of ReflectionProperty used to access protected property from parent class.
- fixed some annotations


stephanvierkant:
Looks good to me, but the tests are failing. Can you take a look at it?

PR #1964, opened 2018-10-23
allow to make a whole embeddable versioned

mneuhaus:
this patch enables a user to put the versioned annotation on the embeddable property itself, so that you can use stuff like  https://github.com/kucharovic/money-bundle and make the whole embeddable versioned without modifiying it.


mneuhaus:
sorry, got something fucked up here. will reopen a fresh PR

PR #1966, opened 2018-10-27
Update .travis.yml

andreybolonin:


AkenRoberts:
Thank you for the PR, @andreybolonin. We definitely do need to add testing support for PHP 7.3.

Unfortunately, as you can see by the Travis build, a majority of the tests are affected by 7.3's new warning RE: `continue` statements in used in `switch` statements.

We'll have to find a way to ensure Travis uses the proper Doctrine ORM version dependency so that the test suite runs as expected. Beyond that, we may even need to update DoctrineExtensions to support 7.3.

We definitely want to support 7.3, and ensure we test against it. However, this update alone is only one step. If you want to expand it in a future pull request (if it hasn't been done yet), it would be greatly appreciated.

PR #1967, opened 2018-10-29
Fix php 7.2 warnings on countables

JurRutten:


lib/Gedmo/Tree/Entity/Repository/ClosureTreeRepository.php
```diff
@@ -315,7 +315,13 @@ public function buildTreeArray(array $nodes)
                     $tmp = &$refs[$n['parent_id']][$childrenIndex];
                 }
 
-                $key = count($tmp);
+                if(is_array($tmp) || $tmp instanceof \Countable) {
+                    $key = count($tmp);
+                } elseif(null === $tmp) {
+                    $key = 0;
```
phansys, line 321:
```suggestion
                if (null === $tmp) {
                    $key = 0;
                } elseif (is_array($tmp) || $tmp instanceof \Countable) {
                    $key = count($tmp);
```
BTW, I'm not using this extension, so I don't know what possible types can be held in `$tmp`. By instance, I don't know if we should even consider `iterator_count()`.

phansys (commented):
Thank you so much for the time spent in this fix @JurRutten, and my apologies for the delay.
Could you please confirm if the same issue is present in `main`? If yes, we should apply this fix in that branch.
Do you mind to provide a test covering these cases, in order to avoid regressions?
The same issue is also addressed in #2145, maybe @gvanbeck could provide some help too.

PR #1970, opened 2018-11-11
[Translatable] Set getTranslatableListener method to protected

CaptainFalcon92:
I was in the need to extends the behaviour of the `findTranslations` method.
But private visibilities prevent me from using this class.
So, i changed `getTranslatableListener` from private to protected. I really don't think that private is the way to go. It seams unnecessary and definitely prevents the class from being extended.
If only protected was the de-facto default keyword people were using... private is almost always a bad choice imho, sorry.


stephanvierkant:
I'm not sure if a protected property is a good idea, but I don't see any problems with changed the method to protected.

PR #1985, opened 2019-01-28
[Loggable] Map nested embeddables

robertburczak:
Possibility to map nested embeddables in yaml mappings.
Thanks @ghettovoice for suggestion

### DIFF
--- a/lib/Gedmo/Loggable/Mapping/Driver/Yaml.php
+++ b/lib/Gedmo/Loggable/Mapping/Driver/Yaml.php
@@ -145,5 +145,14 @@ class Yaml extends File implements Driver
                 $config['versioned'][] = $field . '.' . $property;
             }
         }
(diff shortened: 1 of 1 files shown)


mkosiedowski:
Any chances of merging that PR? :) We are using our fork with that change for over a year now and would love to switch back to original version of DoctrineExtensions... @AkenRoberts?

AkenRoberts:
I'd consider adding to the upcoming 3.0 release in `master`, I'd just need to familiarize myself with what this changes, and if we want to test for it. It doesn't break anything, which is a good start. I just want to avoid adding untested features.

andrekutianski:
Didn't this PR enter version 3?

franmomu:
thanks for the time invested in this PR and sorry for the delay. The Yaml drives has been deprecated, you should use `Attribute`,  `Annotation` or `XML` instead.

PR #1986, opened 2019-01-30
Custom strategy for Tree

Smeagolworms4:
Add method for add custom strategy on TreeListener.

### DIFF
--- a/lib/Gedmo/Tree/TreeListener.php
+++ b/lib/Gedmo/Tree/TreeListener.php
@@ -58,7 +58,7 @@ class TreeListener extends MappedEventSubscriber
 
     /**
      * @param string   $name
-     * @param Strategy $instance
+     * @param Gedmo\Tree\Strategy $instance
      * @return self
      */
     public function addStrategyInstance($name, $instance)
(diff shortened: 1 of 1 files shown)


lib/Gedmo/Tree/TreeListener.php
```diff
@@ -56,6 +56,18 @@ public function getSubscribedEvents()
         );
     }
 
+    /**
+     * @param straing  $name
```
NicoBoos: You misspelled "string"

lib/Gedmo/Tree/TreeListener.php
```diff
@@ -56,6 +56,18 @@ public function getSubscribedEvents()
         );
     }
 
+    /**
+     * @param string   $name
+     * @param Strategy $instance
+     * @return self
+     */
+    public function addStrategyInstance($name, $instance)
```
AkenRoberts: What do you think about type-hinting `Gedmo\Tree\Strategy` here for `$instance`?

Smeagolworms4:
You can merge ?

franmomu:
can you please target `main` branch? parameters and return types should be added and also a test

PR #1987, opened 2019-02-05
Fix for wrong UploadFileName on Append Number (#1229)

### LINKED ISSUE #1229: UploadableListener. Wrong fileName with appendNumber.
When I'm using the appendNumber=true, then the number is added to the filePath, but not the fileName. As I understand, there might be some people who don't want the number to be added to the fileName (as the number is only to distinguish the files on the server, and the end-user shouldn't be aware of it). I on the other hand would want the number to be available in my entity without having to store the filePath in the database.
The idea is that the filePath seems to be an absolute path and I don't want this stored. What if the absolute path of my application changes, then I would need to update every entry in my database. I want to store only the fileName (or a relativePath, but there doesn't seem to be such a thing here). The fileName should represent the file name that is located on disk. 
Although, there can be two annotations, one that specifies the file name on disk and the other that specifies the original file name.

sujayjaju:
Fixes #1229
UploadableFileName matches the name with the appended number with this update.

### DIFF
--- a/lib/Gedmo/Uploadable/UploadableListener.php
+++ b/lib/Gedmo/Uploadable/UploadableListener.php
@@ -594,6 +594,8 @@ class UploadableListener extends MappedEventSubscriber
                 ));
             }
         }
+        
+        $info['fileName'] = basename($info['filePath']);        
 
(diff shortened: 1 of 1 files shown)


franmomu:
Closed in favor of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2460

Issue #1992, opened 2019-02-17
Using UUID as ID type and using tree hydrator does not work because of illegal type


lsv:
As the `$this->getPropertyValue($parentProxy, $this->idField);` is expected not be a object due to it is parsed into as a array key, you can not use `ramsey\uuid` as a id type.

I have already fixed it by adding a `getIdPropertyValue($node)` and then using `__toString` if the id value is a object.

Used in repository

```
$tree = $this
            ->createQueryBuilder('node')
            ->getQuery()
            ->getResult('tree');
```

It seems that `$data = parent::hydrateAllData();` does not fetch the parent, and therefor can not build `$this->buildChildrenHashmap($data)`

But if I remove `'tree'` and use doctrine object hydrator, the parent is the correct object

koioannis:
Hi! I know that this is a bit old, but I'm facing the same issue. As far as I can tell this has not been merged. Is there any other way to fix this issue?

karl-foerster:
I'm facing the same issue trying to implement the tree with symfony/uuid as id. Applying all changes of that linked pull request everything works fine. I have no issues with parent.

PR #1993, opened 2019-02-17
Changed TreeHydrator, so its possible to have the IdField as a object

lsv:
Changed `getPropertyValue(object, $this->idField)` to `getIdPropertyValue(object)`

So if the idField is a object, it is now required that the object has a `__toString` method, otherwise a exception will be thrown.

See #1992


lsv:
It seems that `$data = parent::hydrateAllData();` does not fetch the parent, and therefor can not build  `$this->buildChildrenHashmap($data)`

PR #1996, opened 2019-02-19
[BUG] Doctrine Metadata Cache lifetime should not be passed as null

andrey-bondar-dron:
Signature of \Doctrine\Common\Cache\Cache::save does not accept null (https://github.com/doctrine/cache/blob/6f9810e194acace22e360a1398cf228132e5b4b2/lib/Doctrine/Common/Cache/Cache.php#L55) and pass lifetime as null null works not properly on different cache implementations.
For instance, on \Doctrine\Common\Cache\PhpFileCache.

### DIFF
--- a/lib/Gedmo/Mapping/ExtensionMetadataFactory.php
+++ b/lib/Gedmo/Mapping/ExtensionMetadataFactory.php
@@ -2,6 +2,7 @@
 
 namespace Gedmo\Mapping;
 
+use Doctrine\Common\Cache\Cache;
 use Doctrine\Common\Persistence\Mapping\Driver\DefaultFileLocator;
 use Doctrine\Common\Persistence\Mapping\Driver\SymfonyFileLocator;
 use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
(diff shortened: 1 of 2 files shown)


svycka:
I guess this came from https://github.com/doctrine/cache/blob/67d7149abee291a92591f0c31cd925f8b52cb745/Cache.php#L59 not sure when this happened

dron-alterpost:
@svycka may be it is time to merge this PR? This repository use a new version of doctrine/common in composer.json.

AkenRoberts:
@andrey-bondar-dron Thank you for the PR!

You've labeled this a bug - is there a specific error or unexpected result occurring from this? Anything we can add a test against?

dron-alterpost:
@AkenRoberts Hi! 
When https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/PhpFileCache.php is using as doctrine metadata cache, it assumes, that lifetime cannot be null (as in interface) on https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/PhpFileCache.php#L112 and has unexpected behavior.
I have viewed all implementations in https://github.com/doctrine/cache/tree/master/lib/Doctrine/Common/Cache and they all expect lifetime as integer.

Anyway, interface does not accept null and client code should not pass it (in order to prevent code smells =))

AkenRoberts:
I'm concerned, perhaps unnecessarily, about how changing the default might modify current behavior. Because of that, and how I'm trying to deprecate the 2.4.x branch in favor of a cleaner 3.0, I've implemented this change in `master` via 3a3af134e00d7b7d72a7a59c4c3cfe400ccdd91d.

Thank you for the contribution and approval. Keep it up!

PR #2004, opened 2019-03-19
update gitattributes

rvitaliy:
- remove useless files for production from dist
- configure git to automatic line endings normalized

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,11 @@
+# Auto-detect text files, ensure they use LF.
+* text=auto eol=lf
+
+# Exclude non-essential files from dist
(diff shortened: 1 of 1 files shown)


AkenRoberts:
Thank you for the suggestion. I ended up adding what I thought was useful to `master` via 0dcebe18bfc2c5c709dc6d19c84dd5e8f58b8d8f.

I think keeping docs and examples is a good practice, so that you can reference either next to the production code.

I'm not sure where the benefit of the line endings would come from, so I skipped that until I can better understand why it's useful and what it might affect. If you have any links that would help, I'd appreciate reading them.

PR #2009, opened 2019-04-12
[Loggable] Adds missing string cast, causing indexes not to be used due to incorrect type

rcwsr:
Before:

`EXPLAIN select * from ext_log_entries where object_class = 'My\\Class' and object_id = 2;`

```
id    select_type    table    type    possible_keys    key    key_len    ref    rows    Extra
1    SIMPLE    ext_log_entries    ALL    object_id_and_class    NULL    NULL    NULL    200424    Using where
```

After:
object_id is a varchar:

`EXPLAIN select * from ext_log_entries where object_class = 'My\\Class' and object_id = '2';`

```
id    select_type    table    type    possible_keys    key    key_len    ref    rows    Extra
1    SIMPLE    ext_log_entries    ref    object_id_and_class    object_id_and_class    1025    const,const    836    Using where
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Loggable
+#### Fixed
+- Added missing string casting of `objectId` in `LogEntryRepository::revert()` method
 
 ## [2.4.37] - 2019-03-17
 ### Translatable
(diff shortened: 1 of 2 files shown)


AkenRoberts:
Thank you, @rcwsr. Could you please add an entry to the [changelog](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/CHANGELOG.md). Than we can merge. :)

rcwsr:
@AkenRoberts Added!

rcwsr:
Thanks. Missed that

AkenRoberts:
Merged. Thanks again. 👍

PR #2018, opened 2019-05-22
Return `parent` instead of throwing Exception for Translatable

### LINKED ISSUE #2013: [Translatable][Sortable] Weird error while persisting sortable entity
I use Translatable and Sortable in my project. When trying to persist or remove entity, that uses Sortable but not Translatable, I get weird error. 

Here it is:
baguś

The entity is here: https://github.com/bolt/four/blob/master/src/Entity/Relation.php

Related Bolt comment: https://github.com/bolt/four/pull/418#issuecomment-490224719

bobdenotter:
Fixes #2013

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ a release.
 #### Fixed
 - Added missing string casting of `objectId` in `LogEntryRepository::revert()` method (#2009)
 
+### Translatable
+#### Fixed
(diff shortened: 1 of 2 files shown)


AkenRoberts:
Tests passed fine before I modified the changelog, then failed with a random Mongo error afterwards. Going to assume it is more environment-related and not from this PR.

Thank you @bobdenotter :clap:

Issue #1077, opened 2014-05-15
ReferenceIntegrity with mongodb relation.


evancauwenberg:
Goodmorning

I had a problem with references in mongodb, there isn't a vast relation between documents like in mysql so if i delete one document the relation still exists and i received a document not found exception. 

After some googling i came onto the extensions bundle and noticed it provided something like "ReferenceIntegrity" exactly what i needed i thought.
But i have encountered a problem with the ReferenceIntegrity annotation. 

I have the following Documents:
Day.php
List.php

``` php
List.php
    /**
     * @MongoDB\ReferenceMany(targetDocument="Pindabeer\Bundle\DayBundle\Document\Day", inversedBy="list")
     */
    protected $day;
```

``` php
Day.php
    /**
     * @MongoDB\ReferenceMany(targetDocument="Pindabeer\Bundle\ListBundle\Document\List", mappedBy="day")
     * @Gedmo\ReferenceIntegrity("nullify")
     */
    protected $list;
```

So for so good but now if i delete a "Day document" i still get the DocumentNotFoudException. If i switch the "mappedBy" and "inversedBy" it works but then if i persist my list, the day isn't references because it is nog the owner of the relation.

So my question is .. is it possible to do this actually?

Thnx in advance!

PR #2021, opened 2019-06-04
[ReferenceIntegrity] Get class from meta in ReferenceIntegrityListener

Jean-Beru:
Get class from meta to get real document class instead of its proxy. 

Fixes: #1077


AkenRoberts:
Thank you @Jean-Beru 👏

PR #2022, opened 2019-06-05
[Sluggable] Disable soft-deletable filter on searching for similar slugs

rodmarzavala:
Disable the soft-deletable filter to void SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry error while searching for similar slugs


rodmarzavala:
Thanks for your feedback @AkenRoberts, I've convert tabs into spaces.

Issue #2012, opened 2019-05-04
[Fixing Bug] Entity/Document configuration overwritten when using multiple annotations


mohamedaymenkarmous:
Hi,

While I'm using multiple Entity annotations (an example below), I find that Entity configuration `Gedmo\Mapping\MappedEventSubscriber::$configurations` gets different configurations independently and each configuration is overwritten by the next configuration that it will get (old configuration will not exist).

Example using ``Gedmo Loggable`` for the Document and ``Gedmo References`` for referencing ORM entities with ODM documents :

```
use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;
use Gedmo\Mapping\Annotation as Gedmo;

/**
 * @MongoDB\Document(collection="proofs")
 * @MongoDB\Document(repositoryClass="MyBundle\Document\ProofsRepository")
 * @Gedmo\Loggable(logEntryClass="MyBundle\Document\HistoryTransaction")
 */
class Proofs
{
  //...
    /**
     * @MongoDB\Id
     * @Gedmo\Versioned
     */
    private $id;

    /**
     * @Gedmo\ReferenceOne(type="entity", class="MyBundle\Entity\ProofElement", inversedBy="proofs", identifier="proof_id")
     */
    private $proof;

    /**
     * @MongoDB\Field(name="proof_id", type="integer")
     * @Gedmo\Versioned
     */
    private $proof_id;
  //...
}
```

I overrided Gedmo References Listener ``Gedmo\References\ReferencesListener`` in order to customize something and I found this problem: Inside ``Gedmo\References\ReferencesListener::postLoad()`` when I ``dump($config);`` exactly [after this line](https://github.com/l3pp4rd/DoctrineExtensions/blob/1df5277a7e6723bbd572259ed1ca23566ae63eda/lib/Gedmo/References/ReferencesListener.php#L41), I get the output below when I load `Proofs` data:

```
array:4 [
  "loggable" => true
  "logEntryClass" => "MyBundle\Document\HistoryTransaction"
  "versioned" => array:4 [
    0 => "id"
    1 => "proof_id"
  ]
  "useObjectClass" => "MyBundle\Document\Proofs"
]
```

But, when I remove Gedmo Loggable Annotations like this:

```
use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;
use Gedmo\Mapping\Annotation as Gedmo;

/**
 * @MongoDB\Document(collection="proofs")
[...]

PR #2025, opened 2019-06-07
[Bug fixed] Calling Listener's parent constructor for LoggableListener

mohamedaymenkarmous:
This PR if applied to all Listener's constructors, will fix this bug #2012.


AkenRoberts:
This particular PR doesn't add any functionality, since `LoggableListener` doesn't define its own constructor and shouldn't have the bug you noticed.

PR #2026, opened 2019-06-07
[Bug fixed] Calling Listener's parent constructor for ReferencesListener

mohamedaymenkarmous:
This PR if applied to all Listener's constructors, will fix this bug #2012.


AkenRoberts:
I manually made this same adjustment in the two classes who were missing parent constructor calls: `ReferencesListener` and `UploadableListener`.

In the future, a single PR would be just fine, and make sure to read over the [contributing](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/CONTRIBUTING.md) doc so you can include any relevant information and changes in the future. :)

PR #2029, opened 2019-07-08
[Translatable]: fix duplicate inherited properties

sandergo90:
Q | A
-- | --
Branch? | v2.4.x
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no

There currently is an issue with the gedmo translatable. Use case:

- Abstract class A (Product) with single table inheritance has a property $name, which is a doctrine embeddable and contains a property $value. This value property has the Gedmo\Translatable annotation.

- Class B (Pack) extends class A and has it's own properties.

When querying for class B gedmo will search for all translatable properties and will return the name.value property twice because it's first checking the parent class and later on it's own class. With this fix, when the embedded property is inherited, the property will not be added twice to the config. Otherwise this will give a duplicate join with same table alias.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ a release.
 #### Fixed
 - Remove hard-coded parent column name in repository prev/next sibling queries [#2020]
 
+### Translatable
+#### Fixed
+- Fix duplicate inherited properties [#2029]
+
 ## [2.4.37] - 2019-03-17
 ### Translatable
 #### Fixed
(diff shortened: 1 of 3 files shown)


sandergo90:
Can this PR be reviewed ? Thanks!

sandergo90:
@AkenRoberts Can you take a look at this ?

AkenRoberts:
The change looks reasonable enough to me. The test failures don't appear to be coming from this PR.

Can you please add a [changelog entry](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/CONTRIBUTING.md#changelog) for this? That'll help the PR and re-test it as well.

sandergo90:
@AkenRoberts done!

sandergo90:
@AkenRoberts Is it possible to merge this PR please?

AkenRoberts:
Apologies for the delay. Thank you @sandergo90.

PR #2040, opened 2019-08-23
Adding support for ODM 2.x

emulienfou:
This Pool Request add support for ODM 2.x who is provided by the package `doctrine/mongodb-odm`.
This PR doesn't drop support of ODM 1.x, it will check if classes exists.

By example, since in PHP7 the class `MongoRegex` doesn't exist anymore it will use the class `MongoDB\BSON\Regex` instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "doctrine/mongodb-odm": ">=1.0.2 <2.0",
         "doctrine/orm": ">=2.5.0",
         "phpunit/phpunit": "^4.8.35 || ^5.7 || ^6.5",
-        "symfony/yaml": "~2.6 || ~3.0 || ~4.0"
+        "symfony/yaml": "^2.6 || ^3.0 || ^4.0 || ^5.0"
     },
     "suggest": {
         "doctrine/mongodb-odm": "to use the extensions with the MongoDB ODM",
--- a/composer.json
+++ b/composer.json
@@ -39,12 +39,11 @@
         "doctrine/common": "~2.4"
     },
     "conflict": {
-        "doctrine/annotations": "<1.2",
-        "doctrine/mongodb-odm": ">=2.0"
+        "doctrine/annotations": "<1.2"
     },
     "require-dev": {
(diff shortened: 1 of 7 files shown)


markitosgv:
@emulienfou you need to update more extensions like Loggable LogEntryRepository.php that's extending DocumentRepository from Doctrine\ODM\MongoDB\DocumentRepository and now it needs to be extended from Doctrine\ODM\MongoDB\Repository\DocumentRepository

emulienfou:
Hi @markitosgv it's seems there is 3 classes who currently extends from `Doctrine\ODM\MongoDB\DocumentRepository`:
* `TranslationRepository`
* `LogEntryRepository`
* `AbstractTreeRepository`

The best solution to manage that without dropping support of ODM 1.x would be to create a `MiddleManDocumentRepository` class who will extends the one or the other class like:

```php
if (class_exists(\Doctrine\ODM\MongoDB\Repository\DocumentRepository::class)){
   class MiddleManDocumentRepository extends \Doctrine\ODM\MongoDB\Repository\DocumentRepository {}
} else {
   class MiddleManDocumentRepository extends \Doctrine\ODM\MongoDB\DocumentRepository {}
}
```

Do you think its the best solution and where do you think I can put this class?

markitosgv:
Hi @emulienfou I'm not a member of this project but I think your solutions is valid to support both versions.

Alfro:
I think it was mentioned in #2088 and #2089 that compatibility with Doctrine MongoDB ODM 2.0. is being worked on in master, so this probably won't be merged to 2.4.x

AkenRoberts:
☝️ True story. This PR was very useful for getting a head start on implementing ODM 2.0, though, so thank you and much appreciated.

I left it open mostly for visibility, so other people looking for ODM 2.0 compatibility could see an existing issue. It is a common complaint lately, so anything to reduce redundant issues helps. 😄

AkenRoberts:
ODM 2.0 support included in [v3.0](https://github.com/Atlantic18/DoctrineExtensions/releases/tag/v3.0.0).

Issue #1721, opened 2016-12-12
Is it possible to change private function to protected function in listener ?


raphaelriehl:
For example, i would like create my SluggableListener.
I do no want to rewrite everything, and i would like extend Gedmo\Sluggable\SluggableListener.
But, if i do that and i rewrite "makeUniqueSlug" function, this is impossible because this is a private function.
Is it possible to define "protected" instead of "private" in listeners of project for extend all listeners ?

Thanks in advance.

l3pp4rd:
it can be made protected, you can submit a PR. for other things, you can fork it and customize. extensions are pretty much locked for new features.

raphaelriehl:
Ok, my question is related to issue #1720. If my request is accepted i do not need protected function. If not i must custom project.

AkenRoberts:
Hello, and thank you for your contribution.

In the interest of spring cleaning and package modernization, I'm going through and closing old issues that have had no recent activity and/or are no longer relevant.

If this problem is still relevant to you, please open a new issue with the appropriate information and steps to reproduce.

Thank you!

PR #2048, opened 2019-09-24
[References] [Sluggable] [Sortable] [Translatable] [Tree] [Uploadable] Change private to protected in listener

syl20b:
Hello,
here is a fix to solve the issue #1721 
Hope this could help.


AkenRoberts:
Thank you for your contribution @syl20b. Right now I don't think this is the best direction to go. What we really need to strive for in the core is composition instead of inheritance. Adding proper layers and dependencies to help separate functionality (e.g. having a `SlugGenerator` interface dependency instead of a private function in the listener). Opening up the listeners to inheritance will make upgrading more difficult in the future as those classes are modified.

PR #2050, opened 2019-10-07
Added PHP 7 support for Doctrine ODM SoftDeletableFilter::addFilterCriteria().

Franky238:
Added PHP 7 support for Doctrine ODM SoftDeletableFilter::addFilterCriteria().
This fix is necessary for using this filter on Doctrine Mongo ODM with PHP 7+


AkenRoberts:
@Franky238 The `php7` branch ended up not being a focus, and instead we're focusing on making `master` PHP 7.1+ compatible. If you'd like to contribute, please open an issue or PR targeting `master`. Apologies for any confusion.

I'm curious, though, why the return type declaration is necessary to make this work? This doesn't fail in our tests on 7+.

Franky238:
@AkenRoberts if you are overriding method you need to make same notation like in parent.

Issue #807, opened 2013-08-16
Allow empty slug?


dirkluijk:
Hi,

Just a question. Is it possible to allow empty slugs for one specific case?

I am using fixtures to persist some 'page' entities to the database. I am using this slug for URL generation (using the `TreeSlugHandler`). But I want the first page (the homepage) to have an "empty" url path.

What I want:

home - `` `
some page -`some-page`
subpage -`some-page/subpage``

(This will be used in a `/{url}` catch-all route)

The entity "home" is now persisted with the slug `home`, how can I force an object to use an empty slug (by exception)?

Regards,

Dirk

PR #2053, opened 2019-10-30
Regenerate slug if it is empty

Guite:
This patch treats an empty slug like it was set to `null` so it gets regenerated automatically.

### DIFF
--- a/lib/Gedmo/Sluggable/SluggableListener.php
+++ b/lib/Gedmo/Sluggable/SluggableListener.php
@@ -289,8 +289,8 @@ class SluggableListener extends MappedEventSubscriber
             $oldSlug = isset($changeSet[$slugField]) ? $changeSet[$slugField][0] : $slug;
             $needToChangeSlug = false;
 
-            // if slug is null, regenerate it, or needs an update
-            if (null === $slug || $slug === '__id__' || !isset($changeSet[$slugField])) {
+            // if slug is null or empty, regenerate it, or needs an update
+            if (null === $slug || '' === $slug || $slug === '__id__' || !isset($changeSet[$slugField])) {
                 $slug = '';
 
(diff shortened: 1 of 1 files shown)


AkenRoberts:
Hi @Guite, thank you for the contribution. As indicated by the failing build, we currently only treat `NULL` as the "force regeneration" value.

Whether or not we should treat an empty string as such also, I'm not sure. I can see arguments going both directions. Maybe you can provide your context or reasoning?

This would also be considered a breaking change, so it would need to target `master` instead of 2.4. For these reasons I'm going to close this, but eager to here your perspective still.

Guite:
The context was a Symfony form type which includes an input field for the slug. This input field is optional so if it contains an empty value the slug should be generated automatically.

Originally I used this code for defining the form field:
```php
$builder->add('slug', TextType::class, [
            'label' => $this->__('Permalink') . ':',
            'required' => false,
            'empty_data' => '',
            'attr' => [
                'maxlength' => 255,
                'class' => 'validate-unique'
            ]
        ]
);
```

When creating a new entity using this form without entering a manual slug value the entity's slug was empty. I expected it to be generated automatically though. This worked as soon as I removed `'empty_data' => '',` so `null` is assigned as a fallback value instead of the empty string.

This PR was mainly intended to avoid similar problems (and efforts to search the reason) in future.

Guite:
Seems like it's not a bug, but a feature #807 -> 4ec518afc1aee9615188f91163be670d04a6d5dd

AkenRoberts:
Thank you for the context. I can definitely see why you tried the empty string in that situation.

I agree with the feature conclusion, though. It makes more sense to me personally that setting `NULL` means to let the extension handle it, and an empty string is more purposeful.

I'll try to keep this in mind when updating documentation or something along those lines. If you have any other suggestions for improvements in the future, please share them.

PR #2057, opened 2019-11-11
Add lazy blameable user support

ro0NL:
We could use this to compute a user entity from a JWT decoded user lazily.

The check for an existing user (database call) should be deferred as late as possible, to have a datebase-less authentication by default. This way the user is only created/loaded when needed (on write).

### DIFF
--- a/lib/Gedmo/Blameable/BlameableListener.php
+++ b/lib/Gedmo/Blameable/BlameableListener.php
@@ -19,6 +19,7 @@ use Gedmo\Blameable\Mapping\Event\BlameableAdapter;
 class BlameableListener extends AbstractTrackingListener
 {
     protected $user;
+    protected $lazyUser;
 
     /**
      * Get the user value to set on a blameable field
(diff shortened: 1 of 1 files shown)


AkenRoberts:
Thank you for the PR. This is a good idea, though I think it should be more properly implemented with a `UserProvider` interface layer, and then implementations can handle how readily they want to make the user to the listener.

That's one of the plans for Extensions 3.0. `v2.4.x` is basically feature frozen at this point. More information in the [contributing doc](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/CONTRIBUTING.md).

ro0NL:
i was hoping for a quick win :) all good then, i keep our `getFieldValue` override for a while.

AkenRoberts:
Totally get it. If this project were in a better place, I'd work with you to refine a bit and get it in. There are just way bigger fish to fry right now, when as you said, you can do a quick override to get that functionality.

Definitely a good idea still, though. So thank you. 😄

PR #2060, opened 2019-11-26
correction proposal on regex

TTMelki:
The Sluggable extension is inoperative because of the change made to the mongodb library. Using the MongoDB\BSON\Regex class it works fine on php7.3, mongodb 4.2 and symfony 4.4 environment


AkenRoberts:
Thank you. This is a backwards incompatible change, so it will need to target `master` instead of `v2.4.x`. We should also be explicit about what dependency versions we need, e.g. what the minimum version of `ext-mongodb` is, etc.

Please revise to target `master` and this will be a welcomed addition. 👍  (Also, please avoid the unneeded white space changes.)

PR #2061, opened 2019-11-28
[Tree] The value of path source property is cast to string type

kosarinin:
Hi! During the migration to php7.4, we faced with an error. We use integer value as a tree path source and it causes this error: 

> Notice: Trying to access array offset on value of type int

on this line https://github.com/Atlantic18/DoctrineExtensions/blob/81681364331b131518060e4776300a5346df1eb5/lib/Gedmo/Tree/Strategy/AbstractMaterializedPath.php#L299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Tree
+### Fixed
+- The value of path source property is cast to string type for Materialized Path Tree strategy (#2061)
 
 ## [2.4.38] - 2019-11-08
 ### Global / Shared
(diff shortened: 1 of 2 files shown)


kosarinin:
Hi @AkenRoberts! I'm sorry for bothering you but what do I need to do to review this MR? Looks like the CI failure doesn't relate to the MR.

michaelgracious:
@AkenRoberts Could this be looked at? it seems Travis fails due to a config error of Travis and not due to an error made by @kosarinin 
We need these changes to have this work on PHP 7.4

AkenRoberts:
Thank you. ❤️

PR #2067, opened 2019-12-09
Determine the correct value to set for the date field

oojacoboo:
The SoftDeleteable extension failed to determine the appropriate value to set to the date field.  This pull request uses the metadata to get the field type and sets the date value accordingly.


AkenRoberts:
Hi @oojacoboo, thank you for your first contribution!

I agree that we should better support other `datetime` types. I think it would be better to somehow tap into Doctrine's native types, though, and allow them to do as many of the implementations as possible. This would make Extensions more flexible to custom types as well.

Or, at the very least, have some sort of `Clock` interface that is injected and can be implemented as seen fit, with a sensible default provided by Extensions (like what you've added here).

That said, while this change is welcomed, it also would need to be in the `master` branch with appropriate tests and changelog entry. If you're willing to improve upon your idea there, I'd be grateful and pleased to review and assist you.

PR #2068, opened 2019-12-10
[Blameable] Add callable support for setUserValue

oojacoboo:
Okay, so a few things here.  

1.  The term "user" is too presumptuous and should be removed for a more neutral "value" term.
2.  The `getFieldValue` method should check if the value is a callable, and call it if so.

Regarding the callable - this allows for the "user value" to be set at a later point, or updated easily.  Using a callable means that it'll execute that function to get the value when it's needed and doesn't require that it be set during initialization only.

This solves multiple use cases and is a very simple addition.

### DIFF
--- a/lib/Gedmo/Blameable/BlameableListener.php
+++ b/lib/Gedmo/Blameable/BlameableListener.php
@@ -18,7 +18,7 @@ use Gedmo\Blameable\Mapping\Event\BlameableAdapter;
  */
 class BlameableListener extends AbstractTrackingListener
 {
-    protected $user;
+    protected $value;
 
     /**
      * Get the user value to set on a blameable field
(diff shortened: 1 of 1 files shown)


AkenRoberts:
I'm appreciating the PR activity and discussion! Thank you.

The concept of a callable/lazy user provider and its added flexibility I definitely agree with. This was discussed a bit in #2057. I still think having an interface for the provider injected is the best route, allowing for any implementation that helps developers (with a sensible default included in Extensions).

The term "user" being presumptuous I can somewhat agree with. It does create a connotation of a typical User or Person. I think "value" is far too ambiguous, though. I'd prefer something like "actor", as in "the thing performing an action on this record."

That said, these changes are considered breaking and aren't appropriate for 2.4.x. The `$user` property isn't private or marked as internal, so we'd risk breaking any extensions to the listener. The callable syntax also isn't supported in PHP 5.x (check the [CI builds](https://travis-ci.org/Atlantic18/DoctrineExtensions/builds/623317689)).

I'd be excited to review and assist an updated PR targeting `master` with the above in mind, if you're up for it!

oojacoboo:
@AkenRoberts closing this out.  I've submitted a new PR along the lines of your suggestions.

PR #2073, opened 2019-12-15
Added new Blameable on "remove" event listener support

oojacoboo:
Added support for a new on="remove" event listener for the Blameable extension.

Simply add `@Gedmo\Blameable(on="remove")` annotation to the property.


AkenRoberts:
Thank you, Jacob. I'm guessing this is useful when combined with `SoftDeleteable` (otherwise it's kind of pointless since the record gets deleted right after modification, haha).

In order to merge, you should also add along with this change:

* Relevant tests
* Updated documentation
* Changelog entry

If you have any questions about those items, don't hesitate to ask.

AkenRoberts:

PR #2074, opened 2019-12-15
Add deferred assignment support for the Blameable actor

oojacoboo:
An example Actor class that implements the new BlameableActorInterface

```php
<?php

class Actor implements BlameableActorInterface
{

    protected RoleStack $roleStack;


    /**
     * Constructor
     *
     * @param RoleStack $roleStack
     */
    public function __construct(RoleStack $roleStack)
    {
        $this->roleStack = $roleStack;
    }


    /**
     * Gets the actor/Role
     *
     * @return Role
     */
    public function getActor()
    {
        return $this->roleStack->top();
    }
}
```

### DIFF
--- a/doc/annotations.md
+++ b/doc/annotations.md
@@ -122,6 +122,15 @@ $treeListener = new Gedmo\Tree\TreeListener;
 $treeListener->setAnnotationReader($cachedAnnotationReader);
 $evm->addEventSubscriber($treeListener);
 
+// blameable
+$actor = new App\Actor(); // Extends Gedmo\Blameable\BlameableActorInterface
+$blameableListener = new Gedmo\Blameable\BlameableListener();
+$blameableListener->setAnnotationReader($cachedAnnotationReader);
+// Also can pass raw object or string to setActor, but using an Actor proxy object
+// allows for deferred actor assignment
+$blameableListener->setActor($actor);
+$eventManager->addEventSubscriber($blameableListener);
+
 // loggable, not used in example
 $loggableListener = new Gedmo\Loggable\LoggableListener;
 $loggableListener->setAnnotationReader($cachedAnnotationReader);
--- a/src/Blameable/BlameableActorInterface.php
+++ b/src/Blameable/BlameableActorInterface.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Gedmo\Blameable;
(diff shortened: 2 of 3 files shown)


AkenRoberts:
Hmm. This is different than what I was envisioning, but I think this is a better in-between step for 3.0. Since it won't require any modification to the Symfony bundle in order to keep using.

I was envisioning something more like this:

```php
<?php

/**
 * Blamable Actor, which can be used for an association as a whole
 * or get the string value (such as a username or email).
 */
interface Actor
{
    public function toActorString(): string;
}

/**
 * Provider for returning the current actor.
 */
interface ActorProvider
{
    public function getActor(): Actor;
}

class BlamableListener
{
    /**
     * @var ActorProvider
     */
    private $actorProvider;

    public function __construct(ActorProvider $actorProvider)
    {
        $this->actorProvider = $actorProvider;
    }

    // ...

    public function getFieldValue($meta, $field, $eventAdapter)
    {
        $actor = $this->actorProvider->getActor();

        if ($meta->hasAssociation($field)) {
            return $actor;
        }

        return $actor->toActorString();
    }
}
```

Doing this would mean breaking existing Symfony bundle installs, though, and I don't think 3.0 is going to be different enough to warrant that. Not at its current level of velocity, anyway.

I'm up for merging this, with the usual requirements:

* Relevant tests
* Updated docs
* Changelog entry

oojacoboo:
@AkenRoberts I don't understand why this extension would be making the string assumption.  We use actors as relationships and store the id in the database with FK constraints.  This is far better than a string IMO.  This is part of the reason the current design is limiting, because that line of thinking has polluted the API.  The interface that I created here doesn't make any assumptions about what type of value an actor is being represented as, nor should it IMO.

The annotation markdown was updated to include the Blameable details.  It wasn't even there previously.  Is there other documentation that needs to be updated?  I wasn't able to find anything.

I'll get a test and changelog added.

AkenRoberts:
I just based my example off of existing functionality, in order to demonstrate my thoughts on the architecture. I'm not opposed to added flexibility.

AkenRoberts:

oojacoboo:
@AkenRoberts can this get merged?  I don't understand why the actor cannot be mutable.

PR #2076, opened 2019-12-16
Don't check if metadata is already loaded for parent classes

ahilles107:
In case of expired cache for metadata (memcached/redis) - this blocked factory on reading parent clasess metadata.


ahilles107:
This 7.1 test fail is not connected to this change - i see that it fails also in last two merged (to 2.4) pull requests.

ahilles107:
@AkenRoberts - can You take a look?

AkenRoberts:
Seems like a reasonable request. However, it will need a new PR with:

* Target `master` branch
* A changelog entry
* A test to confirm parent entity metadata is loaded as-expected

Ping if I can help answer any questions!

PR #2084, opened 2020-01-15
Do we support latest php versions?

svycka:


svycka:
That's a problem we do not support any supported php versions

svycka:
@AkenRoberts maybe you have eta  for 3.0? or should we try to fix this(at least make tests pass)?

svycka:
I agree we should drop support for unsupported PHP versions and release 2.5 before 3.0

andrewtch:
@svycka  - actually I did a bit of digging and most errors come from phpunit (which should be upgraded to 8+), not from library itself. I'll try to compile 7.2 branch a bit later.

svycka:
@andrewtch the problem is that we support old PHP versions and can't update PHPUnit

andrewtch:
For 2.4, yes. For 2.5 - 3.0 don't see a reason why to continue support EOL versions.

AkenRoberts:
We'll want to do this on `master`, which is meant to be a new major release (3.0). It's already bumped to a 7.1 minimum, but we should definitely add 7.4 to Travis and see if there's any new issues to address.

No 2.6 release is planned, and even if it was I wouldn't modify minimum PHP versions - that's what major releases are for.

Anything you'd like to contribute to help with this, feel free to PR on `master`! 😄

svycka:
@AkenRoberts PHP version bump is not a major change. Also when we can expect 3.0 release because we don't have any supported version right now. We may be able to help you, but it's hard to track down what needs to be done for 3.0 what's holding you from releasing it?

AkenRoberts:
Changing a package's minimum PHP requirement is definitely a breaking/major change, and will be reserved for 3.0. It's already done on `master`, along with the upgrade to PHPUnit 8.

As for when, I'm hoping in the next couple of weeks. Right now I'm working on MongoDB ODM 2.0 support. Next, double check any other dependencies for changes in supported versions (the major goal with 3.0 is to not use any unmaintained dependency versions). I'd also like to [finish setting up CS Fixer](https://github.com/Atlantic18/DoctrineExtensions/pull/1939). I'll also give the issues and PR backlog a once over, to see if there is anything that might be crucial to include. In general, though, 3.0 is meant to change the library's functionality as little as possible, and focus more on infrastructure and dependencies.

svycka:
@AkenRoberts I don't understand why it is major change are you saying that it is only for this repository? for example, doctrine bumped min PHP version from [5.3.2](https://github.com/doctrine/dbal/blob/2.1.x/composer.json#L15) to [7.2](https://github.com/doctrine/dbal/blob/2.10.x/composer.json#L35) in 2.x and many other libraries do this. I am not saying we should release 2.5 but just curious what makes you think that's a major change? As I understand composer will not update this package if the user is on an older version but it won't break anything. And if this is major change why bumping other dependencies is not a BC break?.

Also, I see a bigger problem we require `"doctrine/common": "~2.4"` but testing against `"doctrine/common": ">=2.5.0"`. Why not just bump the required package instead of adding it as a dev?

AkenRoberts:
My assumption is that bumping the PHP requirement in a minor or patch version would cause anyone on a lower PHP version to experience an error, if their `composer.json` version constraint matches the update (e.g. `^2.4`). It is an assumption, though, so I'll play around with it later and see for sure.

I never said bumping other dependencies is not a BC break. It can be, just depends on what is required to install those dependencies and if they affect your package's functionality/code.

The `doctrine/common` thing is definitely wacky. That was fixed in `master`, too. All of this dependency stuff is fixed there already. Are there other concerns?

svycka:
@AkenRoberts don't see at the moment. just that we do not support any supported PHP version. I think if someone uses PHP 5.3 then they most likely do not update and this extension. If they run un PHP 5.3 I think they have more problems than this library.

> My assumption is that bumping the PHP requirement in a minor or patch version would cause anyone on a lower PHP version to experience an error if their composer.json version constraint matches the update (e.g. ^2.4).

if required PHP version does not match the required version will not update the package that's the same as with other dependencies.

anyway I am waiting for 3.0 ping me when alpha or beta will be ready to test

AkenRoberts:
Doctrine Extensions works fine on all currently-supported versions of PHP. I realize they weren't explicitly tested, and maybe that's obvious to me and not others.

I added PHP 7.3 & 7.4 to the `v2.4.x` build, [with all green lights](https://travis-ci.org/Atlantic18/DoctrineExtensions/builds/640624419). I did have to suppress deprecations, since they're coming from PHPUnit and I have no intentions of upgrading it on `v2.4.x`.

Thank you for the concern and confirming with me. 👍

PR #2085, opened 2020-01-18
Add PHP 7.4 Compatibility to 3.0

AkenRoberts:
Starting off PHP 7.4 compatibility by adding it to Travis and seeing what the build result is.

If there are new errors or deprecations to be concerned about, we'll add focus on adding them to the `php7.4` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ php:
     - 7.1
     - 7.2
     - 7.3
+    - 7.4
 
 env: MONGODB_SERVER=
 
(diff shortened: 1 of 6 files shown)


AkenRoberts:
The only PHP 7.4 issue was a deprecated `ReflectionType::__toString()` method being used in PHPUnit 6.x. Upgrading to PHPUnit 7.x and making slight adjustments to support it resolved all obvious PHP 7.4 issues (no others shown running tests).

We will want to move to PHPUnit 8 still, to take advantage of modern & supported tools. Out of the scope of this PR, though.

PR #2096, opened 2020-02-13
[Sluggable][Tree] Fix MongoRegex to Regex for ODM 2.0 compat

Alfro:
* Fixes
`Error: Class 'MongoRegex' not found`

* Also removed the regex escape character '/' as `new Regex()` expects it not to include those.


Alfro:
Hi! I did this against the feature/mongodb-odm-2.0 branch instead of master thinking maybe work would continue in that branch. Feel free to move it to master or Close if the change has already been done in local.

AkenRoberts:
Thank you.

PR #2098, opened 2020-02-27

Evertt:
I needed this, because there were discrepancies between the names of the files that were stored in my public folder and the filenames that were saved in my database.


Evertt:
Sorry I see the errors in the tests. I'll fix those first.

Evertt:
Okay, I fixed the bugs that I had created. There are still tests failing, but those were already failing. 😅

AkenRoberts:
Thank you for the contribution! A couple things before we can get this to a merge state:

1. Can you further explain what was wrong here that you're fixing? The filename saved to the DB was different than the generated filename? Any way you can write a test to assert the desired behavior failing before and fixed with this PR?
2. When completed, this will need to be directed at `master` instead of the `2.4.x` branch. You may want to either start over or rebase your changes there. 😄

Evertt:
Okay, I messed up the rebasing. 😅

Issue #2087, opened 2020-01-23
[Tree] Problem nestedset with doctrine second-level-cache


Nastro:
There is a problem with the nestedset tree working with doctrine second-level-cache.

I create the following nestedset tree via sonata-admin with doctrine second-level-cache enabled:

```
\root
    \branch1
        \leaf11
        \leaf12
    \branch2
        \leaf21
        \leaf22
```
Сreation order is: root, branch1, branch2, leaf11, leaf12, leaf21, leaf22.

As a result, I have such data in database:

```
 id  | root_id | parent_id |  title  | lft | rgt | lvl |  path   |        slug         
-----+---------+-----------+---------+-----+-----+-----+---------+---------------------
 296 |     296 |           | root    |   1 |  14 |   0 | root    | root
 298 |     296 |       296 | branch2 |   2 |  11 |   1 | branch2 | root/branch2
 297 |     296 |       296 | branch1 |  12 |  13 |   1 | branch1 | root/branch1
 299 |     296 |       297 | leaf11  |   9 |  10 |   2 | leaf11  | root/branch1/leaf11
 300 |     296 |       297 | leaf12  |   7 |   8 |   2 | leaf12  | root/branch1/leaf12
 301 |     296 |       298 | leaf21  |   5 |   6 |   2 | leaf21  | root/branch2/leaf21
 302 |     296 |       298 | leaf22  |   3 |   4 |   2 | leaf22  | root/branch2/leaf22
```

It can be seen that branch1 fields `lft` and `rgt` are set to 12 and 13 respectively, that is incorrect because branch1 has 2 children, so its `lft` and `rgt` should differ by no less than 4.

When doctrine second-level-cache is disabled, all the `lfts` and `rgts` are assigned correctly.

It seems that setting `Query::HINT_CACHE_EVICT` hint to `true` for all the update queries in https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Tree/Strategy/ORM/Nested.php fixes the issue.

Is there any other solution that does not require fixing DoctrineExtensions code? Or shall I set new pull-request?

My doctrine config:

```
doctrine:
    orm:
        auto_generate_proxy_classes: true
        metadata_cache_driver: ~
        query_cache_driver:
            type: pool
            pool: doctrine.local_cache_pool
        result_cache_driver:
            type: pool
            pool: doctrine.local_cache_pool
        second_level_cache:
            enabled: true
[...]

PR #2099, opened 2020-02-28
[TREE]: Second Level Cache invalidatation + verify and recover for single node

stipic:
Gedmo TREE dont work with second level cache see -> #2087 

And verify and recover method work with ALL nodes in repository. I have over 250 root nodes and verify and recover is very slow for that much nodes so i made that two methods accept $node param so i can acually verify($node) and recover($node) and not ALL nodes in DB.

Please check and say what you think, not supporting second level cache is a big problem for me.

Thanks


AkenRoberts:
Hello and thank you for the first contribution!

I'm okay with potentially merging the second-level cache query hints. The other verify and recover performance issues should be issued and PR'ed separately, as that is a different issue. Also can't merge all of the other changes that appear more for your dev environment or project.

If you have any questions about that or if I can clarify anything, please ask!

Issue #1978, opened 2019-01-02
Columns root_id, lft, lvl, rgt not updating if using UUID in BINARY type


dsmatana:
Hello

I am using [ramsey/uuid-doctrine](https://github.com/ramsey/uuid-doctrine) as ID generator with binary type column in Laravel 5.4, but after every insertion of node crucial values are NULL or zero.

After running SQLs in logs, UPDATEs affected 0 rows because of UUID as string in WHERE. Correct WHERE is id = UNHEX("103ea5ca9f9d4a2586d2978a857d0553").

I think that SQL logs have substituted values for readability but I don't know how to figure out real SQL sent to DB.

Can you please suggest workaround for this situation ?

Thank you

Composer requires:
```
"require": {
    "php": ">=5.6.4",
    "bshaffer/oauth2-server-php": "^1.10",
    "gedmo/doctrine-extensions": "^2.4",
    "guzzlehttp/guzzle": "^6.3",
    "laravel-doctrine/extensions": "^1.0",
    "laravel-doctrine/orm": "1.3.*",
    "laravel/framework": "5.4.*",
    "laravel/tinker": "~1.0",
    "lcobucci/jwt": "^3.2",
    "m7/oauth2-client": "^0.5.5",
    "ramsey/uuid-doctrine": "^1.5"
},
```
Table users_groups:
![doc1](https://user-images.githubusercontent.com/4609432/50594608-8e34f380-0e9d-11e9-9659-f02ef305310f.PNG)

State after inserting nodes:
![doc2](https://user-images.githubusercontent.com/4609432/50594616-90974d80-0e9d-11e9-82e8-9110b5e8a747.PNG)

Logs from laravel:
(Ran Flush after every entity persist)
```
[2019-01-02 14:52:44] local.DEBUG: "START TRANSACTION" [0.00032186508178711] 
[2019-01-02 14:52:44] local.DEBUG: INSERT INTO users_groups (id, name, lft, lvl, rgt, hidden, root_id, parent_id) VALUES ("103ea5ca-9f9d-4a25-86d2-978a857d0553", "Root", "0", "0", "0", "", "", "") [0.0010349750518799] 
[...]

PR #2103, opened 2020-03-11
Support Custom ID Types

pdugas:
Add support for IDs in nested trees that use a custom Doctrine Type where the default use of __toString() when building UPDATE queries is incorrect. This allows use of the binary UUID types like those provided by  https://github.com/ramsey/uuid-doctrine.


pdugas:
Intended to resolve #1978

pdugas:
Standby...  Corresponding changes in the repository class are needed too.

pdugas:
Works for me now.

pdugas:
I'm cancelling this effort.

PR #2109, opened 2020-03-30

Evertt:
I now made the problem clear with a modified test in the first commit.

Before my fix the test results in:

    Failed asserting that two strings are equal.
    Expected: '123-2.text'
    Actual: '123-2.txt'

### DIFF
--- a/src/Uploadable/UploadableListener.php
+++ b/src/Uploadable/UploadableListener.php
@@ -449,11 +449,12 @@ class UploadableListener extends MappedEventSubscriber
     /**
      * Moves the file to the specified path
      *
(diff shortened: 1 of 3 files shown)


Evertt:
Why did this got closed and not merged? This was a real issue.

AkenRoberts:
Sorry @Evertt! The `master` branch was renamed/removed, and GitHub auto-closed this because the target branch doesn't exist now.

I encourage you to reopen targeting `main`. I am going to stick a label on it regardless so I know to return to it.

Evertt:
Ah okay, I opened #2170 now for the main branch.

PR #2111, opened 2020-04-09
[SoftDeleteable] Fix filter not invalidating Query cache

alanpoulain:
When the query cache is enabled for the ORM entity manager, enabling or disabling the soft deleteable filter for an entity doesn't change the hash of the filter. Therefore, the result cannot change.

The only way to invalidate the query cache is to set a varying (depending of the entity enabled or disabled) boolean parameter (the hash of a filter is calculated by serializing the parameters: https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Query/Filter/SQLFilter.php#L110-L113).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Doctrine Behavioral Extensions
 
-[![Build Status](https://secure.travis-ci.org/Atlantic18/DoctrineExtensions.png)](http://travis-ci.org/Atlantic18/DoctrineExtensions)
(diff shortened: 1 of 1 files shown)


alanpoulain:
I'm not sure why the first commit hasn't failed, since it's the case for my branch (https://travis-ci.com/github/alanpoulain/DoctrineExtensions/jobs/317632440). It fails also in local.

alanpoulain:
Replaced by https://github.com/Atlantic18/DoctrineExtensions/pull/2112 (to show the failing commit).

PR #2112, opened 2020-04-09
[SoftDeleteable] Fix filter not invalidating Query cache

alanpoulain:
When the query cache is enabled for the ORM entity manager, enabling or disabling the soft deleteable filter for an entity doesn't change the hash of the filter. Therefore, the result cannot change.

The only way to invalidate the query cache is to set a varying (depending of the entity enabled or disabled) boolean parameter (the hash of a filter is calculated by serializing the parameters: https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Query/Filter/SQLFilter.php#L110-L113).

The first commit is a failing test to show the issue, the second one fixes it.


AkenRoberts:
Great work, Alan. Thank you very much. 👏

AkenRoberts:
Tagged `v2.4.40`

PR #2114, opened 2020-04-21
[SoftDeleteable] POC: Add ability to change default SoftDeleteable behavior

zsapozhnikov:
Here is the case: if we want to soft delete some entity, but it should be visible in the related entities. We can disable filter in ALL those places and only when show list of this entities the disabled will be hidden. But this POC add ability to reverse this behavior when there are too many places to refactor. For example we have `User` entity. `User` entity is also used all across the system and we don't want to change the behavior and hide it if the some user will be soft deleted. We want him to be shown in logs, orders, etc. But we don't want to show him in the list of users and we don't want to modify the query for that by the hands. We can add `disabled=true` to the annotation and it still be shown in all relations (though it is soft deleteable). And in the list of users we'll use 
```php
$filter = $em->getFilters()->getFilter('soft_deleteable');
$filter->enableForEntity(User::class);
```
to hide all soft deleted entities

### DIFF
--- a/doc/softdeleteable.md
+++ b/doc/softdeleteable.md
@@ -12,6 +12,7 @@ Features:
 - Annotation, Yaml and Xml mapping support for extensions
 - Support for 'timeAware' option: When creating an entity set a date of deletion in the future and never worry about cleaning up at expiration time.
 - Support for 'hardDelete' option: When deleting a second time it allows to disable hard delete.
+- Support for 'disabled' option: When disabled is defined the filter uses it as the fallback behavior if `enableForEntity` or `disableForEntity` are not applied
 
 Content:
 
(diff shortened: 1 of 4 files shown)


zsapozhnikov:
@AkenRoberts do you have any questions maybe? I'll be glad to see your thought about that)

AkenRoberts:
I think this is a prime example of how the term and usage of "soft delete" gets blurry, and is used in a way it wasn't intended.

You should think of soft delete as **deleted**. The record isn't there anymore, regardless of where it used to live, what relationships it has, etc. Obviously it does still exist in the DB, but to the ORM and application, the record is supposed to be non-existent.

Your use case indicates that you can't delete the records, because they're still crucial to your application and relationship integrity. Perhaps a different _state_ is more appropriate, such as a User being "archived" or "inactive".

AkenRoberts:
Hi! This was auto-closed because I renamed/removed `master`. However, I will repeat my original sentiment in that I don't believe this sort of logic should live in the Soft Delete feature.

I encourage you to open a new issue and/or PR if this continues to be an issue for you. 😄

PR #2116, opened 2020-04-28
[Tree] [NestedSet] fixes and enhancements

pmishev:
- Added and fixed some wrong comments
- Verify and recover wrong levels in nested set
- childrenQueryBuilder() to allow specifying sort order separately for each field
- Added option to reorder only direct children in reorder() method
- Added 'includeNode' option in getPathQueryBuilder() to specify whether you want the starting node included or not
- Added 'base' property for tree level annotation
- Added 'treeRootNode' options in verify() in case you want to verify a single tree in a forest
- Added options to recover() for sibling order, tree root in a forest, verification skip and auto-flushing
- Added recoverFast() method for where speed is more important than safety and entity manager state
- Added getPathAsString() method to entity repository

### DIFF
--- a/doc/tree.md
+++ b/doc/tree.md
@@ -24,10 +24,10 @@ Thanks for contributions to:
 
 Update **2018-02-26**
 
-- Nodes with no Parent can now be sorted based on a tree root id being an id from another table. Existing behaviour 
-  is unchanged unless you add properties to the `@TreeRoot` annotation. Example: You have two categories with no parent, 
-  horror and comedy, which are actually categories of 'Movie', which is in another table. Usually calling `moveUp()` or 
-  `moveDown()` would be impossible, but now you can add `@TreeRoot(identifierMethod="getRoot")`, where `getRoot` is the 
+- Nodes with no Parent can now be sorted based on a tree root id being an id from another table. Existing behaviour
+  is unchanged unless you add properties to the `@TreeRoot` annotation. Example: You have two categories with no parent,
+  horror and comedy, which are actually categories of 'Movie', which is in another table. Usually calling `moveUp()` or
+  `moveDown()` would be impossible, but now you can add `@TreeRoot(identifierMethod="getRoot")`, where `getRoot` is the
   name of your class method returning the root id/entity.
 
 
@@ -1334,7 +1334,7 @@ There are repository methods that are available for you in all the strategies:
       * nodeDecorator: Closure (null) - uses $node as argument and returns decorated item as string
       * rootOpen: string || Closure ('\<ul\>') - branch start, closure will be given $children as a parameter
       * rootClose: string ('\</ul\>') - branch close
-      * childStart: string || Closure ('\<li\>') - start of node, closure will be given $node as a parameter
+      * childOpen: string || Closure ('\<li\>') - start of node, closure will be given $node as a parameter
       * childClose: string ('\</li\>') - close of node
       * childSort: array || keys allowed: field: field to sort on, dir: direction. 'asc' or 'desc'
   - *includeNode*: Using "true", this argument allows you to include in the result the node you passed as the first argument. Defaults to "false".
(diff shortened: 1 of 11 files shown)


AkenRoberts:
@pmishev Really grateful for the work you've put in here. It'll take me a bit to digest everything.

I'm concerned about changing the method APIs by adding new parameters. One intention with a 3.0 release was to limit the number of breaking changes. And on a more personal opinion, I dislike the use of an associative `$options` array that changes based on the method you're calling. It's repetitive, hard to document, and requires extra code if you want type safety / validation on the values.

What do you think of splitting up this PR into non-breaking improvements (including anything you'd consider a bug fix), and then the API-changing stuff? Should make it easier to digest and get some things merged sooner, instead of as a whole.

pmishev:
@AkenRoberts Thanks for having a look at it. All of these are changes I've been maintaining in my fork for a few years now, as we use them for our project. This time it took me a while as a hard push into master has overwritten the commit tree, so I had to reapply them line-by-line.
Seeing that a new major release is planned, I took the opportunity to divide everythin in as many separate commits as possible, so they can be individually reviewed.

These are the fixes:
https://github.com/Atlantic18/DoctrineExtensions/pull/2116/commits/7eb2fea9297ba28e6aac2b560d98d894c8b6dad1
https://github.com/Atlantic18/DoctrineExtensions/pull/2116/commits/dc355e10082eef6acffa1ed539844edcc62955df?w=1

This I consider a fix as well, or rather finishing the currently half-baked implementation:
https://github.com/Atlantic18/DoctrineExtensions/pull/2116/commits/b296c582dd79e69b6eb4f3254e9176c15b49cd43

This one is adding an additional param to the reorder() method - it's not a breaking change:
https://github.com/Atlantic18/DoctrineExtensions/pull/2116/commits/3184f6a6ddb5fad52c256b832a354a28070996d6

And then the big one:
https://github.com/Atlantic18/DoctrineExtensions/pull/2116/commits/7ae9cb160b76a75ee0a4a2425016d1fd8761ba2b

I couldn't really separate it as all of the changes are somehow connected.
There are no breaking changes here either, everything is backward-compatible.

About the associative `$options` array. I get where you're coming from, but in my experience, I've found that this approach has its benefits:
If your method has a dozen of parameters and you only want to specify the last one, you have no choice but filling in 11 other default values, trying hard not to mess up the order or the default values, just so you can pass the one you need. This way, it's much cleaner to specify only the parameters you actually need to change.

If you decide to remove an option in a newer version of the API, it gives you more flexibility.

It does require extra code for type safety/validation and that's why I normally use this approach with Symfony's OptionsResolver component, where this can be done with minimal code. I just didn't want to introduce another dependency in the project in this case, but I'd be happy to do so, if it would be an acceptable solution.

I don't really understand the point of being harder to document - I have documented the options in their respective docblocks.

In the end, whatever you decide to accept would save me time next time I have to upgrade, so I'll be happy.

AkenRoberts:
> a hard push into master has overwritten the commit tree

Yeah, sorry about that. The old `master` was an in-progress but very messy attempt at a new major release, and it was just all over the place while no longer being maintained. I made the executive decision to ice it, and focus on a more reduced and non-breaking 3.0 while the overall state and direction could be reevaluated.

> I took the opportunity to divide everythin in as many separate commits as possible, so they can be individually reviewed

Awesome, that'll help me review more individually. Thank you!

> There are no breaking changes here either, everything is backward-compatible.

True, I suppose an optional parameter is only breaking if someone has overloaded the method with their own parameter. I do want to be careful about adding parameters, though, because once they're in the public API we need to maintain them and introduce future BC concerns.

And I appreciate the additional thoughts for the array options. Regarding documentation, I'm thinking more about outside of docblocks, and then just the intuitiveness of these methods in general. There may be an opportunity to create a better API instead of jamming options into existing methods. Just something I want to fully evaluate before committing to it.

Thank you again, looking forward to pouring into this more closely.

AkenRoberts:
Hi! This was auto-closed because I renamed/removed `master`. I've marked this as Still Relevant so I know there is some validity to this addition still.

PR #2117, opened 2020-04-29
SortableListener Remove unused $uow variable

symfonyaml:
Just removed unsued `$uow` variable in `SortableListener.php`


AkenRoberts:
🥂

PR #2119, opened 2020-05-03
Remove duplicate root definition from XML mapping

maxbaldanza:
This links to the following commit: https://github.com/Atlantic18/DoctrineExtensions/pull/1777
Where `root` was removed from the yaml example but not from the XML one.


AkenRoberts:
Cheers!

PR #2120, opened 2020-05-06
[Sluggable] Remove interpreted annotations from phpdoc

flohw:
I had an issue with these samples as they are interpreted by Donctrine Annotation Reader.
I use the library (2.4.39) with antishov/doctrine-extensions-bundle in a Symfony 4.4 environment which require to load all annotation mapping by configuration even if they are not used.
I get the following error when running the schema drop command.

```
In AnnotationException.php line 54:

  [Semantical Error] Annotation @Gedmo\Slug is not allowed to be declared on class Gedmo\Mapping\Annotation\SlugHandler. You may only use this annotation on these code elements: PROPERTY.
```

I choose to remove the examples here are they are also provided in the documentation section.

Hopefully a new release will be published soon with this fix as my project will run in production shortly :slightly_smiling_face: 
Also, note that the 2.4.40 release on github is not present in packagist.

Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Sluggable
+#### Fixed
+- Remove PHPDoc samples as they are interpreted by Annotation Reader.
 
 ## [2.4.40] - 2020-04-27
 ### SoftDeleteable
--- a/lib/Gedmo/Mapping/Annotation/SlugHandler.php
+++ b/lib/Gedmo/Mapping/Annotation/SlugHandler.php
@@ -7,18 +7,6 @@ use Doctrine\Common\Annotations\Annotation;
 /**
  * SlugHandler annotation for Sluggable behavioral extension
  *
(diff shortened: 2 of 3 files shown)


AkenRoberts:
Cheers 🥂

I do see [2.4.40 on Packagist](https://packagist.org/packages/gedmo/doctrine-extensions#v2.4.40), so maybe it was just a hiccup somewhere. Good looking out, though!

AkenRoberts:
[v2.4.41 tagged.](https://github.com/Atlantic18/DoctrineExtensions/releases)

jordisala1991:
The latest tagged: 2.4.41 is not on packagist. Seems like an issue on the auto-update from packagist. Maybe you need to check the configuration of that.

AkenRoberts:
@jordisala1991 Thank you, you're right this package isn't set to auto-update on Packagist. I'll look into that.

flohw:
Thank you @AkenRoberts. Works fine. :ok_hand: 

Sorry for the delay, vacation week. :slightly_smiling_face:

PR #2123, opened 2020-05-17
added soft deletable filter configuration

nuryagdym:


doc/symfony4.md
```diff
@@ -122,6 +122,26 @@ orm:
             dir: "%kernel.root_dir%/../vendor/gedmo/doctrine-extensions/lib/Gedmo/Tree/Entity"
 ```
 
+<a name="ext-filtering"></a>
+## Filters
+
+The **softdeleteable** ORM filter also needs to be configured, so that soft deleted records are filtered when querying.
+To do so, add this filter info to your **doctrine.orm** configuration, edit **app/config/config.yml**:
```
AkenRoberts, line 129: I believe all Symfony 4+ installs have the `packages` config setup now. So this would be `/config/packages/doctrine.yaml`.

nuryagdym: you are right, those pathes need to be changed, I did not want to make too much, changes.
If you want I cant update those pathes.

AkenRoberts: Yes please!

AkenRoberts:
You could have just updated this PR, FYI. :)

PR #2125, opened 2020-05-17
added soft deletable filter configuration and updated codes and paths to fit Symfony 4

nuryagdym:


AkenRoberts:
Dope, thank you.

PR #2126, opened 2020-05-20
Fix verify left & right checks as int

symfonyaml:
## Problem
There is an issue when we verify a tree with multiple roots.
`node [2] has invalid left or right values`

## Context
PHP 7.2

Build a tree just with one root element in the table:
| id     	| root    	| level 	| left 	| right 	| parent_id |
|--------	|---------	|-------	|------	|-------	|-------	  |
| 1      	| 1       	| 0     	| -1   	| 0     	| NULL       |

```php
$repo = $this->doctrine->getRepository('My\Entity\Tree');
$repo->verify();
```
At [this line](https://github.com/Atlantic18/DoctrineExtensions/blob/v2.4.x/lib/Gedmo/Tree/Entity/Repository/NestedTreeRepository.php#L1007) , the check does not pass because `left = -1` and `right = 0`
```php
// NestedTreeRepository.php::verifyTree()
...
if (!$right || !$left) {
```

## Remarks
This issue looks to me so general, so maybe I have done something wrong, if you know please comment.


symfonyaml:
I was wrong my bad :)

PR #2127, opened 2020-05-25
Allow composite identifiers in loggable

MisatoTremor:
Composite identifiers with foreign entities availaible for ORM > 2.6 due to earlier version not handling identifiers for newly inserted foreign objects


MisatoTremor:
Closed in favor of #2128

PR #2129, opened 2020-05-29
[TREE][Materialized] Fixed path not correctly insert entity ID when TreeSource is type int

iceroads:
Fixed a bug where the entity ID was not correctly inserted in path if the TreePathSource is of type integer and appendId is true.
See example:
```php
    /**
     * @Gedmo\TreePath(appendId=true)
     * @ORM\Column(name="path", type="string", length=3000, nullable=true)
     */
    private $path;

    /**
     * @Gedmo\TreePathSource
     * @ORM\Column(name="integer_sort", type="integer")
     */
    private $integerSort;
```


AkenRoberts:

PR #1895, opened 2018-03-21
[bugfix] Add support for "date*_immutable" types

phansys:
|Q            |A     |
|---          |---   |
|Branch       |v2.4.x|
|Bug fix?     |yes   |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

**TODO:**
- [x] Add tests;
- [x] ~~Update docs~~.

PR #2135, opened 2020-06-16
Add support for datetime_immutable fields to SoftDeleteable

jorrit:


jorrit:
@AkenRoberts Would this PR be acceptable on the 2.4 branch? Or should I target the master branch instead?

phansys:
See #1895.

AkenRoberts:
> Would this PR be acceptable on the 2.4 branch?

Unfortunately no, it would need to target `main`.

Something like #2248 is probably a better solution, though, as it is more flexible with the mapping type and uses DBAL as the factory for the end value. Perhaps you could test that PR and contribute any improvements.

PR #2144, opened 2020-07-02
Added missing index 'general_translations_lookup_idx'

cbclicer:
Created a new index 'general_translations_lookup_idx' and renamed the previous index 'translations_lookup_idx' to 'locale_translations_lookup_idx' so that the query perfomance for translations is significantly improved. No index was used in this select query before.

Reason:
The select query in method \Gedmo\Translatable\Entity\Repository\TranslationRepository::findTranslations() only uses the two columns 'foreignKey' and 'objectClass' but not 'locale' column in the WHERE condition. 
This leads to the fact that the queries all happen without indexes, because the index 'translations_lookup_idx' does not work. Especially in MySQL 8.0 this leads to a significant slowdown and overuse of the system.

### DIFF
--- a/lib/Gedmo/Translatable/Entity/Translation.php
+++ b/lib/Gedmo/Translatable/Entity/Translation.php
@@ -13,8 +13,11 @@ use Doctrine\ORM\Mapping\Entity;
  * @Table(
  *         name="ext_translations",
  *         options={"row_format":"DYNAMIC"},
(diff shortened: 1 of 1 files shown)


franmomu:
Thanks @cbclicer! closeing in favor of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2399 which is targeting `3.x`.

PR #2145, opened 2020-07-02
Fix Warning: count(): Parameter must be an array or an object that im…

gvanbeck:
Fix Warning: count(): Parameter must be an array or an object that implements Countable


phansys:
Related to #1967.

PR #2147, opened 2020-07-04
fix docs url

kevinpapst:
fixed url, pointing to wrong repository


AkenRoberts:
Thank you for pointing this out. 😄 I changed it to point to the correct branch.

Issue #2130, opened 2020-05-29
Support doctrine/common 3


jaikdean:
The [release notes](https://github.com/doctrine/common/releases/tag/3.0.0) for `doctrine/common` 3.0.0 suggest it's easy to support both 2.x and 3.0.

sixpeteunder:
Has any work been done on this? Or is there a form in which doctrine/common 3.0 is supported?

jaikdean:
@sixpeteunder See #2150

sixpeteunder:
Thanks @jaikdean. I left a comment on there.

svycka:
we have to drop php <=7.1 versions if we want `doctrine/common:3.0.0`

AkenRoberts:
Thank you for bringing this up and doing some initial recon. 👍 

There are no plans to do any of this on the 2.4.x branch. `master` is targeting a 3.0 release and has already done some initial work here, such as updating some dependencies, bumping the minimum PHP version to 7.2, etc.

I haven't added support for doctrine/common 3.0 yet, though. I'll take care of that now. :)

AkenRoberts:
⬆️ Completed in #2150.

PR #2150, opened 2020-07-09
Add doctrine/commons 3.0 support

svycka:
let us see what will happen with tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "doctrine/mongodb-odm": ">=2.0"
     },
     "require-dev": {
-        "doctrine/common": "^2.13 || ^3.0",
+        "doctrine/common": "^2.7 || ^3.0",
         "doctrine/annotations": "^1.0",
         "doctrine/cache": "^1.0",
         "doctrine/event-manager": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,11 @@
         "doctrine/mongodb-odm": ">=2.0"
     },
     "require-dev": {
-        "doctrine/common": ">=2.5.0",
+        "doctrine/common": "^2.13 || ^3.0",
+        "doctrine/annotations": "^1.0",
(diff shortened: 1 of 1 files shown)


svycka:
okay, I guess it's not possible to support PHP 5.x and doctrine/commons 3.0. Just wondering do we have users who need support for PHP 5.x versions? Even if they use it I guess they do not plan to update this lib also.

sixpeteunder:
I don't know if this is relevant, but as per the doctrine/orm page on Packagist, they only have support for PHP 7.1 from [2.6.0](https://packagist.org/packages/doctrine/orm#v2.6.0) onwards. 

The [latest version, 2.7.3](https://packagist.org/packages/doctrine/orm#v2.7.3), installs doctrine/common 3.0 by default, and the [2.8.x-dev branch](https://packagist.org/packages/doctrine/orm#2.8.x-dev) has support for 3.0 only.

Again, I'm not sure this is relevant, and it's far from a professional opinion. Just sharing this in the hopes of providing insight on what the doctrine community looks like.

svycka:
we just have to drop <=7.1 PHP versions if we want to get the latest doctrine packages there is no other way and I don't think this is a bad thing. Why we need to support EOL versions of PHP.
@AkenRoberts maybe we can vote for this?

jaikdean:
@svycka I agree. Prioritising support for EOL versions of PHP over current versions of package dependencies doesn't make sense.

AkenRoberts:
I'm remote right now, but check `master`, this may have been done there already.

One of the goals there was to bump dependencies and prepare for more modern usage.

2.4.x is essentially deprecated, with only basic bug fixes being applied.

AkenRoberts:
I appreciate the insights and whatnot shared here.

Just to confirm that nothing like this is happening on 2.4.x. So closing mostly because of that.

See #2130 for some general discussion on supporting `doctrine/common` 3.0, and #2159 for an implementation. 👍

PR #2151, opened 2020-07-09
Add doctrine/commons 3.0 support

svycka:


AkenRoberts:
It was a bit more than just Compose changes. See #2159.

Bumping `doctrine/annotations` and removing the conflict is a good idea. 👍

PR #2153, opened 2020-07-10
Fix translatable

### LINKED ISSUE #2152: Translatable doesn't fallback if translations are null
In our system we're using a2lix/translation-form-bundle for managing translations. And this bundle saves missing translations as null. Expected behavior will be that for missing translations it should fallback to default locale.

But after #1709 it's no longer the case: null considered as valid translation now.

mente:
Resolves #2152 while keeping behavior described in #1709

### DIFF
--- a/lib/Gedmo/Translatable/TranslatableListener.php
+++ b/lib/Gedmo/Translatable/TranslatableListener.php
@@ -470,17 +470,15 @@ class TranslatableListener extends MappedEventSubscriber
             );
             // translate object's translatable properties
             foreach ($config['fields'] as $field) {
-                $translated = '';
-                $is_translated = false;
+                $translated = null;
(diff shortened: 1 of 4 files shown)


mente:
I'm sorry, I was targeting master first and it required 7.2, now I check 2.4.x branch and it looks like 5.3 is minimum dependency. While change in class itself is no problem, tests are obviously not 5.3 friendly. Will fix it

mente:
Great, now works :). I can also provide separate PR for master branch if you find this valuable

AkenRoberts:
Much obliged. 🥇

AkenRoberts:
No worries on the separate PR, I'll get it merged upstream. The tests in general need a ton of updating, so no sense in having one perfect one. 😂

Issue #2110, opened 2020-04-08
SoftDeleteableEntity trait getDeletedAt() method doesn't return null


benjamintoussaint:
Hi,

How come the `SoftDeleteableEntity` trait `getDeletedAt()` method doesn't return `null`?

```php
/**
 * Returns deletedAt.
 *
 * @return \DateTime
 */
public function getDeletedAt()
{
    return $this->deletedAt;
}
```

It should be either `null` or a `\DateTime()`, or am I missing something here?

Thank you!

cf. https://github.com/Atlantic18/DoctrineExtensions/blob/0b7bdbefd3d166def27928dcd62ab67c11c8f172/lib/Gedmo/SoftDeleteable/Traits/SoftDeleteableEntity.php#L40

benjamintoussaint:
I guess we are encouraged to use `isDeleted()` method prior to using `getDeletedAt()`.

AkenRoberts:
Good observation. Contrary to the docblock, it will return `null` if `$deletedAt` is not set to a DateTime value.

Updating the docblock for now sounds like a good idea. Then the next major release can focus on adding proper type declarations.

PR #2157, opened 2020-08-04
Document the SoftDeleteable::$deletedAt as nullable property

olix21:
We have some PhpStan issues when using the traits provided by this package.

The issue comes from the fact that the property `deletedAt` will not always be hydrated and can be null


AkenRoberts:
Thank you, @olix21! This sounded familiar, it turns out I fixed in it `master` for #2110 but not 2.4.x. I've cherry picked those changes to 2.4.x also (via 46a6ccfbf8a0342ccca680e3c41c46b96fd09f9c).

PR #2158, opened 2020-08-05
Fix Metadata Cache

pohlaniacz:
The old code (from past 10 years) put third parametr to `save()` as null which prevent symfony from caching it.. Removing it it use 0 (as default parametr from cache providers) which act like infinity.

Without it, my memcache each time regenerates this file..


AkenRoberts:
This has been fixed in `master`. It would be very useful if you could try `3.0-dev` version and provide any feedback.

PR #2159, opened 2020-08-16
Add support for Doctrine/Common 3.0

### LINKED ISSUE #2130: Support doctrine/common 3

AkenRoberts:
Fixes #2130

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,6 @@
     },
     "require-dev": {
         "alcaeus/mongo-php-adapter": "^1.1",
-        "doctrine/mongodb": "^1.6",
         "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.6",
         "friendsofphp/php-cs-fixer": "^2.16",
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
     "require-dev": {
         "alcaeus/mongo-php-adapter": "^1.1",
         "doctrine/mongodb-odm": "^2.0",
-        "doctrine/orm": "^2.6",
+        "doctrine/orm": "^2.6.3",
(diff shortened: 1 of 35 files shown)


composer.json
```diff
@@ -36,22 +36,24 @@
...
         "alcaeus/mongo-php-adapter": "^1.1",
-        "doctrine/mongodb": "^1.6",
         "doctrine/mongodb-odm": "^2.0",
-        "doctrine/orm": "^2.6",
+        "doctrine/orm": "^2.6.3",
         "friendsofphp/php-cs-fixer": "^2.16",
         "phpunit/phpunit": "^8.5",
         "symfony/yaml": "^4.1"
     },
     "conflict": {
```
svycka, line 56: @AkenRoberts why do we need the `conflict` section?

svycka:
@AkenRoberts maybe release when this merged? or there is something else that keeps from releasing? It's a bit problematic that 2.* not supported at least you decided to decline all feature requests or updates like supporting new doctrine libraries and we don't have any version where we could work. So I suggest we release 3.0 if it works, we should have released it long ago. I do have write permissions to release 3.0 but you did a lot of work and I will let you decide when it is ready I believe you know better. I believe 3.0 release should have been just versions of bumps and some cleanups but it got much more.

AkenRoberts:
@sbczk Nothing really keeping a release at this point except for lack of feedback to tell if the things we've updated so far are stable and working as expected.

I suppose there's no better way to find that out than releasing and getting it into people's projects. I'll tag a release candidate for 3.0 tonight.

PR #2163, opened 2020-08-24
Migrated zend to laminas

svycka:
Renamed Zend Framework to Laminas and removed `zenddate` date format because `\Zend_Date` is no longer supported for many years


AkenRoberts:
Nice change 👍 I'm happy to merge this in if you could also note the BC change in the Changelog.

svycka:
@AkenRoberts done

AkenRoberts:
Thank you! ❤️

PR #2166, opened 2020-08-25
[Translatable] Add an aggregate function to TranslationWalker when using GROUP BY

### LINKED ISSUE #2164: [Translatable] GROUP BY in MySQL 5.7 with sql_mode=only_full_group_by
Loading entities with the Translatable extension and grouping those in any way creates SQL:

```sql
SELECT d0_.id AS id_0, COALESCE(t1_.content, d0_.title) AS title_1, 
COALESCE(t2_.content, d0_.description) AS description_2

FROM Data d0_ 
LEFT JOIN DataTranslation t1_ ON t1_.locale = 'de' AND t1_.field = 'title' AND t1_.object_id = d0_.id 
LEFT JOIN DataTranslation t2_ ON t2_.locale = 'de' AND t2_.field = 'description' AND t2_.object_id = d0_.id 

GROUP BY d0_.id 
```
Which produces the error:
SQLSTATE[42000]: Syntax error or access violation: 1055 Expression # 2 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'data_test.t1_.content' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

Solution would be to either add `content` field to the `GROUP BY` by the TranslationWalker, or wrap those into `MAX()` aggregate function, as it anyway produces one result per row.

biozshock:
Fixes #2164

### DIFF
--- a/src/Translatable/Query/TreeWalker/TranslationWalker.php
+++ b/src/Translatable/Query/TreeWalker/TranslationWalker.php
@@ -346,6 +346,10 @@ class TranslationWalker extends SqlWalker
                 $originalField = $compTblAlias.'.'.$quoteStrategy->getColumnName($field, $meta, $this->platform);
                 $substituteField = $tblAlias.'.'.$quoteStrategy->getColumnName('content', $transMeta, $this->platform);
 
+                if (null !== $q->getAST()->groupByClause) {
+                    $substituteField = 'MAX('.$substituteField.')';
+                }
+
                 // Treat translation as original field type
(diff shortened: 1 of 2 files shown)


biozshock:
Bump

franmomu:
Hi @biozshock, thanks for the PR, can you please target `main` instead of `2.4`?

I have to take a look at the issue again, but looking at the test I think it does not hit this new `if`.

PR #2170, opened 2020-09-24

Evertt:
I made the problem clear with a modified test in the first commit.

Before my fix the test results in:

    Failed asserting that two strings are equal.
    Expected: '123-2.text'
    Actual: '123-2.txt'

### DIFF
--- a/tests/Gedmo/Uploadable/UploadableEntityTest.php
+++ b/tests/Gedmo/Uploadable/UploadableEntityTest.php
@@ -377,19 +377,24 @@ class UploadableEntityTest extends BaseTestCaseORM
 
     public function testFileWithCustomFilenameGenerator()
     {
-        $file = new FileWithCustomFilenameGenerator();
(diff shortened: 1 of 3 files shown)


tests/Gedmo/Uploadable/UploadableEntityTest.php
```diff
@@ -377,19 +377,24 @@ public function testFileWithFilenameAlphanumericGenerator()
...
+            $this->em->refresh($file);
 
-        $filename = substr($file->getFilePath(), strrpos($file->getFilePath(), '/') + 1);
+            $filename = substr($file->getFilePath(), strrpos($file->getFilePath(), '/') + 1);
+
+            $this->assertEquals($expected, $filename);
+        };
 
-        $this->assertEquals('123.txt', $filename);
+        $uploadFile('123.text');
```
phansys, line 396: If we need to perform the same test through multiple cases, I think we should use a data provider instead.

PR #2128, opened 2020-05-25
[Loggable] Allow composite identifiers

MisatoTremor:
Composite identifiers with foreign entities availaible for ORM > 2.6 due to earlier version not handling identifiers for newly inserted foreign objects

PR #2171, opened 2020-10-06
[Loggable] Allow composite identifiers

MisatoTremor:
Composite identifiers with foreign entities

New version of #2128 as of `main` branch rename

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ a release.
 ### Added
 - Tree: `setSibling()` and `getSibling()` methods in the `Node` interface through the BC `@method` annotation
 - Tree: Support array of fields and directions in the `$sortByField` and `$direction` parameters at `AbstractTreeRepository::recover()`
+- Loggable: Support for composite identifiers.
 
 ### Changed
 - Named arguments have precedence over the values passed in the `$data` array in annotation classes at `Gedmo\Mapping\Annotation\`
@@ -34,6 +35,8 @@ a release.
 
 ### Deprecated
 - Tree: Not implementing `Node` interface in classes that are used as nodes
+- Implementing the `Gedmo\Tool\WrapperInterface::getIdentifier()` method without the second argument (`$flatten`) is deprecated and will
+  be required in version 4.0.
 
 ## [3.11.1] - 2023-02-20
 ### Fixed
--- a/src/Loggable/Document/Repository/LogEntryRepository.php
+++ b/src/Loggable/Document/Repository/LogEntryRepository.php
@@ -106,10 +106,11 @@ class LogEntryRepository extends DocumentRepository
             throw new UnexpectedValueException('Count not find any log entries under version: '.$version);
         }
 
-        $data = [];
+        $data = [[]];
         while ($log = array_shift($logs)) {
-            $data = array_merge($data, $log->getData());
+            $data[] = $log->getData();
         }
+        $data = array_merge(...$data);
         $this->fillDocument($document, $data);
     }
 
--- a/src/Loggable/Mapping/Driver/Annotation.php
+++ b/src/Loggable/Mapping/Driver/Annotation.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Loggable\Mapping\Driver;
 
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ClassMetadataODM;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Gedmo\Exception\InvalidMappingException;
 use Gedmo\Mapping\Annotation\Loggable;
@@ -40,7 +41,7 @@ class Annotation extends AbstractAnnotationDriver
 
     public function validateFullMetadata(ClassMetadata $meta, array $config)
     {
-        if ($config && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
+        if ($config && $meta instanceof ClassMetadataODM && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
             throw new InvalidMappingException("Loggable does not support composite identifiers in class - {$meta->getName()}");
         }
         if (isset($config['versioned']) && !isset($config['loggable'])) {
@@ -87,7 +88,7 @@ class Annotation extends AbstractAnnotationDriver
         }
 
         if (!$meta->isMappedSuperclass && $config) {
-            if (is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
+            if ($meta instanceof ClassMetadataODM && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
                 throw new InvalidMappingException("Loggable does not support composite identifiers in class - {$meta->getName()}");
             }
             if ($this->isClassAnnotationInValid($meta, $config)) {
--- a/src/Loggable/Mapping/Driver/Xml.php
+++ b/src/Loggable/Mapping/Driver/Xml.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Loggable\Mapping\Driver;
 
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as ClassMetadataODM;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Gedmo\Exception\InvalidMappingException;
 use Gedmo\Mapping\Driver\Xml as BaseXml;
@@ -37,7 +38,7 @@ class Xml extends BaseXml
 
         $xml = $xml->children(self::GEDMO_NAMESPACE_URI);
 
-        if ('entity' === $xmlDoctrine->getName() || 'document' === $xmlDoctrine->getName() || 'mapped-superclass' === $xmlDoctrine->getName()) {
+        if (in_array($xmlDoctrine->getName(), ['mapped-superclass', 'entity', 'document'], true)) {
             if (isset($xml->loggable)) {
                 /**
                  * @var \SimpleXMLElement;
@@ -74,7 +75,7 @@ class Xml extends BaseXml
         }
 
         if (!$meta->isMappedSuperclass && $config) {
-            if (is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
+            if ($meta instanceof ClassMetadataODM && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
                 throw new InvalidMappingException("Loggable does not support composite identifiers in class - {$meta->getName()}");
             }
             if (isset($config['versioned']) && !isset($config['loggable'])) {
--- a/src/Tool/Wrapper/EntityWrapper.php
+++ b/src/Tool/Wrapper/EntityWrapper.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Tool\Wrapper;
 
+use Doctrine\Common\Util\ClassUtils;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Proxy\Proxy;
@@ -77,33 +78,32 @@ class EntityWrapper extends AbstractWrapper
         return $this->meta->rootEntityName;
     }
 
-    public function getIdentifier($single = true)
+    public function getIdentifier($single = true, $flatten = false)
     {
+        $flatten = 1 < \func_num_args() && true === func_get_arg(1);
         if (null === $this->identifier) {
-            if ($this->object instanceof Proxy) {
-                $uow = $this->om->getUnitOfWork();
-                if ($uow->isInIdentityMap($this->object)) {
-                    $this->identifier = $uow->getEntityIdentifier($this->object);
-                } else {
-                    $this->initialize();
-                }
-            }
-            if (null === $this->identifier) {
-                $this->identifier = [];
-                $incomplete = false;
-                foreach ($this->meta->identifier as $name) {
-                    $this->identifier[$name] = $this->getPropertyValue($name);
-                    if (null === $this->identifier[$name]) {
-                        $incomplete = true;
-                    }
-                }
-                if ($incomplete) {
-                    $this->identifier = null;
-                }
+            $uow = $this->om->getUnitOfWork();
+            $this->identifier = $uow->isInIdentityMap($this->object)
+                ? $uow->getEntityIdentifier($this->object)
+                : $this->meta->getIdentifierValues($this->object);
+            if (is_array($this->identifier) && empty($this->identifier)) {
+                $this->identifier = null;
             }
         }
-        if ($single && is_array($this->identifier)) {
-            return reset($this->identifier);
+        if (is_array($this->identifier)) {
+            if ($single) {
+                return reset($this->identifier);
+            }
+            if ($flatten) {
+                $id = $this->identifier;
+                foreach ($id as $i => $value) {
+                    if (is_object($value) && $this->om->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+                        $id[$i] = (new self($value, $this->om))->getIdentifier(false, true);
+                    }
+                }
+
+                return implode(' ', $id);
+            }
         }
 
         return $this->identifier;
--- a/src/Tool/Wrapper/MongoDocumentWrapper.php
+++ b/src/Tool/Wrapper/MongoDocumentWrapper.php
@@ -76,7 +76,7 @@ class MongoDocumentWrapper extends AbstractWrapper
         return (bool) $this->getIdentifier();
     }
 
-    public function getIdentifier($single = true)
+    public function getIdentifier($single = true, $flatten = false)
     {
         if (!$this->identifier) {
             if ($this->object instanceof GhostObjectInterface) {
--- a/src/Tool/WrapperInterface.php
+++ b/src/Tool/WrapperInterface.php
@@ -75,10 +75,13 @@ interface WrapperInterface
      * Get the object identifier, single or composite.
      *
      * @param bool $single
+     * @param bool $flatten
      *
      * @return array|mixed Array if a composite value, otherwise a single scalar
+     *
+     * @todo Uncomment the second parameter for 4.0
      */
-    public function getIdentifier($single = true);
+    public function getIdentifier($single = true/* , $flatten = false */);
 
     /**
      * Get the root object class name.
--- a/tests/Gedmo/Loggable/Fixture/Entity/Composite.php
+++ b/tests/Gedmo/Loggable/Fixture/Entity/Composite.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Tests\Loggable\Fixture\Entity;
+
+use Doctrine\DBAL\Types\Types;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Mapping\Annotation as Gedmo;
+
+/**
+ * @ORM\Entity
+ * @Gedmo\Loggable
+ */
+#[ORM\Entity]
+#[Gedmo\Loggable]
+class Composite
+{
+    /**
+     * @var int
+     *
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     */
+    #[ORM\Id]
+    #[ORM\Column(name: 'one', type: Types::INTEGER)]
+    private $one;
+
+    /**
+     * @var int
+     *
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     */
+    #[ORM\Id]
+    #[ORM\Column(name: 'two', type: Types::INTEGER)]
+    private $two;
+
+    /**
+     * @var string
+     *
+     * @ORM\Column(length=8)
+     * @Gedmo\Versioned
+     */
+    #[ORM\Column(name: 'title', type: Types::STRING, length: 8)]
+    #[Gedmo\Versioned]
+    private $title;
+
+    public function __construct(int $one, int $two)
+    {
+        $this->one = $one;
+        $this->two = $two;
+    }
+
+    public function getOne(): int
+    {
+        return $this->one;
+    }
+
+    public function getTwo(): int
+    {
+        return $this->two;
+    }
+
+    public function setTitle(string $title): void
+    {
+        $this->title = $title;
+    }
+
+    public function getTitle(): string
+    {
+        return $this->title;
+    }
+}
--- a/tests/Gedmo/Loggable/LoggableEntityTest.php
+++ b/tests/Gedmo/Loggable/LoggableEntityTest.php
@@ -16,6 +16,8 @@ use Gedmo\Loggable\Entity\Repository\LogEntryRepository;
 use Gedmo\Tests\Loggable\Fixture\Entity\Address;
 use Gedmo\Tests\Loggable\Fixture\Entity\Article;
 use Gedmo\Tests\Loggable\Fixture\Entity\Comment;
+use Gedmo\Tests\Loggable\Fixture\Entity\Composite;
+use Gedmo\Tests\Loggable\Fixture\Entity\CompositeRelation;
 use Gedmo\Tests\Loggable\Fixture\Entity\Geo;
 use Gedmo\Tests\Loggable\Fixture\Entity\GeoLocation;
 use Gedmo\Tests\Loggable\Fixture\Entity\RelatedArticle;
@@ -30,6 +32,8 @@ abstract class LoggableEntityTest extends BaseTestCaseORM
 {
     public const ARTICLE = Article::class;
     public const COMMENT = Comment::class;
+    public const COMPOSITE = Composite::class;
+    public const COMPOSITE_RELATION = CompositeRelation::class;
     public const RELATED_ARTICLE = RelatedArticle::class;
     public const COMMENT_LOG = \Gedmo\Tests\Loggable\Fixture\Entity\Log\Comment::class;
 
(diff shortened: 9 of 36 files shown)


rotdrop:
Could this possibly be merged? IMHO it is an important feature to add support for composite identifiers.

src/Tool/Wrapper/EntityWrapper.php
```diff
@@ -80,34 +81,32 @@ public function getRootObjectName()
...
+                foreach ($id as $i => $value) {
+                    if (is_object($value) && $this->om->getMetadataFactory()->hasMetadataFor(ClassUtils::getClass($value))) {
+                        $id[$i] = (new EntityWrapper($value, $this->om))->getIdentifier(false, true);
                     }
                 }
-                if ($incomplete) {
-                    $this->identifier = null;
-                }
+
+                return implode(' ', $id);
```
rotdrop, line 107: Is it clear the the identifier always "come" in the same ordering, or is it maybe necessary to sort them by their keys?

MisatoTremor: Nope. AFAIK there is no defined identifier order in Doctrine. As of now they are in the order they are defined in the mapping.
For annotations this means if you change the order of the Id-annotated properties in your code, the order of the composite identifier does change as well.

rotdrop: Well, one could try to sort the identifier array by its keys: 
```
ksort($id);
return implode(' ', $id);
```
BTW, thank you for your patch, which unfortunately still is not merged. I have also copied over you approach to the "Translatable" extension here: 990252dd72ca3132f177758f6314ce96740acebd

MisatoTremor: Yep. But that would be the 'different behaviour' I mentioned. I'm unsure if that is a good idea or if such change would better be proposed in ORM.

Also, nice patch there!

I'm already using my patch in our fork, and was half-planning this as a proof-of-concept and if approved planned to implement it in the other components, too.

rotdrop: Concerning "different behaviour": the "flatten" option is something Gedmo-specific, so here one has the freedom to define it to ones liking, this does not interfere with the ORM. The mere implode already looses information about the field-names, which are there in the original id-array, but are forgotten when just imploding the values. The ksort would make the flatten procedure independent from the more-or-less arbitrary ordering of the  FIELD => FIELD_VALUE pairs in the ORM id-array while still collapsing to the previous single-value thing for non-composite ids.

MisatoTremor: Yeah, it is. And I'm completely with you on having sorted identifiers. Maybe we should do it as a temporary solution, but I still prefer this to be "fixed" on ORM level in the long term.

Also TIL that ORM itself has a IdentifierFlattener. It might make sense to use this instead of the simple foreach. 🤔

MisatoTremor:
Rebased and fixed PHP and XML code style

MisatoTremor:
Well, finally adopted the changed test namespaces.

@phansys Could you give me a heads up, where this breaks BC, please? Maybe this can be changed

src/Tool/Wrapper/EntityWrapper.php
```diff
@@ -81,34 +82,32 @@ public function getRootObjectName()
     /**
      * {@inheritdoc}
      */
-    public function getIdentifier($single = true)
+    public function getIdentifier($single = true, $flatten = false)
```
phansys, line 85: IMO, adding an argument to a method that can be overridden by the user is a BC break.

phansys:
>@phansys Could you give me a heads up, where this breaks BC, please? Maybe this can be changed

Sure: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2171#discussion_r749102248.

src/Tool/Wrapper/EntityWrapper.php
MisatoTremor: Ah yes. Forgot about that. You are right, I'd also see that as a BC break.
I could alleviate this by leaving the argument list as is and using `func_get_args` for the time beeing and adding a deprecation message. Then just change the argument list when the next major is going to be released.
This approach has been used by e.g. Symfony for forward compatibility.

Would it be OK to use it here?

phansys: Sure, we can provide a BC layer this way.

MisatoTremor: Ich have implemented the method in the BC compatible way. I just ommited the deprecation message, as I found no soild way to have it always run.

MisatoTremor:
Resolved conflicts and code style

@phansys Could you please verify this no longer breaks BC?

src/Loggable/Mapping/Driver/Annotation.php
```diff
@@ -35,7 +35,7 @@ class Annotation extends AbstractAnnotationDriver
      */
     public function validateFullMetadata(ClassMetadata $meta, array $config)
     {
-        if ($config && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
+        if ($config && $meta instanceof \Doctrine\ODM\MongoDB\Mapping\ClassMetadata && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
```
phansys, line 38: Could we import this class?

src/Tool/Wrapper/EntityWrapper.php
```diff
@@ -81,34 +82,33 @@ public function getRootObjectName()
     /**
      * {@inheritdoc}
      */
-    public function getIdentifier($single = true)
+    public function getIdentifier($single = true/*, $flatten = false*/)
     {
+        $flatten = 1 < \func_num_args() && func_get_arg(1);
```
phansys, line 87:
```suggestion
        $flatten = 1 < \func_num_args() && true === func_get_arg(1);
```

src/Tool/WrapperInterface.php
```diff
@@ -63,10 +63,11 @@ public function getMetadata();
      * Get the object identifier, single or composite.
      *
      * @param bool $single
+     * @param bool $flatten
      *
      * @return array|mixed Array if a composite value, otherwise a single scalar
      */
-    public function getIdentifier($single = true);
+    public function getIdentifier($single = true/*, $flatten = false*/);
```
phansys, line 70: Please, add a `@todo` docblock explaining what to do in the next major release.
BTW, this change also requires an entry in the changelog.

tests/Gedmo/Loggable/Fixture/Entity/Composite.php
```diff
@@ -0,0 +1,56 @@
...
+     */
+    private $two;
+
+    /**
+     * @ORM\Column(length=128)
+     * @Gedmo\Versioned
+     */
+    private $title;
+
+    public function __construct($one, $two)
```
phansys, line 31:
```suggestion
    public function __construct(int $one, int $two)
```

tests/Gedmo/Loggable/Fixture/Entity/Composite.php
```diff
@@ -0,0 +1,56 @@
...
+ * @ORM\Entity
+ * @Gedmo\Loggable
+ */
+class Composite
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     */
+    private $one;
```
phansys, line 18: Please, add type hints for these properties and argument / return declarations in the methods.

tests/Gedmo/Loggable/LoggableEntityTest.php
```diff
@@ -155,13 +159,117 @@ public function testLogEmbedded()
         static::assertCount(5, $logEntries[3]->getData());
     }
 
+    public function testComposite()
+    {
+        $logRepo = $this->em->getRepository('Gedmo\Loggable\Entity\LogEntry');
```
phansys, line 164: Please, use `LogEntry::class` instead of the hardcoded string.

tests/Gedmo/Loggable/LoggableEntityTest.php
```diff
@@ -155,13 +159,117 @@ public function testLogEmbedded()
         static::assertCount(5, $logEntries[3]->getData());
     }
 
+    public function testComposite()
```
phansys, line 162:
```suggestion
    public function testComposite(): void
```

tests/Gedmo/Loggable/LoggableEntityTest.php
```diff
@@ -155,13 +159,117 @@ public function testLogEmbedded()
...
+        $this->em->remove($composite);
+        $this->em->flush();
+        $this->em->clear();
+
+        $log = $logRepo->findOneBy(['version' => 3, 'objectId' => $cmpId]);
+        static::assertSame('remove', $log->getAction());
+        static::assertNull($log->getData());
+    }
+
+    public function testCompositeRelation()
```
phansys, line 210:
```suggestion
    public function testCompositeRelation(): void
```

tests/Gedmo/Loggable/LoggableEntityTest.php
```diff
@@ -155,13 +159,117 @@ public function testLogEmbedded()
...
+        $compositeRepo = $this->em->getRepository(self::COMPOSITE);
+        static::assertCount(0, $logRepo->findAll());
+
+        $cmp0 = new Composite(1, 2);
+        $cmp0->setTitle('Title2');
+
+        $this->em->persist($cmp0);
+        $this->em->flush();
+
+        $cmpId = sprintf('%s %s', 1, 2);
```
phansys, line 174:
```suggestion
        $compositeIds = [1, 2];

        $cmp = new Composite(...$compositeIds);
        $cmp->setTitle('Title2');

        $this->em->persist($cmp);
        $this->em->flush();

        $cmpId = sprintf('%s %s', ...$compositeIds);
```

tests/Gedmo/Mapping/LoggableMappingTest.php
```diff
@@ -36,6 +40,15 @@ protected function setUp(): void
             new YamlDriver([__DIR__.'/Driver/Yaml']),
             'Gedmo\Tests\Mapping\Fixture\Yaml'
         );
+        $reader = new \Doctrine\Common\Annotations\AnnotationReader();
+        \Doctrine\Common\Annotations\AnnotationRegistry::registerAutoloadNamespace(
+            'Gedmo\\Mapping\\Annotation',
+            VENDOR_PATH.'/../lib'
```
phansys, line 46: This path does not exist.

tests/Gedmo/Mapping/LoggableMappingTest.php
```diff
@@ -36,6 +40,15 @@ protected function setUp(): void
...
+        );
+        $chainDriverImpl->addDriver(
+            new \Doctrine\ORM\Mapping\Driver\AnnotationDriver($reader),
```
phansys, line 49: Please, import this class.

tests/Gedmo/Mapping/LoggableMappingTest.php
```diff
@@ -36,6 +40,15 @@ protected function setUp(): void
             new YamlDriver([__DIR__.'/Driver/Yaml']),
             'Gedmo\Tests\Mapping\Fixture\Yaml'
         );
+        $reader = new \Doctrine\Common\Annotations\AnnotationReader();
+        \Doctrine\Common\Annotations\AnnotationRegistry::registerAutoloadNamespace(
```
phansys, line 44: Please, import this class.

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -58,6 +62,47 @@ public function testProxy()
...
     }
 
+    public function testComposite()
+    {
+        $test = $this->em->getReference(self::COMPOSITE, ['one' => 1, 'two' => 2]);
+        static::assertInstanceOf(self::COMPOSITE, $test);
+        $wrapped = new EntityWrapper($test, $this->em);
+
+        $id = $wrapped->getIdentifier(false);
+        static::assertTrue(is_array($id));
```
phansys, line 72:
```suggestion
        static::assertIsArray($id);
```

phansys (changes requested):

phansys:
>@phansys Could you please verify this no longer breaks BC?

I think the current changes respect BC, thank you :+1:

tests/Gedmo/Loggable/Fixture/Entity/Composite.php
MisatoTremor: Can do properties as this is still set to support PHP < 7.4

phansys: With type hints I mean `@var` docblocks, not type declarations.

MisatoTremor:
Resolved all requests

src/Loggable/Mapping/Driver/Xml.php
```diff
@@ -68,7 +68,7 @@ public function readExtendedMetadata($meta, array &$config)
         }
 
         if (!$meta->isMappedSuperclass && $config) {
-            if (is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
+            if ($meta instanceof \Doctrine\ODM\MongoDB\Mapping\ClassMetadata && is_array($meta->getIdentifier()) && count($meta->getIdentifier()) > 1) {
```
phansys, line 71: Please, import the classes and apply the same change with all the new class usages introduced in this PR.

tests/Gedmo/Mapping/Fixture/Loggable.php
```diff
@@ -0,0 +1,41 @@
...
+ * @Gedmo\Loggable
+ */
+class Loggable
+{
+    /**
+     * @ORM\Id
+     * @ORM\GeneratedValue
+     * @ORM\Column(type="integer")
+     */
+    private $id;
```
phansys, line 19: Please, use type hints and type declarations where possible (in this and the remaining classes / methods introduced by this PR).

phansys (changes requested):

MisatoTremor:
@phansys Hopefully I catched all missing type hints and imports. The PHPStan error seems to be unrelated to this PR

src/Loggable/Document/Repository/LogEntryRepository.php
```diff
@@ -88,10 +88,11 @@ public function revert($document, $version = 1)
             $logs = $logs->toArray();
         }
         if ($logs) {
-            $data = [];
+            $data = [[]];
             while (($log = array_shift($logs))) {
```
phansys, line 92:
```suggestion
            while ($log = array_shift($logs)) {
```

src/Loggable/Mapping/Driver/Xml.php
```diff
@@ -37,7 +38,7 @@ public function readExtendedMetadata($meta, array &$config)
 
         $xml = $xml->children(self::GEDMO_NAMESPACE_URI);
 
-        if ('entity' == $xmlDoctrine->getName() || 'document' == $xmlDoctrine->getName() || 'mapped-superclass' == $xmlDoctrine->getName()) {
+        if (in_array($xmlDoctrine->getName(), ['mapped-superclass', 'entity', 'document'])) {
```
phansys, line 41:
```suggestion
        if (in_array($xmlDoctrine->getName(), ['mapped-superclass', 'entity', 'document'], true)) {
```

src/Tool/Wrapper/EntityWrapper.php
```diff
@@ -87,34 +88,33 @@ public function getRootObjectName()
...
-                if ($uow->isInIdentityMap($this->object)) {
-                    $this->identifier = $uow->getEntityIdentifier($this->object);
-                } else {
-                    $this->initialize();
-                }
+            $uow = $this->om->getUnitOfWork();
+            $this->identifier = $uow->isInIdentityMap($this->object)
+                ? $uow->getEntityIdentifier($this->object)
+                : $this->meta->getIdentifierValues($this->object);
+            if ((is_array($this->identifier) && empty($this->identifier))) {
```
phansys, line 99:
```suggestion
            if (is_array($this->identifier) && empty($this->identifier)) {
```

src/Tool/Wrapper/MongoDocumentWrapper.php
```diff
@@ -86,7 +86,7 @@ public function hasValidIdentifier()
     /**
      * {@inheritdoc}
      */
-    public function getIdentifier($single = true)
+    public function getIdentifier($single = true, $flatten = false)
```
phansys, line 89: This class / method is not final, so it is a BC break.

tests/Gedmo/Mapping/Xml/LoggableMappingTest.php
```diff
@@ -83,6 +85,36 @@ public function testLoggableMetadata()
         static::assertContains('status', $config['versioned']);
     }
 
+    public function testLoggableCompositeMetadata()
```
phansys, line 88:
```suggestion
    public function testLoggableCompositeMetadata(): void
```

tests/Gedmo/Mapping/Xml/LoggableMappingTest.php
```diff
@@ -83,6 +85,36 @@ public function testLoggableMetadata()
...
+        static::assertSame(LogEntry::class, $config['logEntryClass']);
+        static::assertArrayHasKey('loggable', $config);
+        static::assertTrue($config['loggable']);
+
+        static::assertArrayHasKey('versioned', $config);
+        static::assertCount(1, $config['versioned']);
+        static::assertContains('title', $config['versioned']);
+    }
+
+    public function testLoggableCompositeRelationMetadata()
```
phansys, line 103:
```suggestion
    public function testLoggableCompositeRelationMetadata(): void
```

tests/Gedmo/Mapping/Yaml/LoggableMappingTest.php
```diff
@@ -63,6 +65,36 @@ protected function setUp(): void
         ], $chain);
     }
 
+    public function testLoggableCompositeMetadata()
```
phansys, line 68:
```suggestion
    public function testLoggableCompositeMetadata(): void
```

tests/Gedmo/Mapping/Yaml/LoggableMappingTest.php
```diff
@@ -63,6 +65,36 @@ protected function setUp(): void
...
+        static::assertSame(LogEntry::class, $config['logEntryClass']);
+        static::assertArrayHasKey('loggable', $config);
+        static::assertTrue($config['loggable']);
+
+        static::assertArrayHasKey('versioned', $config);
+        static::assertCount(1, $config['versioned']);
+        static::assertContains('title', $config['versioned']);
+    }
+
+    public function testLoggableCompositeRelationMetadata()
```
phansys, line 83:
```suggestion
    public function testLoggableCompositeRelationMetadata(): void
```

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -63,6 +67,47 @@ public function testProxy()
         static::assertSame('test', $wrapped->getPropertyValue('title'));
     }
 
+    public function testComposite()
```
phansys, line 70:
```suggestion
    public function testComposite(): void
```

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -63,6 +67,47 @@ public function testProxy()
...
+        static::assertSame(2, $id['two']);
+
+        $id = $wrapped->getIdentifier(false, true);
+        static::assertIsString($id);
+        static::assertSame('1 2', $id);
+
+        static::assertSame('test', $wrapped->getPropertyValue('title'));
+    }
+
+    public function testCompositeRelation()
```
phansys, line 91:
```suggestion
    public function testCompositeRelation(): void
```

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -83,6 +128,28 @@ public function testDetachedProxy()
         static::assertSame('test', $wrapped->getPropertyValue('title'));
     }
 
+    public function testDetachedCompositeRelation()
```
phansys, line 131:
```suggestion
    public function testDetachedCompositeRelation(): void
```

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -83,6 +128,28 @@ public function testDetachedProxy()
...
+    {
+        $test = $this->em->getReference(self::COMPOSITE_RELATION, ['article' => 1, 'status' => 2]);
+        $this->em->clear();
+        $wrapped = new EntityWrapper($test, $this->em);
+
+        static::assertSame('1 2', $wrapped->getIdentifier(false, true));
+        static::assertSame('test', $wrapped->getPropertyValue('title'));
+    }
+
+    public function testCompositeRelationProxy()
```
phansys, line 141:
```suggestion
    public function testCompositeRelationProxy(): void
```

tests/Gedmo/Wrapper/EntityWrapperTest.php
```diff
@@ -83,6 +128,28 @@ public function testDetachedProxy()
...
+    {
+        $this->em->clear();
+        $art1 = $this->em->getReference(self::ARTICLE, ['id' => 1]);
+        $test = $this->em->getReference(self::COMPOSITE_RELATION, ['article' => $art1->getId(), 'status' => 2]);
+        static::assertInstanceOf('Doctrine\\ORM\\Proxy\\Proxy', $test);
```
phansys, line 146: Could we use the real class instead of this string?

src/Tool/Wrapper/MongoDocumentWrapper.php
MisatoTremor: Missed that one, thanks

src/Loggable/Mapping/Driver/Xml.php
MisatoTremor: I also fixed all other `$xmlDoctrine->getName()` comparisons (with multiple values)

doc/loggable.md
```diff
@@ -250,7 +250,7 @@ Now lets revert it to previous version:
 ``` php
 <?php
 // first check our log entries
-$repo = $em->getRepository('Gedmo\Loggable\Entity\LogEntry'); // we use default log entry class
+$repo = $em->getRepository(LogEntry::class); // we use default log entry class
```
phansys, line 253:
```suggestion
$repo = $em->getRepository(\Gedmo\Loggable\Entity\LogEntry::class); // we use default log entry class
```
Otherwise, we should import the class with `use`.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ a release.
 - Support for doctrine/dbal 3.x
 - Timestampable: Support to use annotations as attributes on PHP >= 8.0.
 
+### Deprecated
+- Implementing the `Gedmo\Tool\WrapperInterface::getIdentifier()` method without the `$flatten` parameter is deprecated and will
```
phansys, line 27:
```suggestion
- Implementing the `Gedmo\Tool\WrapperInterface::getIdentifier()` method without the second argument (`$flatten`) is deprecated and will
```

phansys:
BTW, I think the changelog section should include an `Added` entry, as this is a new feature.

MisatoTremor:
You are right of course, doesn't make sense to not mention it 😅

CHANGELOG.md
```diff
@@ -22,6 +22,11 @@ a release.
 ### Added
 - Support for doctrine/dbal 3.x
 - Timestampable: Support to use annotations as attributes on PHP >= 8.0.
+- Loggable: Support for composite identifiers
+
+### Deprecated
+- Implementing the `Gedmo\Tool\WrapperInterface::getIdentifier()` method second argument (`$flatten`) parameter is deprecated and will
```
phansys, line 28:
```suggestion
- Implementing the `Gedmo\Tool\WrapperInterface::getIdentifier()` method without the second argument (`$flatten`) is deprecated and will
```

MisatoTremor:
Resolved conflicts.

Could you please give me a heads up why my changes might remove coverage [here](https://app.codecov.io/gh/doctrine-extensions/DoctrineExtensions/compare/2171/changes#D1L120)  @phansys?  Or should I ignore that?

phansys:
>Could you please give me a heads up why my changes might remove coverage here @phansys? Or should I ignore that?

I think it is due the way Codecov makes the [comparison with the parent commit](https://docs.codecov.com/docs/comparing-commits#choosing-a-parent-commit).
I guess the coverage report will be more accurate after a rebasing and squashing the commits in this PR. I'd sat you can ignore this warning by now.
BTW, please note that the PR still has conflicts.

MisatoTremor:
@phansys Thank you for your answer.

I can't see any conflicts, though.

phansys:
>I can't see any conflicts, though.

Don't you see this message?
>This branch cannot be rebased due to conflicts

image

MisatoTremor:
Nope, I get this
Screenshot 2021-12-07 at 21 32 06

CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- Loggable: Support for composite identifiers
```
phansys, line 23:
```suggestion
- Loggable: Support for composite identifiers.
```

phansys (approved):

MisatoTremor:
OK done @phansys. Just wanted to ask if you could also check that I did not reintroduce some of the old issues?

phansys:
>OK done @phansys. Just wanted to ask if you could also check that I did not reintroduce some of the old issues?

I'm not aware about the related issues, as I'm not using this extension. If you have in mind some previous issues related to these changes, please try to add tests covering the issues with a clear reference to them.

MisatoTremor:
@phansys I was just referring to the issues you pointed out in your reviews. To make sure I did not reintroduce one of them while rebasing.

phansys:
>@phansys I was just referring to the issues you pointed out in your reviews. To make sure I did not reintroduce one of them while rebasing.

Oh, don't worry about that then :+1:

franmomu:
Hi, thanks for the PR! 

Just a couple of comments, looking at the method calls `->getIdentifier(false, true)` seems a little bit weird to me and the way the ids are flattened (`return implode(' ', $id);`) seems like it should belong somewhere else to me.

So what about making `getIdentifier` return always an array (deprecating passing `true`) like the one from `ClassMetadata` and also extracting the "flattener" to a class so it could be used somewhere else.

MisatoTremor:
I'm open to that @franmomu of course.
But it might be out of scope for this PR as it would need changes all over DoctrineExtensions. So maybe it warrants a prerequisite PR for itself.

But would leave that decision up to you maintainers before starting work on it.

rotdrop:
> So what about making `getIdentifier` return always an array (deprecating passing `true`) like the one from `ClassMetadata` and also extracting the "flattener" to a class so it could be used somewhere else.

The flattener should probably kept in sync with

https://github.com/doctrine/orm/blob/536b65f02b853cc9c8bfb184d23a4cc4feecf002/lib/Doctrine/ORM/Utility/IdentifierFlattener.php#L15-L20

which actually just does what Steffen does. Or perhaps use the `IdentifierFlattener` from ORM instead of coding the same thing again?

I'm still trying to find time to post a corresponding pull request for Translatable, and other things ... it would be nice, if _this_ PR could finally be merged.

MisatoTremor:
@rotdrop 
I remember using `IdentifierFlattener` during initial development of this, which did not work at the time, but that may have change since then ...

@phansys 
Could you please have another look at this? 
I am currently lost as how to prevent the deprecation reports by PHPUnit which lead to the failed tests. Also I have no idea why the PHPStan error in Geo.php suddenly appears.

phansys:
>I am currently lost as how to prevent the deprecation reports by PHPUnit which lead to the failed tests.

I guess you should add the new `$flatten` argument in the existing calls to `WrapperInterface::getIdentifier()`.

>Also I have no idea why the PHPStan error in Geo.php suddenly appears.

I'll try to take a look at this.

phansys:
>Also I have no idea why the PHPStan error in Geo.php suddenly appears.

Regarding this, I think we could safely remove that ignored error from the baseline (see https://phpstan.org/r/b92f2814-4530-4cda-aaf4-236c6c3eaadf).

https://github.com/doctrine-extensions/DoctrineExtensions/blob/118a001c1d0855df5c9ee35c1588d80a4d517d97/phpstan-baseline.neon#L808-L812

MisatoTremor:
> I guess you should add the new `$flatten` argument in the existing calls to `WrapperInterface::getIdentifier()`.

The calls were not the problem, but the classes implementing the interface, only adding the second parameter there resolved the error. Changing the interface was also not necessary. 

> Regarding this, I think we could safely remove that ignored error from the baseline (see https://phpstan.org/r/b92f2814-4530-4cda-aaf4-236c6c3eaadf).

Done. Thank you!

MisatoTremor:
@phansys And another error which I think is unrelated. My guess would it to be a hiccup on GitHubs end. Can you please re-run the failed jobs as I can't do that here. It ran successful local.

phansys:
The issues reported by PHPStan should be resolved by #2475.

Nirzol:
Hi ! Some news about this feature ? Hope will be soon in the latest version !! 👍  nice job
I tried @MisatoTremor Fork , do the job as perfect ! thanks !

LouWii:
Oh no! Is there any reason why this was never merged in?

LouWii:
@phansys If I were to fix this, I would to branch off this branch, fix and create my own PR, correct?

phansys:
>@phansys If I were to fix this, I would to branch off this branch, fix and create my own PR, correct?

Yes, that's correct.
Thank you in advance.

MisatoTremor:
Sorry for the delay, due to the age of this PR I lost track.

PR is now rebased

phansys:
Thank you @MisatoTremor!

PR #2174, opened 2020-10-18
Fix dir path in symfony4.md

geoffrey-brier:
Hi there,

I had some trouble when upgrading to v3, composer told me that the loggable extension wasn't loaded ... after digging a few minutes I realized that the dir path had changed so I updated the doc accordingly.

Hope this helps :v:


AkenRoberts:
Excellent, thank you! 💯

PR #2178, opened 2020-10-29
[Tree] Bug fix. If the parent set to null, the tree does not recalculate

roman-shustrov:
Bug fix. If the parent set to null, the tree does not recalculate


stephanvierkant (approved):

franmomu:
done in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2415

PR #2185, opened 2020-12-05
Fix wrong usage of Comparable interface & misleading docs

alexsegura:
Hi, 

When using a deeply nested object as the sortable group, it is still very common to have the following error: 

> Nesting level too deep - recursive dependency?

Actually, there is a wrong usage of the `Doctrine\Common\Comparable` interface introduced in 4017af65a0a5cca6c6646cda728c92e925cfdabe, & a misleading documentation. 

1/ The documentation suggests to implement the interface on the item, while in the code it is actually used **on the group**. 

2/ The error happens, because when the group is an object, did not exist before and **DOES NOT** implement the `Comparable` interface, the comparison is done with `==` ([at this line](https://github.com/Atlantic18/DoctrineExtensions/blob/2cd7277ae49ec8c2640b247e5af4e36c58e78f22/lib/Gedmo/Sortable/SortableListener.php#L470)). When objects are deeply nested, `==` will compare all properties of the object recursively. 

See the piece of code below with comments: 

```php
while ($matches && ($group = key($relocation['groups']))) {
    $gr = $meta->getReflectionProperty($group)->getValue($object);
    if (null === $value) {
        $matches = $gr === null;
    } elseif (is_object($gr) && is_object($value) && $gr !== $value) {
        if ($gr instanceof Comparable) {
            // First problem
            // compareTo should return 0 if objects are equal
            // but $matches is actually as boolean
            $matches = $gr->compareTo($value);
        } else {
            // Second problem
            // In this code block, we are sure both $gr & $value are objects
            // but we still use == comparison, instead of ===
            $matches = $gr == $value;
        }
    } else {
        $matches = $gr === $value;
    }
    $value = next($relocation['groups']);
}
if ($matches) {
    // ...
}
```

**TL;DR the only way to make the `Doctrine\Common\Comparable` work, is to return a boolean, not 0/1/-1**

---

The previous interface, `Gedmo\Sortable\Comparable` is undocumented, but from what I understand it was expecting a boolean, thus it was working. When the code was change to use `Doctrine\Common\Comparable`, it was supposed that both interfaces were the same, but they are not. Actually, we are looking for a [`isEquatable`](https://github.com/symfony/security-core/blob/030d56f6b70c4532ebdd8add29f54b8b8f10070f/User/EquatableInterface.php) interface.

### DIFF
--- a/doc/sortable.md
+++ b/doc/sortable.md
@@ -304,6 +304,8 @@ $item2->setPosition(-1);
 Sortable works by comparing objects in the same group to see how they should be positioned. From time to time you may want to customize the way these
 objects are compared by simply implementing the Doctrine\Common\Comparable interface
 
+**Warning** the `Comparable` interface should be implemented on the group object
+
 ``` php
 <?php
 namespace Entity;
@@ -311,17 +313,12 @@ namespace Entity;
 use Doctrine\Common\Comparable;
 
 /**
- * @ORM\Table(name="items")
- * @ORM\Entity(repositoryClass="Gedmo\Sortable\Entity\Repository\SortableRepository")
+ * @ORM\Table(name="category")
  */
-class Item implements Comparable
+class Category implements Comparable
 {
     public function compareTo($other)
     {
-        // return 1 if this object is considered greater than the compare value
-
-        // return -1 if this object is considered less than the compare value
-
         // return 0 if this object is considered equal to the compare value
     }
 }
--- a/lib/Gedmo/Sortable/SortableListener.php
+++ b/lib/Gedmo/Sortable/SortableListener.php
@@ -465,9 +465,9 @@ class SortableListener extends MappedEventSubscriber
                                 // If the object implements Comparable interface we can use its compareTo method
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
-                                    $matches = $gr->compareTo($value);
+                                    $matches = 0 === ($gr->compareTo($value));
                                 } else {
-                                    $matches = $gr == $value;
+                                    $matches = $gr === $value;
                                 }
                             } else {
                                 $matches = $gr === $value;


lib/Gedmo/Sortable/SortableListener.php
```diff
@@ -465,9 +465,9 @@ public function postFlush(EventArgs $args)
                                 // If the object implements Comparable interface we can use its compareTo method
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
-                                    $matches = $gr->compareTo($value);
+                                    $matches = 0 === ($gr->compareTo($value));
```
phansys, line 468:
```suggestion
                                    $matches = 0 === $gr->compareTo($value);
```

doc/sortable.md
```diff
@@ -304,24 +304,21 @@ $item2->setPosition(-1);
...
- * @ORM\Table(name="items")
- * @ORM\Entity(repositoryClass="Gedmo\Sortable\Entity\Repository\SortableRepository")
+ * @ORM\Table(name="category")
  */
-class Item implements Comparable
+class Category implements Comparable
 {
     public function compareTo($other)
     {
-        // return 1 if this object is considered greater than the compare value
```
phansys, line 321: Why we should remove these comments?
The description at [`Doctrine\Common\Comparable`](https://github.com/doctrine/common/blob/3f7b18fb66f2cf65cc7e170ad321c1272b643efc/lib/Doctrine/Common/Comparable.php#L15-L16) states the following:
>Returns 0 if they are semantically equal, 1 if the other object is less than the current one, or -1 if its more than the current one.

phansys (changes requested):

doc/sortable.md
alexsegura: Because at [this line](https://github.com/doctrine-extensions/DoctrineExtensions/blob/ea1e0253f5c148754a525865c7277f38396033e9/src/Sortable/SortableListener.php#L467), the comparison is done against a _truthy_ value, and `if (-1)` would evaluate to `true`. So if you return literally _anything else_ than 0, the objects would be considered as different. 

This is what I explained in the pull request initial comment, we don't want to compare objects, we want to guess if they are equal. 
> Actually, we are looking for a isEquatable interface.

phansys: Well, even if I guess the loose comparison against `-1` is a valid reason by now, please be aware that this kind of condition will be surely replaced by its strict variant in the next releases.
As I'm not using this extension, I thought the possible values returned by the method could be used to determine the position in the sorted result, using a similar logic than the provided by the `<=>` operator.

lib/Gedmo/Sortable/SortableListener.php
```diff
@@ -465,9 +465,9 @@ public function postFlush(EventArgs $args)
                                 // If the object implements Comparable interface we can use its compareTo method
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
-                                    $matches = $gr->compareTo($value);
+                                    $matches = 0 === $gr->compareTo($value);
                                 } else {
-                                    $matches = $gr == $value;
+                                    $matches = $gr === $value;
```
phansys, line 470: We want to check for identical values here?
Looking at a previous condition, I think we already know that these values are not identical:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/84b5768c37b989508ec06cc20c1d5ff554a10bc1/lib/Gedmo/Sortable/SortableListener.php#L463

PR #2201, opened 2021-02-10
Remove some obsolete checks

phansys:
Remove some obsolete checks related to old PHP versions which are not supported anymore from this package.

I guess these changes are pedantic and don't require a changelog.
Please, let me know if should I create one.


stephanvierkant (approved):
LGTM. I agree a changelog isn't required in this case.

stephanvierkant (approved):

PR #2206, opened 2021-03-05
SoftDeletable is not working when the entity is already in the identity map

phansys:
SoftDeletable is not working when the entity is already in the identity map.

I tried to check if calling `UnitOfWork::clear($softDeletedClass)` at `SoftDeleteableListener::onFlush()` helps, but I got no good results.

Is there something we can do to cover this case? Otherwise, I think it deserves a note in the docs.

See #1917.


phansys:
Closing based on the reasons explained at #1917.

PR #2208, opened 2021-03-17
[Tree] Restore tree repository traits and remove abstract tree repositories

jmontoyaa:
Originally it was added here:

https://github.com/doctrine-extensions/DoctrineExtensions/tree/archive/master-2019-03-17
https://github.com/doctrine-extensions/DoctrineExtensions/pull/1496


src/Tree/Entity/Repository/AbstractTreeRepository.php
```diff
@@ -2,245 +2,23 @@
 
 namespace Gedmo\Tree\Entity\Repository;
 
-use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityManager;
```
hacfi, line 5: not sure if this is desirable

PR #2212, opened 2021-04-02
Add phpdoc to object property

franmomu:
If not specified, when using static analysis, it infers `$object` as `mixed`. 

Another improvement could be to use generics defining this class as generics like:

```php
/**
 * @template T of object
 */
abstract class AbstractPersonalTranslation
{
    /**
     * @psalm-var T
     * 
     * Related entity with ManyToOne relation
     * must be mapped by user
     */
    protected $object;
// ...
```

So users can specify their type with something like:
```php
/**
 * @psalm-extends AbstractPersonalTranslation<PersonalTranslatable>
 */
class PersonalTranslation extends AbstractPersonalTranslation
{
    
}
```
If that's ok, I'll make the PR.

### DIFF
--- a/src/Translatable/Document/MappedSuperclass/AbstractPersonalTranslation.php
+++ b/src/Translatable/Document/MappedSuperclass/AbstractPersonalTranslation.php
@@ -26,8 +26,10 @@ abstract class AbstractPersonalTranslation
     protected $locale;
 
     /**
-     * Related entity with ManyToOne relation
+     * Related document with ManyToOne relation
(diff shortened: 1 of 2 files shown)


AkenRoberts:
Let's stick with the regular docblock. Can you move the `@var` to below the description/comments, please?

franmomu:
> Let's stick with the regular docblock. Can you move the `@var` to below the description/comments, please?

sure!

AkenRoberts:
Thank you ❤️

PR #2213, opened 2021-04-04
Remove ext-mongo references

franmomu:
Since there is a conflict with `"doctrine/mongodb-odm": "<2.0"`, I think this code should belong to the client. In another PR I'll remove references to non-existing classes of `doctrine/mongodb-odm >= 2`


phansys (approved):

phansys:
Thank you @franmomu!

PR #2214, opened 2021-04-04
Migrate to Github Actions

franmomu:
It has added a test for lowest dependencies, the result can be seen in: https://github.com/franmomu/DoctrineExtensions/pull/1


stephanvierkant (approved):
Looks good to me!

svycka (approved):
I would like a code style check added but other than this I guess all ok if you tested.

franmomu:
> I would like a code style check added but other than this I guess all ok if you tested.

I think that deserves another file (and PR because code style should be fixed first), this is just about migrating the existing travis configuration to Github Actions.

svycka:
> > I would like a code style check added but other than this I guess all ok if you tested.
> 
> I think that deserves another file (and PR because code style should be fixed first), this is just about migrating the existing travis configuration to Github Actions.

PR #2217, opened 2021-04-09
[Tree][MaterializedPath] Use path_separator when removing children

stephanvierkant:
Consider this tree (without `path_ends_with_separator: false`):
* A
* A-B
* A-B-A
* A-B-B
* A-BA
* A-BB

When removing A-B, not only A-B-A and A-B-B are removed, but A-BA and A-BB are being removed too, while they're not children. This PR uses the `path_separator` configuration option to prevent this.


AkenRoberts:
@stephanvierkant Thanks Stephan! Btw when you tag releases, please make sure there are changelog entries, and change `Gedmo\DoctrineExtensions::VERSION` to match 👍🏼

stephanvierkant:
Sorry, missed that! I'll fix it in the next release.

PR #2218, opened 2021-04-15
id property can be nullable

franmomu:
In case of MongoDB, the Id by default it's a string: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/basic-mapping.html#doctrine-mapping-types

`id: string to ObjectId by default, but other formats are possible`

I've added `null` to the identifiers since the current configuration is to autogenerate them, so instantiating (an object extending these abstract classes) them and calling `getId` will return `null`.


stephanvierkant (approved):

phansys (approved):

phansys:
Thank you @franmomu!

PR #2220, opened 2021-04-26
Uploadable - Support multiple files in a single entity

ajay-gupta:
Add support of multiple uploadable files in a single entity.

https://github.com/ajay-gupta/DoctrineExtensions/blob/main/doc/uploadable.md


doc/uploadable.md
```diff
@@ -1,5 +1,114 @@
 # Uploadable behavior extension for Doctrine 2
 
+## About this fork
```
phansys, line 3: This repository is not a fork.

src/Uploadable/FilenameGenerator/FilenameGeneratorAlphanumeric.php
```diff
@@ -17,7 +17,7 @@ class FilenameGeneratorAlphanumeric implements FilenameGeneratorInterface
     /**
      * {@inheritdoc}
      */
-    public static function generate($filename, $extension, $object = null)
+    public static function generate($filename, $extension, $object = null, $identifier = null)
```
phansys, line 20: Adding a argument to a non final method is considered a BC break.

franmomu:
I haven't checked the implementation yet, but looking at the docs I guess `Uploadables` is not necessary, `Uploadable` could be repeatable, that would work for attributes using php 8.0.

Tests are failing and the docs comment should be addressed.

PR #2222, opened 2021-04-30
Add Coding standards check for Github Actions

franmomu:
Triggered by https://github.com/doctrine-extensions/DoctrineExtensions/pull/2214#pullrequestreview-649006936, this adds a coding standards check job for Github Actions.

~It is draft because~ there are a couple of things to discuss, the job can be seen in https://github.com/franmomu/DoctrineExtensions/pull/4

Seeing the changes proposed by `php-cs-fixer`, just a couple of questions:

```diff
--- Original
+++ New
@@ @@
         if (isset($mapping->field)) {
-            /**
+            /*
              * @var \SimpleXmlElement
              */
```
Is this change ok? or should we change [`phpdoc_to_comment`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/rules/phpdoc/phpdoc_to_comment.rst)? to not remove this.

And the other one:

```diff
--- Original
+++ New
@@ @@
                 // update translation
-                if ($translated !== null
+                if (null !== $translated
```

I personally [don't like yoda conditions](https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not), I would disable them, but it's not my call.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,28 @@
+name: "Coding Standards"
+
+on:
+  pull_request: null
+
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-20.04"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "7.4"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v1"
+        with:
+          dependency-versions: "highest"
+
+      - name: "Run PHP-CS-Fixer"
+        run: "bin/php-cs-fixer fix --ansi --verbose --diff --dry-run"
+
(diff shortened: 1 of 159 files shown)


phansys (approved):

svycka:
```diff
--- Original
+++ New
@@ @@
         if (isset($mapping->field)) {
-            /**
+            /*
              * @var \SimpleXmlElement
              */
````
I vote to keep the old one `/**`
````diff
--- Original
+++ New
@@ @@
                 // update translation
-                if ($translated !== null
+                if (null !== $translated
````
I am ok with this change

stephanvierkant:
> I personally don't like yoda conditions, I would disable them, but it's not my call.
Hear hear! I really hate it. The only reason one want to use it is invalid since the invention of modern static analysers.

Let's use the Doctrine Coding Standard.

franmomu:
I would use Doctrine Coding Standard too, but it would require some more work, IMHO this could be merge as is for new PRs to stick to a CS and another PR could be done to change to Doctrine CS.

AkenRoberts:
Thank you!

I am anti-Yoda condition as well, but that can be dialed in another time. Consistency is more important than what I prefer as an individual.

And agreed on getting this started and possibly moving to Doctrine Coding Standard in the future.

PR #2226, opened 2021-05-15
Allow installing doctrine/cache 2.0

alcaeus:
This PR uses doctrine/cache to create a cached annotations reader. With doctrine/cache being sunset and the 2.0 release no longer containing implementations, we need a different alternative.

This PR updates the annotations library and will attempt to create a cached reader using either symfony/cache (if it is installed), or doctrine/cache (if the implementations are present, i.e. 1.x is installed). If neither is found, the reader will not cache its parsing result. This logic can be expanded to cover additional cache libraries if desired. Note that this is not performance critical, as annotation parsing is only done when building metadata, which should be cached anyways.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,8 +36,8 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "behat/transliterator": "~1.2",
-        "doctrine/annotations": "^1.2",
-        "doctrine/cache": "^1.0",
+        "doctrine/annotations": "^1.13",
+        "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0"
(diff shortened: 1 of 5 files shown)


franmomu (approved):
After this, I guess a custom cache can be created instead of using the one from `ClassMetadataFactory` (which actually is from `AbstractClassMetadataFactory` and deprecated).

https://github.com/doctrine-extensions/DoctrineExtensions/blob/f956c3c4d0c0ffdc5dd879288073772e439b6c1a/src/Mapping/MappedEventSubscriber.php#L123

composer.json
```diff
@@ -36,8 +36,8 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "behat/transliterator": "~1.2",
-        "doctrine/annotations": "^1.2",
-        "doctrine/cache": "^1.0",
+        "doctrine/annotations": "^1.13",
+        "doctrine/cache": "^1.11 || ^2.0",
```
derrabus, line 40: This dependency could be moved to `require-dev` now. This package is fully functional without `doctrine/cache`.

derrabus (approved):

composer.json
alcaeus: Changed, thanks for pointing this out!

alcaeus:
@stephanvierkant can you please take a look? This is one of the packages blocking the upgrade to doctrine/cache 2.0 for a number of people.

PR #2238, opened 2021-07-05
Update .gitattributes

drupol:
Remove obsolete files from dist.

Let's save space and trees by removing obsolete files that we do not need on dist (more info: https://php.watch/articles/composer-gitattributes)

Removing `docker-compose.yml` will also prevent automatic analysis tools to complain about running on docker with root user.


phansys (approved):

phansys:
Thank you @drupol!

drupol:
Thanks!

Issue #2235, opened 2021-06-21
PHP 8+: Support for attributes


makraz:
What are your plans for PHP 8+?
Is there any WIP about mapping attributes so we could use #[ORM\Column] instead of /** @var ORM\Column */?
If not, I'll like to work on this.

ndench:
I'd love to see this. I currently can't migrate to `#[ORM\Column]` because the extensions won't detect the properties anymore. Let me know if there's anything I can do to help.

fridde:
@makraz You might want to edit your title to "PHP 8+: Support for attributes" since I initially thought you were missing some special attribute when in fact you wish for attributes to be supported *at all*.

Good initiative!

makraz:
Thanks @fridde for the feedback

bpolaszek:
> I'd love to see this. I currently can't migrate to `#[ORM\Column]` because the extensions won't detect the properties anymore. Let me know if there's anything I can do to help.

```php
/**
 * @Gedmo\Timestampable(on="create")
 */
#[ORM\Column(type: Types::DATETIMETZ_IMMUTABLE)]
#[Groups([self::LIST, self::VIEW])]
public \DateTimeImmutable $createdAt;
```

Not elegant, but works on my side.

fridde:
@bpolaszek Woah, for some reason I didn't realize that we could mix annotations and attributes. I thought that Gedmo would use the system I declared in ```doctrine.yaml``` with the setting for doctrine.orm.mappings.App.type (which I set to "attribute").

Thanks for pointing out this workaround. This will be useful during the transition.

ndench:
That's a good workaround @bpolaszek. Unfortunately it doesn't work for blameable because the BlameableListener uses doctrine annotations to determine if the property is an association. 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/0bce65005d4430e1037df143eab47b929d127c84/src/Blameable/BlameableListener.php#L29

scroach:
> That's a good workaround @bpolaszek. Unfortunately it doesn't work for blameable because the BlameableListener uses doctrine annotations to determine if the property is an association.
> 
> https://github.com/doctrine-extensions/DoctrineExtensions/blob/0bce65005d4430e1037df143eab47b929d127c84/src/Blameable/BlameableListener.php#L29

Why shouldn't this work? The BlameableListener works on the ClassMetadata provided by doctrine so it should not matter how you mapped your fields.

[...]

PR #2241, opened 2021-07-16
feat: Add support to php 8 attributes

djpremier:
https://github.com/doctrine-extensions/DoctrineExtensions/issues/2235

### DIFF
--- a/src/Mapping/Annotation/Blameable.php
+++ b/src/Mapping/Annotation/Blameable.php
@@ -2,18 +2,22 @@
 
 namespace Gedmo\Mapping\Annotation;
 
+use Attribute;
 use Doctrine\Common\Annotations\Annotation;
+use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 
 /**
  * Blameable annotation for Blameable behavioral extension
  *
  * @Annotation
+ * @NamedArgumentConstructor()
  * @Target("PROPERTY")
  *
  * @author David Buchmann <mail@davidbu.ch>
  * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
  */
-final class Blameable extends Annotation
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class Blameable
 {
     /** @var string */
     public $on = 'update';
@@ -21,4 +25,19 @@ final class Blameable extends Annotation
     public $field;
     /** @var mixed */
     public $value;
+
+    /**
+     *
+     * @param string $on
+     * @param null|string|array $field
+     * @param mixed|null $value
+     *
+     * @return void
+     */
+    public function __construct($on = 'update', $field = null, $value = null)
+    {
+        $this->on = $on;
+        $this->field = $field;
+        $this->value = $value;
+    }
 }
--- a/src/ReferenceIntegrity/Mapping/Driver/Attribute.php
+++ b/src/ReferenceIntegrity/Mapping/Driver/Attribute.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Gedmo\ReferenceIntegrity\Mapping\Driver;
+
+use Doctrine\ORM\Mapping\Driver\AttributeReader;
+use Gedmo\Exception\InvalidMappingException;
+use Gedmo\Mapping\Driver\AbstractAnnotationDriver;
+use Gedmo\ReferenceIntegrity\Mapping\Validator;
+
+/**
+ * This is an attribute mapping driver for ReferenceIntegrity
+ * behavioral extension. Used for extraction of extended
+ * metadata from Attributes specifically for ReferenceIntegrity
+ * extension.
+ *
+ * @author Evert Harmeling <evert.harmeling@freshheads.com>
+ * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
+ */
+class Attribute extends AbstractAnnotationDriver
+{
+    /**
+     * Annotation to identify the fields which manages the reference integrity
+     */
+    const REFERENCE_INTEGRITY = 'Gedmo\\Mapping\\Annotation\\ReferenceIntegrity';
+
+    /**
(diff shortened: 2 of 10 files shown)


djpremier:
Helps are very much welcome 🙂

src/ReferenceIntegrity/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @author Evert Harmeling <evert.harmeling@freshheads.com>
+ * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
+ */
+class Attribute extends AbstractAnnotationDriver
+{
+    /**
+     * Annotation to identify the fields which manages the reference integrity
+     */
+    const REFERENCE_INTEGRITY = 'Gedmo\\Mapping\\Annotation\\ReferenceIntegrity';
```
ker0x, line 24:
```suggestion
    const REFERENCE_INTEGRITY = ReferenceIntegrity::class;
```

src/ReferenceIntegrity/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * ReferenceIntegrityAction extension annotation
+     */
+    const ACTION = 'Gedmo\\Mapping\\Annotation\\ReferenceIntegrityAction';
```
ker0x, line 29:
```suggestion
    const ACTION = ReferenceIntegrityAction::class;
```

src/Mapping/Annotation/Blameable.php
```diff
@@ -2,23 +2,42 @@
...
+
+    /**
+     *
+     * @param string $on
+     * @param null|string|array $field
+     * @param mixed|null $value
+     *
+     * @return void
+     */
+    public function __construct($on = 'update', $field = null, $value = null)
```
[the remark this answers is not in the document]
ndench: The other fields should probably be typed as well then? `?string`

src/ReferenceIntegrity/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function __construct()
+    {
+        $this->setAnnotationReader(new AttributeReader());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function readExtendedMetadata($meta, array &$config)
```
ndench, line 39: This appears to be copy pasted from `\Gedmo\ReferenceIntegrity\Mapping\Driver\Annotation`. Should it maybe extend the Annotation driver and overwrite the `AnnotationReader`? Or perhaps decorate it?

ndench (commented):
Thanks for you contribution! I think all of @ker0x suggestions are valid, we might as well type the properties since the min supported version is 7.2.

src/Mapping/Annotation/Blameable.php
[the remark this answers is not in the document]
ker0x: Unfortunatly no, according to the PHPDoc, they can have multiple types, and union types is only available in PHP 8


```php
/** @var string|array */
public $field;

/** @var mixed */
public $value;
```

AkenRoberts, line 37: This is a breaking change (removing the `Annotation` parent and changing the constructor signature). It would require a major release. Is it necessary, or can we find a backwards-compatible option?

franmomu: It's technically a break, but in terms of usage it should be fine because of the `@NamedArgumentConstructor` annotation: https://www.doctrine-project.org/projects/doctrine-annotations/en/1.13/custom.html#optional-constructors-with-named-parameters

So there is no need of extending `Annotation` class anymore.

mbabker: It's a bit of a weird case.  From a purely PHP API perspective, it is 100% a B/C break:

- Class inheritance changes (`$foo instanceof Doctrine\Common\Annotations\Annotation` goes from passing to failing with this PR applied)
- Constructor signature changes

From a practical perspective, it really starts to get murky.  `Gedmo\Mapping\Driver\AnnotationDriverInterface` doesn't have a hard dependency on the `Doctrine\Common\Annotations\Reader` interface, so there isn't an assurance that the `doctrine/annotations` package is even being used to read the annotations.  Even if there were, if someone is implementing the interface themselves and not using the readers or parser from that package (`Doctrine\Common\Annotations\DocParser` is where the logic to deal with the `@NamedArgumentConstructor` annotation lies), that reader might not be able to construct the annotation classes with the changed constructor signatures.  Granted, the odds of someone using a custom annotation reader and NOT using their parser are pretty low, but even if Doctrine's interfaces were typehinted, you can't guarantee a reader is using that compatibility layer.

franmomu: Thanks for the explanation, so what we can do is to keep the `array $data = []` as first parameter of the constructor and add the rest after it.

I've re-added it in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2253

art-cg:
Hi. @djpremier  Thanks for starting the PR. Are there any news of the actual state? 
I saw that 3.4.0 includes the first PHP8-Attributes
https://github.com/doctrine-extensions/DoctrineExtensions/releases/tag/v3.4.0

franmomu:
> Hi. @djpremier Thanks for starting the PR. Are there any news of the actual state? I saw that 3.4.0 includes the first PHP8-Attributes https://github.com/doctrine-extensions/DoctrineExtensions/releases/tag/v3.4.0

I've created https://github.com/doctrine-extensions/DoctrineExtensions/issues/2363 to keep track of the needed work.

franmomu:
Closing in favor of https://github.com/doctrine-extensions/DoctrineExtensions/issues/2363

PR #2244, opened 2021-08-27
Add missing XSD definitions

phansys:
Update XSD in order to add missing definitions:
* `tree-path` type;
* `hard-delete` and `time-aware` attributes to "soft-deleteable" type;

See
https://github.com/doctrine-extensions/DoctrineExtensions/blob/b26ed3b92091e5aa14d9fa1382d8c370084aa422/src/SoftDeleteable/Mapping/Driver/Xml.php#L47-L55

https://github.com/doctrine-extensions/DoctrineExtensions/blob/b26ed3b92091e5aa14d9fa1382d8c370084aa422/src/Tree/Mapping/Driver/Xml.php#L97


AkenRoberts:
Thank you 🥳

fri0z:
And when is it planned to update http://gediminasm.org/schemas/orm/doctrine-extensions-mapping ?

PR #2245, opened 2021-08-28
Update development dependencies

phansys:


mbabker:
> the "doctrine/cache:^2.0" is not resolvable

This package is compatible with `doctrine/cache:^2.0` though, is it not?  The entire point of #2226 was to ensure compatibility with the new major version.  IMO the constraint is right, even if upstream dependencies might block being able to use that version at this point in time (`doctrine/mongodb-odm:^2.3` does add compatibility for the new major BTW).

composer.json
```diff
@@ -46,15 +46,15 @@
...
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.9.6",
         "friendsofphp/php-cs-fixer": "^3.0",
-        "phpunit/phpunit": "^8.5",
-        "symfony/cache": "^4.4 || ^5.0",
-        "symfony/yaml": "^4.1"
+        "phpunit/phpunit": "^8.5 || ^9.5",
+        "symfony/cache": "^4.4 || ^5.3",
+        "symfony/yaml": "^4.1 || ^5.3"
```
mbabker, line 57:
```suggestion
        "symfony/yaml": "^4.4 || ^5.3"
```

Might as well make the Symfony component versions consistent, plus if bumping `symfony/cache` to not include EOM versions, `symfony/yaml` should have the same treatment applied.

mbabker (approved):

franmomu (approved):

franmomu:
Thanks @phansys!

Issue #1946, opened 2018-08-31
[Timestampable] Implement "Clock" provider layer


AkenRoberts:
The Timestampable extension should have a specific layer for providing the current date/time object. Commonly referred to as a Clock. This will help with testing primarily, along with adding another abstraction for customization.

A more detailed explanation: https://blog.frankdejonge.nl/being-in-control-of-time-in-php/

Example of a failing test due to passage of time: https://travis-ci.org/Atlantic18/DoctrineExtensions/jobs/422502612

Targeting the 3.0 refresh release.

AkenRoberts:
This is actually part of a broader issue/feature, where the "value provider" layer needs to be interfaced and customizable. For example, allowing users to specify a custom DateTime object for Timestampable (e.g. Chronos or Carbon).

juherr:
Any update about this issue?

I'm using php 7.4 and I had the following issue:
> PHP Fatal error:  Uncaught TypeError: Typed property Chargemap\Recurly\Domain\Subscription::$dateCreated must be an instance of Carbon\Carbon, DateTime used in /app/vendor/gedmo/doctrine-extensions/lib/Gedmo/AbstractTrackingListener.php:220

But `datetime` is overriden by my own CarbonType

```php
class CarbonType extends DateTimeType
{
    /**
     * @inheritdoc
     */
    public function convertToPHPValue($value, AbstractPlatform $platform)
    {
        if ($value === null) {
            return null;
        }
        return Carbon::instance(parent::convertToPHPValue($value, $platform));
    }

    /**
     * @inheritDoc
     */
    public static function override(): void
    {
        static::overrideType(Types::DATETIME_MUTABLE, get_called_class());
    }
}
```

Moving from 

```php
    /**
     * @Timestampable(on="create")
     * @Column(type="datetime",name="date_created")
     */
    private Carbon $dateCreated;
```
to

```php
    /**
     * @Timestampable(on="create")
     * @Column(type="datetime",name="date_created")
     */
    private \DateTime $dateCreated;
```

is resolving the issue.

AkenRoberts:
@juherr No updates on this unfortunately. I'm not sure when this will be dealt with. Grateful for any ideas or pull request contributions you may have to help.

FYI, extensions doesn't use DBAL's type system to generate the values, which is why your override didn't work. That's the goal for whenever this is updated, so we can relegate value generation to DBAL and not reinvent that wheel.

nesl247:
[...]

PR #2248, opened 2021-08-31
Convert DateTime to PHP value using Doctrine converter

k0ka:
Convert DateTime to PHP value using Doctrine converter. It allows using Timestampable with other date objects like Carbon.

Fixes https://github.com/doctrine-extensions/DoctrineExtensions/issues/1946

I guess the same should be done to `SoftDeleteableAdapter`. We will have to add the `nesbot/carbon` dev dependency to fully test it.


src/Timestampable/Mapping/Event/Adapter/ORM.php
```diff
@@ -20,6 +22,20 @@ final class ORM extends BaseAdapterORM implements TimestampableAdapter
...
+        $platform = $this->getObjectManager()->getConnection()->getDriver()->getDatabasePlatform();
+
+        return $converter->convertToPHPValue($this->getRawDateValue($mapping), $platform);
+    }
+
+    /**
+     * Generates current timestamp for the specified mapping
+     *
+     * @param object $mapping
+     * @return mixed
```
phansys, line 35:
```suggestion
     *
     * @return \DateTimeInterface|int
```

src/Timestampable/Mapping/Event/Adapter/ORM.php
```diff
@@ -20,6 +22,20 @@ final class ORM extends BaseAdapterORM implements TimestampableAdapter
...
+        $converter = Type::getType($mapping['type'] ?? Types::DATETIME_MUTABLE);
+        $platform = $this->getObjectManager()->getConnection()->getDriver()->getDatabasePlatform();
+
+        return $converter->convertToPHPValue($this->getRawDateValue($mapping), $platform);
+    }
+
+    /**
+     * Generates current timestamp for the specified mapping
+     *
+     * @param object $mapping
```
phansys, line 34:
```suggestion
     * @param array<string, mixed> $mapping
```

src/Timestampable/Mapping/Event/Adapter/ORM.php
```diff
@@ -20,6 +22,20 @@ final class ORM extends BaseAdapterORM implements TimestampableAdapter
...
+     * @return mixed
+     */
+    public function getRawDateValue($mapping)
```
phansys, line 37:
```suggestion
    private function getRawDateValue(array $mapping)
```

src/Timestampable/Mapping/Event/Adapter/ORM.php
```diff
@@ -30,4 +46,5 @@ public function getDateValue($meta, $field)
         return \DateTime::createFromFormat('U.u', number_format(microtime(true), 6, '.', ''))
             ->setTimeZone(new \DateTimeZone(date_default_timezone_get()));
     }
+
```
phansys, line 49:
```suggestion
```

phansys (changes requested):

k0ka:
Hi. 

1. Added Carbon support to SoftDeletable
2. Added Carbon as a dev dependency to test
3. Added tests for Timestampable/SoftDeletable
4. Added ChangeLog record

Please let me know If something needs changing

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ a release.
 - SoftDeleteable: Support to use annotations as attributes on PHP >= 8.0.
 - Blameable: Support to use annotations as attributes on PHP >= 8.0.
 - IpTraceable: Support to use annotations as attributes on PHP >= 8.0.
+- SoftDeleteable: Support for Carbon type attributes
```
phansys, line 26: I'm not a Carbon user. Are these changes only compatible with the features provided by this library or with any other custom implementation?

tests/Gedmo/Timestampable/CarbonTest.php
```diff
@@ -0,0 +1,163 @@
...
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        DoctrineType::overrideType(Types::DATE_MUTABLE, DateType::class);
+        DoctrineType::overrideType(Types::DATETIME_MUTABLE, \Doctrine\DBAL\Types\DateTimeType::class);
+    }
+
+    /**
+     * @test
```
phansys, line 61: Please, use "test" as method prefix instead of this annotation.

CHANGELOG.md
k0ka: Any custom implementation. But I guess Carbon is the most frequently used.

tests/Gedmo/Timestampable/CarbonTest.php
k0ka: done

CHANGELOG.md
phansys: Then, I think we should be clear about that.
Consider something like this:
>SoftDeleteable: Support for type attributes in custom `\DateTimeInterface` implementations (like [Carbon](https://carbon.nesbot.com)).

k0ka: I added a clarification. 
```Support for custom column types (like Carbon).```
These columns might be of any type, like someone could write a simple converter to store it as a string.

File: tests/Gedmo/SoftDeleteable/CarbonTest.php [hunk 7]
```diff
@@ -0,0 +1,107 @@
...
+use Doctrine\Common\EventManager;
+use Doctrine\DBAL\Types\Type as DoctrineType;
+use Doctrine\DBAL\Types\Types;
+use Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter;
+use Gedmo\SoftDeleteable\SoftDeleteableListener;
+use Gedmo\Tests\SoftDeleteable\Fixture\Entity\Article;
+use Gedmo\Tests\SoftDeleteable\Fixture\Entity\Comment;
+use Gedmo\Tests\Tool\BaseTestCaseORM;
+
+class CarbonTest extends BaseTestCaseORM
```
franmomu, line 23:
```suggestion
final class CarbonTest extends BaseTestCaseORM
```

tests/Gedmo/Timestampable/CarbonTest.php
```diff
@@ -0,0 +1,160 @@
...
+use Doctrine\DBAL\Types\Type as DoctrineType;
+use Doctrine\DBAL\Types\Types;
+use Gedmo\Tests\Timestampable\Fixture\Article;
+use Gedmo\Tests\Timestampable\Fixture\Author;
+use Gedmo\Tests\Timestampable\Fixture\Comment;
+use Gedmo\Tests\Timestampable\Fixture\Type;
+use Gedmo\Tests\Tool\BaseTestCaseORM;
+use Gedmo\Timestampable\TimestampableListener;
+
+class CarbonTest extends BaseTestCaseORM
```
franmomu, line 28:
```suggestion
final class CarbonTest extends BaseTestCaseORM
```

tests/Gedmo/Timestampable/CarbonTest.php
```diff
@@ -0,0 +1,160 @@
...
+        $this->em->flush();
+
+        static::assertSame($sport->getCreated(), $sc, 'Date created should remain same after update');
+        static::assertNotSame($sport->getUpdated(), $su2, 'Date updated should change after update');
+        static::assertSame($sport->getPublished(), $sp, 'Date published should remain the same after update');
+        static::assertNotSame($sport->getContentChanged(), $scc2, 'Content must have changed after update');
+        static::assertNotSame($sport->getAuthorChanged(), $sa, 'Author must have changed after update');
+    }
+
+    protected function getUsedEntityFixtures()
```
franmomu, line 152:
```suggestion
    protected function getUsedEntityFixtures(): array
```

tests/Gedmo/Timestampable/CarbonTest.php
```diff
@@ -0,0 +1,160 @@
...
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        DoctrineType::overrideType(Types::DATE_MUTABLE, DateType::class);
+        DoctrineType::overrideType(Types::DATETIME_MUTABLE, \Doctrine\DBAL\Types\DateTimeType::class);
+    }
+
+    public function testShouldHandleStandardBehavior()
```
franmomu, line 60:
```suggestion
    public function testShouldHandleStandardBehavior(): void
```

tests/Gedmo/SoftDeleteable/CarbonTest.php
```diff
@@ -0,0 +1,107 @@
...
+        static::assertIsObject($art);
+        static::assertIsObject($art->getDeletedAt());
+        static::assertInstanceOf(Carbon::class, $art->getDeletedAt());
+        $comment = $commentRepo->findOneBy([$commentField => $commentValue]);
+        static::assertIsObject($comment);
+        static::assertIsObject($comment->getDeletedAt());
+        static::assertInstanceOf(Carbon::class, $comment->getDeletedAt());
+    }
+
+    protected function getUsedEntityFixtures()
```
franmomu, line 100:
```suggestion
    protected function getUsedEntityFixtures(): array
```

phansys:
@olml89, could you please confirm if #2273 can be closed in favor of this PR?

tests/Gedmo/SoftDeleteable/CarbonTest.php
```diff
@@ -0,0 +1,107 @@
...
+use Doctrine\Common\EventManager;
+use Doctrine\DBAL\Types\Type as DoctrineType;
+use Doctrine\DBAL\Types\Types;
+use Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter;
+use Gedmo\SoftDeleteable\SoftDeleteableListener;
+use Gedmo\Tests\SoftDeleteable\Fixture\Entity\Article;
+use Gedmo\Tests\SoftDeleteable\Fixture\Entity\Comment;
+use Gedmo\Tests\Tool\BaseTestCaseORM;
+
+final class CarbonTest extends BaseTestCaseORM
```
phansys, line 23: IMO, these classes should not include "Carbon" in its names, but something like "CustomDateTime".

k0ka: Well, I can rename these files.

However they not only test custom field types but also check integration with Carbon converter.

`CustomDateTime` might be confusing too as `DateTime` is one of Doctrine types for physical time. Others are `Time` and `Date` (and all are postfixed by `_immutable` or `_mutable`). And custom converter should not be always of any time type. We can define a new custom type `my_timestamp` that will be stored in the database as a string or blob. And it will be handled by the same code.

So `CustomFieldType` would make more sense, but there is only one simple test there. Is it reasonable to discuss its name?

phansys: I think these are good concerns. The goal of my suggestion is to keep these files agnostic as possible in order to allow more tests in the future for custom implementations other than Carbon.
IMO, if the integration with the Carbon converter can be made separately, I think it could be a good decision, as
it is always better to keep the responsibilities separated.

k0ka: Uhm, I don't get it. What should I do?

phansys: >However they not only test custom field types but also check integration with Carbon converter.

>Uhm, I don't get it. What should I do?

From my perspective, the test should be organized in this way:
1. Any test related to the custom types should cover the responsibilities related to a custom type implementation (Carbon or others).
2. If we have a exclusive test covering the integration with the Carbon converter, I think this test should be provided in a separate class, as it is only related to the Carbon package.

Please, let me know if you need more context about my suggestion.

k0ka: There are only two tests (one for each extension). They are just copies of tests which check the main behavior with added field type assertions. 

In order to test this feature we need a Doctrine Field Converter. We may write a fake one by our own but we will have to maintain it and track when Doctrine change any requirements. Or we may use quite popular and maintained converter, the one provided by the Carbon team.

I've chosen the latter option.

k0ka:
I did a rebase and fixed errors. Should I do anything else?

franmomu (approved):

phansys (approved):

phansys:
Thanks @k0ka!

PR #2249, opened 2021-09-05
fix reserved indicator

julkwel:
When using the yaml config template this lead to : 
``` 
The reserved indicator "%" cannot start a plain scalar.

```

If apply, this commit will fix that.


phansys (approved):

franmomu (approved):

AkenRoberts:
Thank you ❤️

PR #2253, opened 2021-09-26
Add support for attributes to Translatable

franmomu:
I think it's easier if we do this by extension and once we have one ready we can do the same for the others.

 So what this PR does:

**Annotations**:

Annotations no longer extend from `Annotation` from Doctrine, it uses [`NamedArgumentConstructor` annotation](https://www.doctrine-project.org/projects/doctrine-annotations/en/1.13/custom.html#optional-constructors-with-named-parameters) to be compatible with PHP 8.

There is a new `Annotation` interface that is used as a marker to parse the attributes from this package.

**Attribute Driver**

There is a `AttributeReader` class which is [from Doctrine ORM](https://github.com/doctrine/orm/blob/c0f70204d1680d8d804cfcbf91df23668807cd74/lib/Doctrine/ORM/Mapping/Driver/AttributeReader.php), but removing the parts not used here. 

All these new classes are marked as internal since we would probably change them as we add attributes to more extensions.

**Tests**

In order to test attributes, when using PHP 8, it reads attributes instead of annotations.

### DIFF
--- a/doc/translatable.md
+++ b/doc/translatable.md
@@ -89,9 +89,15 @@ on how to setup and use the extensions in most optimized way.
 
 ### Translatable annotations:
 - **@Gedmo\Mapping\Annotation\Translatable** it will **translate** this field
-- **@Gedmo\Mapping\Annotation\TranslationEntity(class="my\class")** it will use this class to store **translations** generated
-- **@Gedmo\Mapping\Annotation\Locale or @Gedmo\Mapping\Annotation\Language** this will identify this column as **locale** or **language**
+- **@Gedmo\Mapping\Annotation\TranslationEntity(class="my\class")** it will use this class to store the generated **translations**
+- **@Gedmo\Mapping\Annotation\Locale** or **@Gedmo\Mapping\Annotation\Language** these will identify this column as **locale** or **language**
 used to override the global locale
+
+### Translatable attributes:
+- **Gedmo\Mapping\Annotation\Translatable** it will **translate** this field
+- **Gedmo\Mapping\Annotation\TranslationEntity(class: MyClass::class)** it will use this class to store the generated **translations**
+- **Gedmo\Mapping\Annotation\Locale** or **Gedmo\Mapping\Annotation\Language** these will identify this column as **locale** or **language**
+  used to override the global locale
 
 <a name="entity-domain-object"></a>
 
@@ -100,6 +106,8 @@ used to override the global locale
 **Note:** that Translatable interface is not necessary, except in cases where
 you need to identify an entity as being Translatable. The metadata is loaded only once when
 cache is activated
+
+### Annotations
 
 ``` php
 <?php
@@ -169,6 +177,72 @@ class Article implements Translatable
 }
 ```
 
+### Attributes
+
+``` php
+<?php
+namespace Entity;
+
+use Gedmo\Mapping\Annotation as Gedmo;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Translatable\Translatable;
+
+ #[ORM\Table(name: 'articles')]
+ #[ORM\Entity]
+class Article implements Translatable
+{
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: 'integer')]
+    private $id;
+
+    #[Gedmo\Translatable]
+    #[ORM\Column(name: 'title', type: 'string', length: 128)]
+    private $title;
+
+    #[Gedmo\Translatable]
+    #[ORM\Column(name: 'content', type: 'text')]
+    private $content;
+
+    /**
+     * Used locale to override Translation listener`s locale
+     * this is not a mapped field of entity metadata, just a simple property
+     */
+    #[Gedmo\Locale]
+    private $locale;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setTitle($title)
+    {
+        $this->title = $title;
+    }
+
+    public function getTitle()
+    {
+        return $this->title;
+    }
+
+    public function setContent($content)
+    {
+        $this->content = $content;
+    }
+
+    public function getContent()
+    {
+        return $this->content;
+    }
+
+    public function setTranslatableLocale($locale)
+    {
+        $this->locale = $locale;
+    }
+}
+```
+
 <a name="document-domain-object"></a>
 
 ## Translatable Document example:
@@ -193,12 +267,14 @@ class Article implements Translatable
      * @Gedmo\Translatable
      * @ODM\Field(type="string")
      */
+    #[Gedmo\Translatable]
     private $title;
 
     /**
      * @Gedmo\Translatable
      * @ODM\Field(type="string")
      */
+    #[Gedmo\Translatable]
     private $content;
 
     /**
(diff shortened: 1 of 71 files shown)


franmomu:
Restarting the build

franmomu:
Sorry for the noise of the force-push, I wanted to add a test that only uses attributes: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2253/files#diff-96cd9e9ed40adb46e75f652d1514f2e46dd7ba3d549d3c3159ff08c8a96df8e7

I think this is ready for review now

mbabker:
For the best B/C here, I'd say take a look at Symfony's [Route annotation](https://github.com/symfony/routing/blob/5.3/Annotation/Route.php) and how it deals with mapping things from the deprecated `$data` argument.  Otherwise, you end up with a situation where the new constructor parameters would overwrite any values set in the `$data` array.

```php
<?php

use Gedmo\Translatable\Translatable;

/**
 * @ORM\Entity
 * @ORM\Table(name="articles")
 */
class Article
{
    /**
     * @Gedmo\Translatable(fallback=true)
     * @ORM\Column(name="title", type="string", length=128)
     */
    public string $title;
}
```

With the `Translatable` annotation class as is, the fallback param will end up being null instead of the expected true (see https://3v4l.org/9pTaj).

tests/Gedmo/Mapping/Annotation/BaseClassAnnotationTestCase.php
```diff
@@ -0,0 +1,58 @@
...
+use Gedmo\Mapping\Annotation\Annotation;
+use PHPUnit\Framework\TestCase;
+
+abstract class BaseClassAnnotationTestCase extends TestCase
+{
+    /**
+     * @requires PHP 8
+     * @dataProvider getValidParameters
+     */
+    public function testLoadFromAttribute(string $annotationProperty, $expectedReturn)
```
phansys, line 15:
```suggestion
    public function testLoadFromAttribute(string $annotationProperty, $expectedReturn): void
```

tests/Gedmo/Mapping/Annotation/BaseClassAnnotationTestCase.php
```diff
@@ -0,0 +1,58 @@
...
+        if ($attributes) {
+            $attributes = $reflection->getAttributes($annotationClass);
+            $annotation = $attributes[0]->newInstance();
+        } else {
+            $reader = new AnnotationReader();
+            $annotation = $reader->getClassAnnotation($reflection, $annotationClass);
+        }
+
+        if (!$annotation instanceof $annotationClass) {
+            throw new \Exception('Can\'t parse annotation');
```
phansys, line 53:
```suggestion
            throw new \LogicException('Can\'t parse annotation.');
```

tests/Gedmo/Mapping/Annotation/BasePropertyAnnotationTestCase.php
```diff
@@ -0,0 +1,58 @@
...
+use Gedmo\Mapping\Annotation\Annotation;
+use PHPUnit\Framework\TestCase;
+
+abstract class BasePropertyAnnotationTestCase extends TestCase
+{
+    /**
+     * @requires PHP 8
+     * @dataProvider getValidParameters
+     */
+    public function testLoadFromAttribute(string $annotationProperty, string $classProperty, $expectedReturn)
```
phansys, line 15:
```suggestion
    public function testLoadFromAttribute(string $annotationProperty, string $classProperty, $expectedReturn): void
```

tests/Gedmo/Mapping/Annotation/TranslatablePropertyTestCase.php
```diff
@@ -0,0 +1,36 @@
...
+
+declare(strict_types=1);
+
+namespace Gedmo\Tests\Mapping\Annotation;
+
+use Gedmo\Mapping\Annotation\Translatable;
+use Gedmo\Tests\Mapping\Fixture\Annotation\TranslatableModel as AnnotationTranslatableModel;
+use Gedmo\Tests\Mapping\Fixture\Attribute\TranslatableModel as AttributeTranslatableModel;
+
+class TranslatablePropertyTestCase extends BasePropertyAnnotationTestCase
```
phansys, line 11: Is this class intended to be extended? Otherwise, I think we can mark it as final.

phansys (changes requested):

phansys (commented):
IMO, we should add a note in the changelog and update [translatable.md](https://github.com/doctrine-extensions/DoctrineExtensions/blob/9a4aec12ff890d8b88d998f9ea699c10a95aa468/doc/translatable.md).

src/Mapping/Annotation/Translatable.php
```diff
@@ -2,19 +2,35 @@
...
 {
-    /** @var bool */
+    /** @var bool|null */
     public $fallback;
+
+    public function __construct(array $data = [], ?bool $fallback = null)
+    {
+        if ([] !== $data) {
+            trigger_error(sprintf(
+                'Passing an array as first argument to "%s" is deprecated. Use named arguments instead.',
```
phansys, line 29:
```suggestion
                'Passing an array as first argument to "%s()" is deprecated. Use named arguments instead.',
```

src/Mapping/Annotation/TranslationEntity.php
```diff
@@ -2,19 +2,35 @@
...
+final class TranslationEntity implements GedmoAnnotation
 {
     /** @var string @Required */
     public $class;
+
+    public function __construct(array $data = [], string $class = '')
+    {
+        if ([] !== $data) {
+            trigger_error(sprintf(
+                'Passing an array as first argument to "%s" is deprecated. Use named arguments instead.',
```
phansys, line 29:
```suggestion
                'Passing an array as first argument to "%s()" is deprecated. Use named arguments instead.',
```

phansys (approved):

doc/translatable.md
```diff
@@ -93,6 +93,12 @@ on how to setup and use the extensions in most optimized way.
 - **@Gedmo\Mapping\Annotation\Locale or @Gedmo\Mapping\Annotation\Language** this will identify this column as **locale** or **language**
 used to override the global locale
 
+### Translatable attributes:
+- **Gedmo\Mapping\Annotation\Translatable** it will **translate** this field
+- **Gedmo\Mapping\Annotation\TranslationEntity(class: MyClass::class)** it will use this class to store **translations** generated
```
phansys, line 98:
```suggestion
- **Gedmo\Mapping\Annotation\TranslationEntity(class: MyClass::class)** it will use this class to store generated **translations**
```

doc/translatable.md
```diff
@@ -93,6 +93,12 @@ on how to setup and use the extensions in most optimized way.
...
+- **Gedmo\Mapping\Annotation\Locale or Gedmo\Mapping\Annotation\Language** this will identify this column as **locale** or **language**
```
phansys, line 99:
```suggestion
- **Gedmo\Mapping\Annotation\Locale** or **Gedmo\Mapping\Annotation\Language** this will identify this column as **locale** or **language**
```

doc/translatable.md
```diff
@@ -619,19 +696,26 @@ your translations by extending the mapped superclass.
 
 ArticleTranslation Entity:
 
+**Note:** that this example is using annotations and attributes for mapping, you should use
```
phansys, line 699:
```suggestion
**Note:** this example is using annotations and attributes for mapping, you should use
```

phansys (approved):
Great work @franmomu.

phansys:
Thank you @franmomu!

PR #2250, opened 2021-09-14
Replace removed constants from `Doctrine\DBAL\Types\Type`

phansys:
* Fixed accessing undefined constants from `Doctrine\DBAL\Types\Type`.
* Added missing dev dependency for "doctrine/dbal".

See [this failed action](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2246/checks?check_run_id=3576379821#step:6:22) in #2246.

PR #2255, opened 2021-10-05
Declare conflict against "doctrine/orm" >= 2.10 in order to guarantee the schema extension

phansys:
"doctrine/orm" 2.10.0 removes the possibility to extend the mapping schema, so we must provide our own validation schema in order to support this version.
See doctrine/orm#8852.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ a release.
 ### Fixed
 - Removed legacy checks targeting older versions of PHP (#2201)
 - Added missing XSD definitions (#2244)
+- Add conflict against "doctrine/orm" >=2.10 in order to guarantee the schema extension (see https://github.com/doctrine/orm/pull/8852) (#2255)
 
 ## [3.1.0] - 2021-06-22
 ### Fixed
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
     "conflict": {
         "doctrine/mongodb": "<1.3",
         "doctrine/mongodb-odm": "<2.0",
+        "doctrine/orm": ">=2.10",
         "sebastian/comparator": "<2.0"
     },
     "suggest": {


AkenRoberts:
Aha, yeah that'll do it. 😓

Any idea what the developer experience will be for anyone updating their Composer dependencies? Are they going to get a straight conflict and unable to upgrade Doctrine ORM/DBAL? How soon can we get compatible?

phansys:
IIUC, the approach suggested by "doctrine/orm" is to declare the full validation schema in our side. IMO, we should discuss about how to implement the "pull" mechanism from the base schema, in order to avoid outdated definitions.

mbabker:
> Any idea what the developer experience will be for anyone updating their Composer dependencies? Are they going to get a straight conflict and unable to upgrade Doctrine ORM/DBAL?

Every consumer of this package will be required to use ORM 2.9 and earlier.  Anyone trying to add this package after being on ORM 2.10 (which potentially also includes DBAL 3.x) might run into conflicts trying to downgrade their dependencies.

phansys:
>Any idea what the developer experience will be for anyone updating their Composer dependencies? Are they going to get a straight conflict and unable to upgrade Doctrine ORM/DBAL? 

I think the current situation is worst for users using "doctrine/orm" 2.10, since the features provided by this package are not working at all (as you can see in some [recent checks](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2250/checks?check_run_id=3798204735)).

AkenRoberts:
Right. Not ideal, but necessary for now.

Is there anything else in flight that's critical or very useful for a next release? Worth getting this out quickly.

phansys:
I think we should merge #2250 and #2246 before the next release.
I'll do the corresponding rebase when this PR is merged.
Thank you,

AkenRoberts:
Sounds reasonable. Thank you so much for your contributions lately, Javier.

Just give me a second to get the changelog updated after merging before you rebase.

phansys:
>Just give me a second to get the changelog updated after merging before you rebase.

I can add the changelog on my side for this PR if you agree.

AkenRoberts:
Please do, makes my life easier 😄

AkenRoberts:
🙌🏼

PR #2257, opened 2021-10-05
Add annotation for createLifecycleEventArgsInstance method

mbabker:
The `SoftDeleteableListener` makes calls to `Gedmo\Mapping\Event\AdapterInterface::createLifecycleEventArgsInstance()`, but the method doesn't actually exist on the interface (the two concrete implementations in this package do provide it though).  This PR annotates the method as being soft-required, and at 4.0, the method should be required on the interface.

### DIFF
--- a/src/Mapping/Event/AdapterInterface.php
+++ b/src/Mapping/Event/AdapterInterface.php
@@ -4,6 +4,8 @@ namespace Gedmo\Mapping\Event;
 
 use Doctrine\Common\EventArgs;
 use Doctrine\ORM\UnitOfWork;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
+use Doctrine\Persistence\ObjectManager;
 
 /**
  * Doctrine event adapter interface is used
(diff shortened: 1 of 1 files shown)


src/Mapping/Event/AdapterInterface.php
```diff
@@ -4,6 +4,8 @@
 
 use Doctrine\Common\EventArgs;
 use Doctrine\ORM\UnitOfWork;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
```
phansys, line 7: IMO, we should declare "doctrine/persistence" as requirement before using these classes.

mbabker: It's covered in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2264 and https://github.com/doctrine-extensions/DoctrineExtensions/pull/2260, no need to add it on a third PR 😅

phansys (approved):

phansys:
Thank you @mbabker!

PR #2258, opened 2021-10-05
Deprecate the annotation reader being allowed to be any object

mbabker:
The annotation reader references in this package look to predate the `Doctrine\Common\Annotations\Reader` interface, and right now any place accepting a reader essentially allows any arbitrary object that mostly implements that interface (only the class and property readers are required, the method readers aren't).  This proposes tightening up the API by documenting that a `Doctrine\Common\Annotations\Reader` should be provided and deprecating support for any arbitrary object with 4.0 changing to require the interface.

### DIFF
--- a/src/Mapping/Driver/AbstractAnnotationDriver.php
+++ b/src/Mapping/Driver/AbstractAnnotationDriver.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Mapping\Driver;
 
+use Doctrine\Common\Annotations\Reader;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 
@@ -23,7 +24,9 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
     /**
      * Annotation reader instance
      *
-     * @var object
+     * @var Reader|AttributeReader|object
+     *
+     * @todo Remove the support for the `object` type in the next major release.
      */
     protected $reader;
 
@@ -43,6 +46,20 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
 
     public function setAnnotationReader($reader)
     {
+        if (!$reader instanceof Reader && !$reader instanceof AttributeReader) {
+            trigger_deprecation(
+                'gedmo/doctrine-extensions',
+                '3.11',
+                'Passing an object not implementing "%s" or "%s" as argument 1 to "%s()" is deprecated and'
+                .' will throw an "%s" error in version 4.0. Instance of "%s" given.',
+                Reader::class,
+                AttributeReader::class,
+                __METHOD__,
+                \TypeError::class,
+                get_class($reader)
+            );
+        }
+
         $this->reader = $reader;
     }
 
--- a/src/Mapping/Driver/AnnotationDriverInterface.php
+++ b/src/Mapping/Driver/AnnotationDriverInterface.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Mapping\Driver;
 
+use Doctrine\Common\Annotations\Reader;
 use Gedmo\Mapping\Driver;
 
 /**
@@ -20,17 +21,21 @@ use Gedmo\Mapping\Driver;
 interface AnnotationDriverInterface extends Driver
 {
     /**
-     * Set annotation reader class
-     * since older doctrine versions do not provide an interface
-     * it must provide these methods:
+     * Set the annotation reader instance
+     *
+     * When originally implemented, `Doctrine\Common\Annotations\Reader` was not available,
+     * therefore this method may accept any object implementing these methods from the interface:
+     *
      *     getClassAnnotations([reflectionClass])
      *     getClassAnnotation([reflectionClass], [name])
      *     getPropertyAnnotations([reflectionProperty])
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param object $reader annotation reader class
+     * @param Reader|AttributeReader|object $reader
      *
      * @return void
+     *
+     * @note Providing any object is deprecated, as of 4.0 a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` will be required
      */
     public function setAnnotationReader($reader);
 }
--- a/src/Mapping/ExtensionMetadataFactory.php
+++ b/src/Mapping/ExtensionMetadataFactory.php
@@ -10,6 +10,8 @@
 namespace Gedmo\Mapping;
 
 use Doctrine\Bundle\DoctrineBundle\Mapping\MappingDriver as DoctrineBundleMappingDriver;
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\Common\Cache\Cache;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\DefaultFileLocator;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
@@ -58,6 +60,7 @@ class ExtensionMetadataFactory
      * Custom annotation reader
      *
      * @var object
+     * @phpstan-var Reader|AttributeReader
      */
     protected $annotationReader;
 
@@ -66,8 +69,22 @@ class ExtensionMetadataFactory
      */
     private $cacheItemPool;
 
+    /**
+     * @param Reader|AttributeReader|object $annotationReader
+     */
     public function __construct(ObjectManager $objectManager, string $extensionNamespace, object $annotationReader, ?CacheItemPoolInterface $cacheItemPool = null)
     {
+        if (!$annotationReader instanceof Reader && !$annotationReader instanceof AttributeReader) {
+            trigger_deprecation(
+                'gedmo/doctrine-extensions',
+                '3.11',
+                'Providing an annotation reader which does not implement %s or is not an instance of %s to %s is deprecated.',
+                Reader::class,
+                AttributeReader::class,
+                static::class
+            );
+        }
+
         $this->objectManager = $objectManager;
         $this->annotationReader = $annotationReader;
         $this->extensionNamespace = $extensionNamespace;
--- a/src/Mapping/MappedEventSubscriber.php
+++ b/src/Mapping/MappedEventSubscriber.php
@@ -24,6 +24,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
+use Gedmo\Mapping\Driver\AttributeReader;
 use Gedmo\Mapping\Event\AdapterInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
@@ -76,7 +77,7 @@ abstract class MappedEventSubscriber implements EventSubscriber
     /**
      * Custom annotation reader
      *
-     * @var object
+     * @var Reader|AttributeReader|object|null
      */
     private $annotationReader;
 
@@ -162,20 +163,36 @@ abstract class MappedEventSubscriber implements EventSubscriber
     }
 
     /**
-     * Set annotation reader class
-     * since older doctrine versions do not provide an interface
-     * it must provide these methods:
+     * Set the annotation reader instance
+     *
+     * When originally implemented, `Doctrine\Common\Annotations\Reader` was not available,
+     * therefore this method may accept any object implementing these methods from the interface:
+     *
      *     getClassAnnotations([reflectionClass])
      *     getClassAnnotation([reflectionClass], [name])
      *     getPropertyAnnotations([reflectionProperty])
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param Reader $reader annotation reader class
+     * @param object $reader
+     * @phpstan-param Reader|AttributeReader $reader
      *
      * @return void
+     *
+     * @note Providing any object is deprecated, as of 4.0 a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` will be required
      */
     public function setAnnotationReader($reader)
     {
+        if (!$reader instanceof Reader && !$reader instanceof AttributeReader) {
+            trigger_deprecation(
+                'gedmo/doctrine-extensions',
+                '3.11',
+                'Providing an annotation reader which does not implement %s or is not an instance of %s to %s() is deprecated.',
+                Reader::class,
+                AttributeReader::class,
+                __METHOD__
+            );
+        }
+
         $this->annotationReader = $reader;
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ a release.
 - In order to close the API, `@final` and `@internal` annotations were added to all non base classes, which means that extending
   these classes is deprecated and can not be inherited in version 4.0.
 - Sortable: Accepting a return type other than "integer" from `Comparable::compareTo()` is deprecated in `SortableListener::postFlush()`. This will not be accepted in version 4.0.
+- Deprecate the annotation reader being allowed to be any object.
+  In 4.0, a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` instance will be required.
 
 ## [3.10.0] - 2022-11-14
 ### Changed
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,8 @@
         "doctrine/event-manager": "^1.2",
         "doctrine/persistence": "^2.2 || ^3.0",
         "psr/cache": "^1 || ^2 || ^3",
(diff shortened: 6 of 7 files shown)


src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -37,6 +38,10 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
      */
     public function setAnnotationReader($reader)
     {
+        if (!($reader instanceof Reader)) {
+            trigger_deprecation('gedmo/doctrine-extensions', '3.3', 'Providing an annotation reader which does not implement %s is deprecated, as of 4.0 the reader must implement the interface.', Reader::class);
```
mbabker, line 42: I don't know what the opinion of those maintaining or using this library is regarding runtime deprecations, but IMO this is a case where they're needed.

Right now, static analysis isn't great with this library and a combination of static analysis and integration tests wouldn't actually catch the deprecated API usage without fully removing the `object` typehints in the doc blocks and just typehinting the `Doctrine\Common\Annotations\Reader` interface.  So, the added runtime deprecations can help to flag the behavior this PR proposes to deprecate while also giving a meaningful explanation on how to mitigate it.

I went for Symfony's [deprecation contracts](https://github.com/symfony/deprecation-contracts) for the runtime usage as it's what I know personally.  There's also Doctrine's [deprecations package](https://github.com/doctrine/deprecations) which they've been using.

AkenRoberts: I like this 👍🏼

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -16,7 +17,7 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
     /**
      * Annotation reader instance
      *
-     * @var object
+     * @var Reader|object
```
phansys, line 20:
```suggestion
     * @var Reader|object
     *
     * @todo Remove the support for the `object` type in the next major release.
```

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -37,6 +38,10 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
      */
     public function setAnnotationReader($reader)
     {
+        if (!($reader instanceof Reader)) {
```
phansys, line 41:
```suggestion
        if (!$reader instanceof Reader) {
```

phansys (commented):
LGTM.
IMO, we need to add an entry in the changelog about the introduced deprecations.

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -47,18 +48,22 @@ class ExtensionMetadataFactory
...
      * Initializes extension driver
      *
-     * @param string $extensionNamespace
-     * @param object $annotationReader
+     * @param string        $extensionNamespace
+     * @param Reader|object $annotationReader
      */
     public function __construct(ObjectManager $objectManager, $extensionNamespace, $annotationReader)
     {
+        if (!($annotationReader instanceof Reader)) {
```
phansys, line 63:
```suggestion
        if (!$annotationReader instanceof Reader) {
```

src/Mapping/MappedEventSubscriber.php
```diff
@@ -170,18 +170,26 @@ public function getExtensionMetadataFactory(ObjectManager $objectManager)
...
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param Reader $reader - annotation reader class
+     * @param Reader|object $reader
+     *
+     * @note Providing any object is deprecated, as of 4.0 a `Doctrine\Common\Annotations\Reader` will be required
      */
     public function setAnnotationReader($reader)
     {
+        if (!($reader instanceof Reader)) {
```
phansys, line 189:
```suggestion
        if (!$reader instanceof Reader) {
```

phansys (approved):

franmomu:
The only _problem_ I see here is that [`AttributeReader`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Mapping/Driver/AttributeReader.php) must implement `Reader` interface as well, it's not ideal since it would depend on `doctrine/annotations` but I guess we'll be supporting annotations and attributes for a while.

mbabker:
> The only _problem_ I see here is that [`AttributeReader`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Mapping/Driver/AttributeReader.php) must implement `Reader` interface as well, it's not ideal since it would depend on `doctrine/annotations` but I guess we'll be supporting annotations and attributes for a while.

I didn't look at the attributes implementation, but it almost feels like there's either a leaky abstraction or an incomplete abstraction layer if the attribute reader has to have the same interface as the annotation reader and inherently be coupled to a doc block annotation parsing package.

franmomu:
> I didn't look at the attributes implementation, but it almost feels like there's either a leaky abstraction or an incomplete abstraction layer if the attribute reader has to have the same interface as the annotation reader and inherently be coupled to a doc block annotation parsing package.

yep that was done on purpose to be able to reuse the [`Annotation`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/cbf538e20fbf95dc674b2149d4c05d04d6d877ab/src/Translatable/Mapping/Driver/Annotation.php) driver. I think I used the same approach as in `doctrine/orm` where the [`$reader` is duck-typed](https://github.com/doctrine/persistence/blob/4ce4712e6dc84a156176a0fbbb11954a25c93103/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php#L79).

We could maybe introduce our `Reader` interface (with adapters and removing the `getMethodAnnotation/s` methods) or allow `AttributeReader` as well.

stof:
No need to duck type. We can make the new type a union type (using 2 type checks before the deprecation)

mbabker:
> We can make the new type a union type (using 2 type checks before the deprecation)

Yeah, that could work.  I'll play with that at some point this week.

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -16,7 +17,9 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
     /**
      * Annotation reader instance
      *
-     * @var object
+     * @var Reader|AttributeReader|object
```
phansys, line 20: `Reader` and `AttributeReader` are subtypes of `object`, so I think we should leave a `@todo` comment explaining what we need to do in the next major release instead of adding these types here.

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -37,6 +40,10 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
      */
     public function setAnnotationReader($reader)
     {
+        if (!$reader instanceof Reader && !$reader instanceof AttributeReader) {
+            trigger_deprecation('gedmo/doctrine-extensions', '3.4', 'Providing an annotation reader which does not implement %s or is an instance of %s is deprecated, as of 4.0 the reader must implement the interface.', Reader::class, AttributeReader::class);
```
phansys, line 44:
```suggestion
            trigger_deprecation('gedmo/doctrine-extensions', '3.4', 'Argument 1 passed to %s() must be an instance of %s or %s. Passing a different type is deprecated and will not be allowed in version 4.0.', __METHOD__, Reader::class, AttributeReader::class);
```

src/Mapping/Driver/AnnotationDriverInterface.php
```diff
@@ -14,15 +15,19 @@
...
+     *
      *     getClassAnnotations([reflectionClass])
      *     getClassAnnotation([reflectionClass], [name])
      *     getPropertyAnnotations([reflectionProperty])
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param object $reader annotation reader class
+     * @param Reader|AttributeReader|object $reader
+     *
+     * @note Providing any object is deprecated, as of 4.0 a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` will be required
```
phansys, line 30:
```suggestion
     * @param object $reader Providing an object different than `Doctrine\Common\Annotations\Reader` or
     *                       `Gedmo\Mapping\Driver\AttributeReader` is deprecated since version 3.4 and
     *                       will not be allowed in 4.0.
```

src/Mapping/Driver/AbstractAnnotationDriver.php
mbabker: If we're going to go that route then it should be `@phpstan-param` annotated.  The entire point of adding the specific subtypes was to make clear the type of object expected.

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -24,6 +25,9 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
      * Annotation reader instance
      *
      * @var object
+     * @phpstan-var Reader|AttributeReader
```
phansys, line 28: There is no special case here for `@phpstan-var`. Conventional tags can be used:
```suggestion
     * @var Reader|AttributeReader
```

src/Mapping/Driver/AnnotationDriverInterface.php
```diff
@@ -20,17 +21,22 @@
...
+     * therefore this method may accept any object implementing these methods from the interface:
+     *
      *     getClassAnnotations([reflectionClass])
      *     getClassAnnotation([reflectionClass], [name])
      *     getPropertyAnnotations([reflectionProperty])
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param object $reader annotation reader class
+     * @param object $reader
+     * @phpstan-param Reader|AttributeReader $reader
```
phansys, line 35:
```suggestion
     * @param Reader|AttributeReader $reader
```

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -43,6 +47,10 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
 
     public function setAnnotationReader($reader)
     {
+        if (!$reader instanceof Reader && !$reader instanceof AttributeReader) {
+            trigger_deprecation('gedmo/doctrine-extensions', '3.11', 'Argument 1 passed to %s() must be an instance of %s or %s. Passing a different type is deprecated and will not be allowed in version 4.0.', __METHOD__, Reader::class, AttributeReader::class);
```
phansys, line 51:
```suggestion
            trigger_deprecation(
                'gedmo/doctrine-extensions',
                '3.11',
                'Argument 1 passed to "%s()" must be an instance of "%s" or "%s". Passing a different type is deprecated'
                .' and will not be allowed in version 4.0.',
                __METHOD__,
                Reader::class,
                AttributeReader::class
            );
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -66,8 +69,15 @@ class ExtensionMetadataFactory
      */
     private $cacheItemPool;
 
+    /**
+     * @phpstan-param Reader|AttributeReader $annotationReader
```
phansys, line 73:
```suggestion
     * @param Reader|AttributeReader $annotationReader
```

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -23,7 +24,9 @@ abstract class AbstractAnnotationDriver implements AnnotationDriverInterface
     /**
      * Annotation reader instance
      *
-     * @var object
+     * @var Reader|AttributeReader
```
phansys, line 27:
```suggestion
     * @var Reader|AttributeReader|object
```

src/Mapping/Driver/AnnotationDriverInterface.php
```diff
@@ -20,17 +21,21 @@
...
+     * When originally implemented, `Doctrine\Common\Annotations\Reader` was not available,
+     * therefore this method may accept any object implementing these methods from the interface:
+     *
      *     getClassAnnotations([reflectionClass])
      *     getClassAnnotation([reflectionClass], [name])
      *     getPropertyAnnotations([reflectionProperty])
      *     getPropertyAnnotation([reflectionProperty], [name])
      *
-     * @param object $reader annotation reader class
+     * @param Reader|AttributeReader $reader
```
phansys, line 34:
```suggestion
     * @param Reader|AttributeReader|object $reader
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -66,8 +69,15 @@ class ExtensionMetadataFactory
      */
     private $cacheItemPool;
 
+    /**
+     * @param Reader|AttributeReader $annotationReader
```
phansys, line 73:
```suggestion
     * @param Reader|AttributeReader|object $annotationReader
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -66,8 +69,15 @@ class ExtensionMetadataFactory
...
+     */
     public function __construct(ObjectManager $objectManager, string $extensionNamespace, object $annotationReader, ?CacheItemPoolInterface $cacheItemPool = null)
     {
+        if (!$annotationReader instanceof Reader && !$annotationReader instanceof AttributeReader) {
+            trigger_deprecation('gedmo/doctrine-extensions', '3.11', 'Providing an annotation reader which does not implement %s or is an instance of %s is deprecated, as of 4.0 the reader must implement the interface.', Reader::class, AttributeReader::class);
```
phansys, line 78: Please, try to use a message that includes the reference to the received object type in order to help the users debug the problem.

src/Mapping/MappedEventSubscriber.php
```diff
@@ -76,7 +77,8 @@ abstract class MappedEventSubscriber implements EventSubscriber
     /**
      * Custom annotation reader
      *
-     * @var object
+     * @var object|null
+     * @phpstan-var Reader|AttributeReader|null
```
phansys, line 81:
```suggestion
     * @phpstan-var Reader|AttributeReader|object|null
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -66,8 +69,22 @@ class ExtensionMetadataFactory
...
+    /**
+     * @param Reader|AttributeReader|object $annotationReader
+     */
     public function __construct(ObjectManager $objectManager, string $extensionNamespace, object $annotationReader, ?CacheItemPoolInterface $cacheItemPool = null)
     {
+        if (!$annotationReader instanceof Reader && !$annotationReader instanceof AttributeReader) {
+            trigger_deprecation(
+                'gedmo/doctrine-extensions',
+                '3.11',
+                'Providing an annotation reader which does not implement %s or is not an instance of %s to %s is deprecated.',
```
phansys, line 81: We SHOULD tell the user the object type of `$annotationReader`.

phansys: Usually, the exception messages regarding unsupported types deliver a message that includes the involved type. Check [this example](https://3v4l.org/Afjhc).

mbabker: > Providing an annotation reader which does not implement Doctrine\Common\Annotations\Reader or is not an instance of Gedmo\Mapping\Driver\AttributeReader to Gedmo\Mapping\ExtensionMetadataFactory is deprecated.

This is already a pretty lengthy message when you factor in the FQCN's of the three things involved.  If I'm changing the message to include the type that was passed in, for me I'd probably then remove the part about what the requirements are to keep the message from being too verbose:

> Providing My\NonCompliant\AnnotationReader as an annotation reader to Gedmo\Mapping\ExtensionMetadataFactory is deprecated and will cause a TypeError in 4.0.

Now the message is concise, but it's not really giving much info on how to fix it.  So to convey every bit of information, you end up with something like this:

> Providing My\NonCompliant\AnnotationReader as an annotation reader to Gedmo\Mapping\ExtensionMetadataFactory is deprecated and will cause a TypeError in 4.0.  As of 4.0, the annotation reader must implement Doctrine\Common\Annotations\Reader or be an instance of Gedmo\Mapping\Driver\AttributeReader.

Great, now we have all the information in the deprecation message.  But, that's a lot of information to consume for my tastes.

Now, to be fair, PHP core itself does specify the type in the deprecation message (see https://3v4l.org/YZmap#v8.2.0), but, I don't think PHP core has too many 300-character deprecation messages like the long version above is and I'd make the same case for information overload there as well.

phansys: This isn't a blocker for me, just a suggestion. If you think there is no need to apply this, please omit the comment.

mbabker: I think I'm going to leave this alone for now, but I do think it's better to have explained why I like it as is instead of just ignoring the suggestion or accepting it without question.

phansys: Thank you! :+1:

mbabker:
The latest PHPStan fail is unrelated to this PR (something in https://github.com/doctrine/mongodb-odm/releases/tag/2.4.3 changed a SA assertation I guess?).

franmomu (approved):

phansys:
Thanks @mbabker!

PR #2259, opened 2021-10-06
Fix some PHPStorm findings

franmomu:
These are automatic fixes made by PHPStorm.

It would be nice to add a namespace for tests like `Gedmo\Tests`, using `::class` would help with the refactoring.


phansys (approved):

franmomu:
I've removed the commit that was using `::class`, otherwise will be quite painful rebase https://github.com/doctrine-extensions/DoctrineExtensions/pull/2280, I'll apply one that one is merged.

PR #2260, opened 2021-10-06
Remove code referencing non-existing classes

franmomu:
Follow-up of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2213

`NotifyPropertyChanged` was moved [in `doctrine/common` `2.2`](https://github.com/doctrine/common/releases/tag/v2.10.0).

And for mongodb we have a conflict with `"doctrine/mongodb-odm": "<2.0"`, so `Cursor`, `Proxy` and `ClassMetadataInfo` do not exist in version 2.

### DIFF
--- a/src/References/Mapping/Event/Adapter/ORM.php
+++ b/src/References/Mapping/Event/Adapter/ORM.php
@@ -3,13 +3,13 @@
 namespace Gedmo\References\Mapping\Event\Adapter;
 
 use Doctrine\ODM\MongoDB\DocumentManager as MongoDocumentManager;
-use Doctrine\ODM\MongoDB\Proxy\Proxy as MongoDBProxy;
 use Doctrine\ODM\PHPCR\DocumentManager as PhpcrDocumentManager;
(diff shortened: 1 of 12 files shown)


src/References/Mapping/Event/Adapter/ORM.php
```diff
@@ -32,7 +32,7 @@ public function getIdentifier($om, $object, $single = true)
 
         if ($om instanceof MongoDocumentManager) {
             $meta = $om->getClassMetadata(get_class($object));
-            if ($object instanceof MongoDBProxy) {
+            if ($object instanceof Proxy) {
```
mbabker, line 35: Per the [ODM upgrade guide](https://github.com/doctrine/mongodb-odm/blob/2.2.x/UPGRADE-2.0.md#proxy-objects), they don't use any of the Doctrine `Proxy` interfaces and instead you should check for `ProxyManager\Proxy\GhostObjectInterface`.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2263, opened 2021-10-17
Allow PHPUnit 9

mbabker:
The tests run fine on PHPUnit 9, it should be allowed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.9.6",
         "friendsofphp/php-cs-fixer": "^3.0",
-        "phpunit/phpunit": "^8.5",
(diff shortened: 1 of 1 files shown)


phansys (approved):

AkenRoberts:
`require-dev` is only installed when this package is at the root. What benefit does having both 8.x and 9.x give us? Typically it's just the one version your tests are run against.

mbabker:
PHPUnit 9 does not support PHP 7.2.  Therefore, you either use one PHPUnit version (which, to my knowledge, PHPUnit 8 will not be updated for PHP 8.1 compatibility) or you allow multiple PHPUnit versions and let Composer pick the one appropriate for your environment.

AkenRoberts:
I see. Let's use this for now, but a larger discussion about minimum versions and whatnot should happen soon.

AkenRoberts:
Thank you, btw!

PR #2265, opened 2021-10-17
Improvements to interface documentation

mbabker:
This is an effort to improve the inline documentation in the package, starting with interfaces and a couple other base classes.  The biggest changes include:

- Review for grammar
- Adds types in doc blocks where there weren't any, and makes existing types more specific where able

### DIFF
--- a/src/Loggable/Mapping/Event/LoggableAdapter.php
+++ b/src/Loggable/Mapping/Event/LoggableAdapter.php
@@ -2,11 +2,11 @@
 
 namespace Gedmo\Loggable\Mapping\Event;
 
+use Doctrine\Persistence\Mapping\ClassMetadata;
 use Gedmo\Mapping\Event\AdapterInterface;
 
 /**
- * Doctrine event adapter interface
- * for Loggable behavior
+ * Doctrine event adapter for the Loggable extension.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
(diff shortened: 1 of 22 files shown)


franmomu (approved):
LGTM 🚀 after the conflict resolution

src/Loggable/Mapping/Event/LoggableAdapter.php
```diff
@@ -2,36 +2,36 @@
...
  * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
  */
 interface LoggableAdapter extends AdapterInterface
 {
     /**
-     * Get default LogEntry class used to store the logs
+     * Get the default object class name used to store the log entries.
      *
-     * @return string
+     * @return class-string
```
phansys, line 19: IMO, in order to keep the "native" docblocks clean, we should use the `phpstan-` prefix for cases like "class-string", "array-key", generics, etc.
```suggestion
     * @return string
     *
     * @phpstan-return class-string
```

src/Mapping/Event/AdapterInterface.php
```diff
@@ -5,12 +5,11 @@
 use Doctrine\Common\EventArgs;
 use Doctrine\ORM\UnitOfWork;
```
phansys, line 6: IMO, if this class is not used, we should remove this import.

phansys (changes requested):

src/Loggable/Mapping/Event/LoggableAdapter.php
mbabker: Covered in the rebase

src/Mapping/Event/AdapterInterface.php
mbabker: It's used.

phansys (approved):

phansys:
Thank you @mbabker!

PR #2271, opened 2021-10-21
Allow ORM 2.10.2 and later

derrabus:
Follows #2255 and doctrine/orm#9136.

Can you please verify that reverting that change unblocked the upgrade to ORM 2.10 for you?

PR #2272, opened 2021-10-22
Allow to install `doctrine/orm` 2.10

### LINKED ISSUE #2261: Tree broken on ORM 2.10
Hi, 
This is probably related to #2256, i've had to downgrade to doctrine/orm 2.9 because i had this error : 
**  Root cannot be changed manually, change parent instead **

Downgrading made the tree component work again for me. 

Here is a stacktrace : 
image

This happened by just persisting a new  "@Gedmo\Tree(type="nested")" tagged entity. 

If someone has an idea it would be really appreciated ! 
Thanks !

franmomu:
For some context: https://github.com/doctrine/orm/issues/9141 and https://github.com/doctrine-extensions/DoctrineExtensions/pull/2271

Since version `doctrine/orm` 2.10, `UnitOfWork` relies on SPL object IDs, not hashes, so in order to be able to use newer versions we need to adapt to this change.

The problem here is that `UnitOfWork` from ODM still uses `spl_object_hash`, so this PR changes all `spl_object_hash` calls to `spl_object_id` and to make it work with both ORM and ODM, these methods have been changed:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/8c02cee09e3dd43799ec3b84b619b19982c47f43/src/Mapping/Event/AdapterInterface.php#L131-L149

To accept an object instead of the OID, that way how the OID is calculated is delegated to the implementation,

ORM:

https://github.com/franmomu/DoctrineExtensions/blob/c23fd27d3d4859b459aee12c11d481de483638bc/src/Mapping/Event/Adapter/ORM.php#L154-L165

ODM:

https://github.com/franmomu/DoctrineExtensions/blob/c23fd27d3d4859b459aee12c11d481de483638bc/src/Mapping/Event/Adapter/ODM.php#L157-L168

This is BC break, but since `doctrine/orm` did it, not sure how many people would be affected by this.

Fixes #2261.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Restored compatibility with doctrine/orm >= 2.10.2 (#2272).
+  Since doctrine/orm 2.10, `Doctrine\ORM\UnitOfWork` relies on SPL object IDs instead of hashes, thus we need to adapt our codebase in order to be compatible with this change.
+  As `Doctrine\ODM\MongoDB\UnitOfWork` from doctrine/mongodb-odm still uses `spl_object_hash()`, all `spl_object_hash()` calls were replaced by `spl_object_id()` to make it work with both ORM and ODM managers.
+
 ## [3.2.0] - 2021-10-05
 ### Added
 - PHP 8 Attributes for Doctrine ORM to entities & traits (#2251) 
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "doctrine/dbal": "^2.13",
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.0",
-        "doctrine/orm": "^2.9.6",
+        "doctrine/orm": "^2.10.2",
         "friendsofphp/php-cs-fixer": "^3.0",
         "phpunit/phpunit": "^8.5 || ^9.5",
         "symfony/cache": "^4.4 || ^5.0",
@@ -56,7 +56,7 @@
     "conflict": {
         "doctrine/mongodb": "<1.3",
         "doctrine/mongodb-odm": "<2.0",
-        "doctrine/orm": ">=2.10",
+        "doctrine/orm": "<2.10.2",
         "sebastian/comparator": "<2.0"
     },
     "suggest": {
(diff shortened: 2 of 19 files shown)


composer.json
```diff
@@ -59,7 +59,7 @@
     "conflict": {
         "doctrine/mongodb": "<1.3",
         "doctrine/mongodb-odm": "<2.0",
-        "doctrine/orm": ">=2.10",
+        "doctrine/orm": "<2.10.2",
```
derrabus, line 62: I'd suggest that the constraint in `require-dev` is adjusted as well:

```
"doctrine/orm": "^2.10.2",
```

phansys (approved):

AkenRoberts:
Please add a changelog entry and I'll get this merged! 👍🏼

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Add compatibility with doctrine/orm > 2.10 (#2272)
```
```suggestion
- Add compatibility with doctrine/orm >= 2.10 (#2272)
```

phansys (approved):

CHANGELOG.md
phansys: Friendly ping @franmomu.

franmomu: I thought I did it 😕 I'll be AFK today, if you want to change it or I can do it tomorrow

phansys (commented):
Given the described BC break, I think we should make a clear explanation about the situation in the changelog.
I'll try to create a release with the pending commits previous to this PR, in order to isolate this change in a exclusive MINOR release.

phansys:
I'd like to know if we could close #2037 based on the changes made here.

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Add compatibility with doctrine/orm >= 2.10 (#2272)
```
```suggestion
- Restored compatibility with doctrine/orm >= 2.10.2 (#2272).
  Since doctrine/orm 2.10, `Doctrine\ORM\UnitOfWork` relies on SPL object IDs instead of hashes, we need to adapt our codebase in order to be compatible with this change.
  As `Doctrine\ODM\MongoDB\UnitOfWork` from doctrine/mongodb-odm still uses `spl_object_hash()`, all `spl_object_hash()` calls were replaced by `spl_object_id()` to make it work with both ORM and ODM managers.
```

SimonVanacco:
This seems to fix #2261 for me, thanks !

phansys:
Thank you @franmomu!

PR #2264, opened 2021-10-17
Tweak Doctrine dependencies and add a DBAL 3 compat fix

mbabker:
`doctrine/persistence` wasn't declared in the dependencies, added it with versions including the `Doctrine\Persistence` namespace

There isn't a big reason for `doctrine/orm` to be pinned to the last 2.9 patch release that I'm aware of (the XML linting should always be done on a setup with the latest supported version installed, so closing down the range just for that step is a little too strict).  I've also added a conflict to older ORM versions since effectively whatever is in dev is the minimum supported version for an optional dependency.

A conflict to `doctrine/cache:<1.11` is added to make sure that only versions of that library with the 2.0 compatibility layer are usable.

DBAL 3 compatibility is improved with this PR, and generally, projects supporting both DBAL 2 and 3 are locking to 2.13 thanks to the forward compatibility layer.

There's no need to conflict against `doctrine/mongodb` anymore.  That library is abandoned anyway and this library realistically doesn't support it anymore to begin with (with https://github.com/doctrine-extensions/DoctrineExtensions/pull/2260 removing the last remnants of references to it).

PR #2277, opened 2021-11-05
Fix some deprecated calls

phansys:
See https://github.com/doctrine-extensions/DoctrineExtensions/pull/2264#issuecomment-961509644.

### DIFF
--- a/tests/Gedmo/Tree/MaterializedPathORMRepositoryTest.php
+++ b/tests/Gedmo/Tree/MaterializedPathORMRepositoryTest.php
@@ -346,7 +346,12 @@ class MaterializedPathORMRepositoryTest extends BaseTestCaseORM
         $this->em->persist($newNode);
         $this->em->flush();
 
-        $this->assertRegexp('/Food\-\d+,New\sNode\-\d+/', $newNode->getPath());
+        // @todo: Remove the condition and the `else` block when dropping support for "phpunit/phpunit" < 9.1.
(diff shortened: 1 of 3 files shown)


phansys:
@mbabker, could you please review these changes?

tests/Gedmo/Tree/MaterializedPathORMRepositoryTest.php
```diff
@@ -346,7 +347,12 @@ public function testIssue458()
         $this->em->persist($newNode);
         $this->em->flush();
 
-        $this->assertRegexp('/Food\-\d+,New\sNode\-\d+/', $newNode->getPath());
+        // @todo: Remove the condition and the `else` block when dropping support for "phpunit/phpunit" < 9.0.
+        if (version_compare(Version::id(), '9.0.0') >= 0) {
```
mbabker, line 351: I'd do `method_exists($this, 'assertMatchesRegularExpression')` here, it's generally more reliable than version checks (plus, the new method was actually introduced in PHPUnit 9.1, so if 9.0 were allowed to install then this would actually fail).

mbabker (approved):

tests/Gedmo/Tree/MaterializedPathORMRepositoryTest.php
phansys: Done.

franmomu (approved):

franmomu:
Thanks @phansys!

PR #2278, opened 2021-11-05
Perform SCA with PHPStan

phansys:
**TODO**:
- [x] Add an upgrade note about the introduced deprecations.
- [x] Check how to resolve `has invalid type Doctrine\ORM\Mapping\ClassMetadata` finding.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,10 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    level: 1
+    paths:
+        - src
+        - tests
+    bootstrapFiles:
+        - tests/bootstrap.php
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,3 +10,5 @@
 /.yamllint
 CONTRIBUTING.md export-ignore
 /docker-compose.yml export-ignore
+phpstan.neon.dist export-ignore
+phpstan-baseline.neon export-ignore
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -1,0 +1,34 @@
+name: Quality Assurance
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  phpstan:
+    name: PHPStan
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install PHP with extensions
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.0'
+          coverage: none
+          extensions: mongodb-1.9.0, zip
+          tools: composer:v2
+
+      - name: Install Composer dependencies (highest)
+        uses: ramsey/composer-install@v1
+        with:
+          dependency-versions: highest
+          composer-options: --prefer-dist --prefer-stable --no-interaction --no-progress
+
+      - name: PHPStan
+        run: bin/phpstan --memory-limit=1G analyse --error-format=github
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 bin
 vendor
 composer.lock
+phpstan.neon
 tests/.phpunit.result.cache
 tests/phpunit.xml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 ---
 
 ## [Unreleased]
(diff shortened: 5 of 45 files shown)


phansys:
I don't understand the error reported about `Doctrine\ORM\Mapping\ClassMetadata`, but if I have no success resolving it, I'll end up adding the exception to the baseline.

franmomu:
Locally I didn't see any error 😕 restarting the build just in case

phansys:
>Locally I didn't see any error confused restarting the build just in case

I am also unable to reproduce the problem locally.

franmomu:
I would ignore for now too, apparently it's a false positive, it will eventually go away I guess

phansys:
>I would ignore for now too, apparently it's a false positive, it will eventually go away I guess

Great. The errors were ignored.
Friendly ping @franmomu.

AkenRoberts:
I'll take a look tonight after my normal 9-5 👍🏼

franmomu (approved):

AkenRoberts:
I'm at a loss for this one, too. The issue isn't reproducible locally, and seems to be inconsistent in our Actions. 🤷🏼 

I'm okay ignoring these. Would it make more sense to add them to the baseline? Something to avoid seeing this:

```
 -- ---------------------------------------------------------------------------------------------------------------------------------------------------------
     Error
 -- ---------------------------------------------------------------------------------------------------------------------------------------------------------
     Ignored error pattern #^Instantiated class Doctrine\\ORM\\Mapping\\ClassMetadata not found\.# was not matched in reported errors.
```

phansys:
>I'm okay ignoring these. Would it make more sense to add them to the baseline? Something to avoid seeing this:

I ignored these errors in the `phpstan.neon.dist` in order to use a wider pattern around these messages. As these errors are not idempotent and are only reported in the scope of the Github Actions environment, I've generated a baseline in the same environment, but the results are changing between executions. Then, sometimes the errors are triggered and in other runs they aren't there.
With the current configuration, there are no errors reported in GH actions, but obviously, the ignored errors are reported as not matched in our local environments.
As this seems to be a problem that only happens in the GH environment, I think the only way to keep all the environments without errors is having a separate config for GH until the problem is solved.
If you search in the actions history for this PR, you'll see my unsuccessful attempts clearing the PHPStan cache: https://github.com/doctrine-extensions/DoctrineExtensions/actions?query=%22Perform+SCA%22.

phansys:
@AkenRoberts, I'm using an exclusive config for GH actions in the last commit. Just check the related actions between a re-run without changes in the codebase between them:
1. https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/1443094774/attempts/1
2. https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/1443094774/attempts/2

phpstan.neon.dist
```diff
@@ -0,0 +1,15 @@
...
+    paths:
+        - src
+        - tests
+    bootstrapFiles:
+        - tests/bootstrap.php
+    treatPhpDocTypesAsCertain: false
+    checkGenericClassInNonGenericObjectType: true
+    checkMissingIterableValueType: true
+    checkMissingVarTagTypehint: true
+    checkMissingTypehints: true
```
franmomu, line 15: I'm not sure if this makes sense without adding the rules: https://github.com/phpstan/phpstan-src/blob/master/conf/config.level6.neon#L10-L16

phansys: I didn't know about this. Is `checkMissingTypehints` the only parameter to remove by now?

franmomu: I'm not 100% sure, but looks like:
- `checkMissingVarTagTypehint`: is used [at level 2 adding a service](https://github.com/phpstan/phpstan-src/blob/09abbe8bd4b634380db569efd28b682b94807232/conf/config.level2.neon#L60-L64)
- `checkMissingTypehints` and `checkMissingIterableValueType` work with [the rules from level 6](https://github.com/phpstan/phpstan-src/blob/master/conf/config.level6.neon#L10-L16).

So these configurations are apparently used when loading the corresponding services/rules, I would remove it for now

phansys: Great! Seems that these parameters were also causing the described issue around `class Doctrine\\ORM\\Mapping\\ClassMetadata not found`.
Thank you!

franmomu: Really? 🤣 we can sleep at night now

PR #2280, opened 2021-11-05
Leverage PSR-4 autoloading by setting "autoload-dev"

phansys:
**TODO**:
- [x] Check if the NS under the `tests/` directory can be updated from `Gedmo\` to `Gedmo\Tests\`.


franmomu (approved):
🚀 it took me a while when I tried to add a test some time ago because of this

PR #2284, opened 2021-11-08
DBAL 3 compatibility, misc. replacements of deprecated API uses

mbabker:
This adds full DBAL 3 compatibility and does a minor API cleanup to avoid uses of deprecated symbols.

A CI build is also added to make sure there is at least one run on each major DBAL branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,12 @@ a release.
 
 ## [Unreleased]
 ### Added
+- Support for doctrine/dbal 3.x
 - Timestampable: Support to use annotations as attributes on PHP >= 8.0.
+
+### Changes
+- Dropped support for doctrine/dbal < 2.13.1
+- The third argument of `Gedmo\SoftDeleteable\Query\TreeWalker\Exec\MultiTableDeleteExecutor::__construct()` requires a `Doctrine\ORM\Mapping\ClassMetadata` instance.
 
 ## [3.3.1] - 2021-11-18
 ### Fixed
--- a/composer.json
+++ b/composer.json
@@ -40,17 +40,18 @@
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/persistence": "^1.3.4 || ^2.0"
+        "doctrine/persistence": "^1.3.3 || ^2.0"
     },
     "conflict": {
-        "doctrine/mongodb": "<1.3",
+        "doctrine/cache": "<1.11",
+        "doctrine/dbal": "<2.13.1 || ~3.0.0",
         "doctrine/mongodb-odm": "<2.0",
         "doctrine/orm": "<2.10.2",
         "sebastian/comparator": "<2.0"
     },
     "require-dev": {
         "doctrine/cache": "^1.11 || ^2.0",
-        "doctrine/dbal": "^2.13",
+        "doctrine/dbal": "^2.13.1 || ^3.1",
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.0",
         "doctrine/orm": "^2.10.2",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,11 +1,17 @@
 parameters:
-    ignoreErrors:
-        -
-            message: "#^Class Doctrine\\\\ODM\\\\PHPCR\\\\DocumentManager not found\\.$#"
-            count: 2
-            path: src/References/Mapping/Event/Adapter/ORM.php
+	ignoreErrors:
+		-
+			message: "#^Class Doctrine\\\\ODM\\\\PHPCR\\\\DocumentManager not found\\.$#"
+			count: 2
+			path: src/References/Mapping/Event/Adapter/ORM.php
 
-        -
-            message: "#^Access to an undefined property ProxyManager\\\\Proxy\\\\GhostObjectInterface\\:\\:\\$identifier\\.$#"
-            count: 1
-            path: src/Tool/Wrapper/MongoDocumentWrapper.php
+		-
+			message: "#^Access to an undefined property ProxyManager\\\\Proxy\\\\GhostObjectInterface\\:\\:\\$identifier\\.$#"
+			count: 1
+			path: src/Tool/Wrapper/MongoDocumentWrapper.php
+
+		-
+			message: "#^Class Doctrine\\\\DBAL\\\\Platforms\\\\PostgreSQLPlatform not found\\.$#"
+			count: 1
+			path: src/Translatable/Query/TreeWalker/TranslationWalker.php
+
--- a/src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
+++ b/src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
@@ -3,7 +3,7 @@
 namespace Gedmo\SoftDeleteable\Query\TreeWalker\Exec;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query\AST\Node;
 use Doctrine\ORM\Query\Exec\MultiTableDeleteExecutor as BaseMultiTableDeleteExecutor;
 
@@ -20,17 +20,19 @@ class MultiTableDeleteExecutor extends BaseMultiTableDeleteExecutor
     /**
      * {@inheritdoc}
      */
-    public function __construct(Node $AST, $sqlWalker, ClassMetadataInfo $meta, AbstractPlatform $platform, array $config)
+    public function __construct(Node $AST, $sqlWalker, ClassMetadata $meta, AbstractPlatform $platform, array $config)
     {
         parent::__construct($AST, $sqlWalker);
 
         $sqlStatements = $this->_sqlStatements;
 
+        $quoteStrategy = $sqlWalker->getEntityManager()->getConfiguration()->getQuoteStrategy();
+
         foreach ($sqlStatements as $index => $stmt) {
             $matches = [];
             preg_match('/DELETE FROM (\w+) .+/', $stmt, $matches);
 
-            if (isset($matches[1]) && $meta->getQuotedTableName($platform) === $matches[1]) {
+            if (isset($matches[1]) && $quoteStrategy->getTableName($meta, $platform) === $matches[1]) {
                 $sqlStatements[$index] = str_replace('DELETE FROM', 'UPDATE', $stmt);
                 $sqlStatements[$index] = str_replace(
                     'WHERE',
--- a/src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
+++ b/src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
@@ -2,6 +2,10 @@
 
 namespace Gedmo\SoftDeleteable\Query\TreeWalker;
 
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\QuoteStrategy;
 use Doctrine\ORM\Query\AST\DeleteClause;
 use Doctrine\ORM\Query\AST\DeleteStatement;
 use Doctrine\ORM\Query\Exec\SingleTableDeleteUpdateExecutor;
@@ -20,13 +24,49 @@ use Gedmo\SoftDeleteable\SoftDeleteableListener;
  */
 class SoftDeleteableWalker extends SqlWalker
 {
+    /**
+     * @var Connection
+     *
+     * @deprecated to be removed in 4.0, use the getConnection method instead
+     */
     protected $conn;
+
+    /**
+     * @var AbstractPlatform
+     *
+     * @deprecated to be removed in 4.0, fetch the platform from the connection instead
+     */
     protected $platform;
+
+    /**
+     * @var SoftDeleteableListener
+     */
     protected $listener;
+
+    /**
+     * @var array
+     */
     protected $configuration;
+
+    /**
+     * @deprecated to be removed in 4.0, unused
+     */
     protected $alias;
+
+    /**
+     * @var string
+     */
     protected $deletedAtField;
+
+    /**
+     * @var ClassMetadata
+     */
     protected $meta;
+
+    /**
+     * @var QuoteStrategy
+     */
+    private $quoteStrategy;
 
     /**
      * {@inheritdoc}
@@ -36,9 +76,11 @@ class SoftDeleteableWalker extends SqlWalker
         parent::__construct($query, $parserResult, $queryComponents);
 
         $this->conn = $this->getConnection();
-        $this->platform = $this->conn->getDatabasePlatform();
+        $this->platform = $this->getConnection()->getDatabasePlatform();
         $this->listener = $this->getSoftDeleteableListener();
-        $this->extractComponents($queryComponents);
+        $this->quoteStrategy = $this->getEntityManager()->getConfiguration()->getQuoteStrategy();
+
+        $this->extractComponents($this->getQueryComponents());
     }
 
     /**
@@ -50,8 +92,8 @@ class SoftDeleteableWalker extends SqlWalker
             case $AST instanceof DeleteStatement:
                 $primaryClass = $this->getEntityManager()->getClassMetadata($AST->deleteClause->abstractSchemaName);
 
-                return ($primaryClass->isInheritanceTypeJoined())
-                    ? new MultiTableDeleteExecutor($AST, $this, $this->meta, $this->platform, $this->configuration)
+                return $primaryClass->isInheritanceTypeJoined()
+                    ? new MultiTableDeleteExecutor($AST, $this, $this->meta, $this->getConnection()->getDatabasePlatform(), $this->configuration)
                     : new SingleTableDeleteUpdateExecutor($AST, $this);
             default:
                 throw new \Gedmo\Exception\UnexpectedValueException('SoftDeleteable walker should be used only on delete statement');
@@ -69,12 +111,13 @@ class SoftDeleteableWalker extends SqlWalker
         $class = $em->getClassMetadata($deleteClause->abstractSchemaName);
         $tableName = $class->getTableName();
         $this->setSQLTableAlias($tableName, $tableName, $deleteClause->aliasIdentificationVariable);
-        $quotedTableName = $class->getQuotedTableName($this->platform);
-        $quotedColumnName = $class->getQuotedColumnName($this->deletedAtField, $this->platform);
 
-        $sql = 'UPDATE '.$quotedTableName.' SET '.$quotedColumnName.' = '.$this->platform->getCurrentTimestampSQL();
+        $platform = $this->getConnection()->getDatabasePlatform();
 
-        return $sql;
+        $quotedTableName = $this->quoteStrategy->getTableName($class, $platform);
+        $quotedColumnName = $this->quoteStrategy->getColumnName($this->deletedAtField, $class, $platform);
+
+        return 'UPDATE '.$quotedTableName.' SET '.$quotedColumnName.' = '.$platform->getCurrentTimestampSQL();
     }
 
     /**
@@ -89,8 +132,8 @@ class SoftDeleteableWalker extends SqlWalker
         if (null === $this->listener) {
             $em = $this->getEntityManager();
 
-            foreach ($em->getEventManager()->getListeners() as $event => $listeners) {
-                foreach ($listeners as $hash => $listener) {
+            foreach ($em->getEventManager()->getListeners() as $listeners) {
+                foreach ($listeners as $listener) {
                     if ($listener instanceof SoftDeleteableListener) {
                         $this->listener = $listener;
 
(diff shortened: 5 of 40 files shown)


composer.json
```diff
@@ -40,17 +40,18 @@
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/persistence": "^1.3.4 || ^2.0"
+        "doctrine/persistence": "^1.3.3 || ^2.0"
     },
     "conflict": {
-        "doctrine/mongodb": "<1.3",
+        "doctrine/cache": "<1.11",
```
mbabker, line 46: Ensures `doctrine/cache` 2.0 compat layer is always present.

mbabker, line 46: Ensure's the `doctrine/cache` 2.0 compat layer is always present.

composer.json
```diff
@@ -40,17 +40,18 @@
...
+        "doctrine/dbal": "<2.13.1 || ~3.0.0",
         "doctrine/mongodb-odm": "<2.0",
         "doctrine/orm": "<2.10.2",
         "sebastian/comparator": "<2.0"
     },
     "require-dev": {
         "doctrine/cache": "^1.11 || ^2.0",
-        "doctrine/dbal": "^2.13",
+        "doctrine/dbal": "^2.13.1 || ^3.1",
```
mbabker, line 54: 2.13.0 had a few bugs with the forward compat layer, and 3.1 had various improvements in the compat layer.  This is the same constraint Symfony is using in the framework, should be good enough here.

composer.json
mbabker, line 46: Ensures the `doctrine/cache` 2.0 compat layer is always available.

src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
```diff
@@ -20,21 +20,19 @@ class MultiTableDeleteExecutor extends BaseMultiTableDeleteExecutor
     /**
      * {@inheritdoc}
      */
-    public function __construct(Node $AST, $sqlWalker, ClassMetadataInfo $meta, AbstractPlatform $platform, array $config)
+    public function __construct(Node $AST, $sqlWalker, ClassMetadata $meta, AbstractPlatform $platform, array $config)
```
mbabker, line 23: Technically could be a B/C break.  But, in practice, Doctrine isn't passing through the parent `ClassMetadataInfo` class, it typehints to `ClassMetadata` somewhat heavily, and this class really isn't meant for outside direct use.

src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
```diff
@@ -20,21 +20,19 @@ class MultiTableDeleteExecutor extends BaseMultiTableDeleteExecutor
...
     {
         parent::__construct($AST, $sqlWalker);
 
-        $getSqlStatements = \Closure::bind(function (): array {
-            return $this->_sqlStatements;
-        }, $this, static::class);
+        $sqlStatements = $this->_sqlStatements;
```
mbabker, line 27: The Closure isn't necessary, `$_sqlStatements` is protected and accessible in the context of a child class.

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
```diff
@@ -20,14 +24,50 @@
  */
 class SoftDeleteableWalker extends SqlWalker
 {
+    /**
+     * @var Connection
+     *
+     * @deprecated to be removed in 4.0, use the getConnection method instead
+     */
     protected $conn;
```
mbabker, line 32: The connection is available through a public getter on the parent class, no need to have a duplicate reference here.

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
```diff
@@ -20,14 +24,50 @@
...
+
+    /**
+     * @var AbstractPlatform
+     *
+     * @deprecated to be removed in 4.0, fetch the platform from the connection instead
+     */
     protected $platform;
```
mbabker, line 39: Since the connection is available through a getter, there isn't a strong need to hold a reference to the platform.  It can be fetched as needed at runtime.

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
```diff
@@ -118,10 +161,7 @@ private function extractComponents(array $queryComponents)
     {
         $em = $this->getEntityManager();
 
-        foreach ($queryComponents as $alias => $comp) {
-            if (!isset($comp['metadata'])) {
```
mbabker, line 122: From a practical perspective, the `metadata` key cannot be missing as the parent class' `setQueryComponent()` method requires it to be present.

tests/Gedmo/Loggable/Fixture/Entity/Address.php
```diff
@@ -16,10 +16,10 @@
 class Address
 {
     /**
-     * @var string
+     * @var int
      * @ORM\Id()
-     * @ORM\Column(type="string", length=36)
-     * @ORM\GeneratedValue(strategy="UUID")
```
mbabker, line 22: The UUID generator is removed in DBAL 3, and support for it is deprecated in ORM 2.10.  If the intent is to have a test with a UUID PK, then the test environment needs to be updated to handle UUID generation.  For now, I switched to a traditional integer PK to avoid breaking things with DBAL 3 installed.

tests/Gedmo/Sortable/Fixture/CustomerType.php
```diff
@@ -88,9 +91,20 @@ public function removeCustomer(Customer $customer)
...
+
+            // DBAL 3.x changes the signature of the DriverException constructor
+            $refl = new \ReflectionMethod(DriverException::class, '__construct');
+            $arg = $refl->getParameters()[0];
+
+            if ($arg->hasType() && DriverExceptionInterface::class === $arg->getType()->getName()) {
+                throw new ForeignKeyConstraintViolationException(DBALPDOException::new($pdoException), null);
+            }
+
+            // TODO - This throw can be removed when dropping DBAL 2.x support
```
phansys, line 107:
```suggestion
            // @todo: This `throw` can be removed when dropping support for doctrine/dbal 2.x.
```

tests/Gedmo/Sortable/Fixture/CustomerType.php
```diff
@@ -88,9 +91,20 @@ public function removeCustomer(Customer $customer)
...
     {
         if ($this->getCustomers()->count() > 0) {
-            // we imitate an foreign key constraint exception, because doctrine
-            // does not support sqlite constraints, which must be tested, too.
+            // we imitate a foreign key constraint exception because Doctrine
+            // does not support SQLite constraints, which must be tested, too.
+
             $pdoException = new \PDOException('SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails', '23000');
+
+            // DBAL 3.x changes the signature of the DriverException constructor
```
phansys, line 99:
```suggestion
            // doctrine/dbal 3.x changes the signature of the `DriverException::__construct()`.
```

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
```diff
@@ -20,14 +24,50 @@
  */
 class SoftDeleteableWalker extends SqlWalker
 {
+    /**
+     * @var Connection
+     *
+     * @deprecated to be removed in 4.0, use the getConnection method instead
```
phansys, line 30:
```suggestion
     * @deprecated to be removed in 4.0, use the `getConnection()` method instead.
```

phansys (changes requested):
IMO, we should add a changelog note, being aware that some changes could be seen as BC breaks.

src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
phansys: Could we mention this in the changelog?

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ a release.
 
 ### Added
 - Support to use Translatable annotations as attributes on PHP >= 8.0.
+- Added support for doctrine/dbal 3.x
```
phansys, line 25:
```suggestion
- Support for doctrine/dbal 3.x.
```

phansys (approved):

src/SoftDeleteable/Query/TreeWalker/Exec/MultiTableDeleteExecutor.php
mbabker: Done

CHANGELOG.md
mbabker: Covered in rebase

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
franmomu: It would be interesting maybe removing the parameter of the method and call `$this->getQueryComponents()` here (to [get the array shape information](https://github.com/doctrine/orm/blob/ca0a6bbf71c5b475291d47cca92925f149f48138/lib/Doctrine/ORM/Query/TreeWalker.php#L28-L36)) or maybe add the array shape as parameter.

tests/Gedmo/Loggable/Fixture/Entity/Address.php
```diff
@@ -16,10 +16,10 @@
 class Address
 {
     /**
-     * @var string
+     * @var int
```
franmomu, line 19:
```suggestion
     * @var int|null
```

tests/Gedmo/Loggable/Fixture/Entity/Address.php
```diff
@@ -45,7 +45,7 @@ class Address
     protected $geo;
 
     /**
-     * @return string
+     * @return int
```
franmomu, line 48:
```suggestion
     * @return int|null
```

tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
```diff
@@ -493,8 +494,7 @@ public function shouldFilterBeQueryCachedCorrectlyWhenToggledForEntity()
             static::markTestSkipped('Test only applies when doctrine/cache 1.x is installed');
         }
 
-        $cache = new ArrayCache();
-        $this->em->getConfiguration()->setQueryCacheImpl($cache);
+        $this->em->getConfiguration()->setQueryCache(CacheAdapter::wrap(new ArrayCache()));
```
franmomu, line 497: Can we use here an instance of `Symfony\Component\Cache\Adapter\ArrayAdapter` instead of the doctrine one as in other places?

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
phansys: I agree. That's why it's important to include SCA in our CI pipeline.

src/Translatable/Query/TreeWalker/TranslationWalker.php
```diff
@@ -457,8 +458,13 @@ private function getCastedForeignKey($component, $typeFK, $typePK)
             return $component;
         }
 
+        /*
+         * DBAL 3.x drops support of PostgreSQL 9.3 and earlier,
```
phansys, line 462:
```suggestion
         * doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier,
```

tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
mbabker: Based on the check just above, it looks like this test explicitly was meant to use the Doctrine Cache classes for some reason.

src/Translatable/Query/TreeWalker/TranslationWalker.php
```diff
@@ -457,8 +458,13 @@ private function getCastedForeignKey($component, $typeFK, $typePK)
             return $component;
         }
 
+        /*
+         * doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier,
+         * the below check prefers the `PostgreSQL94Platform` class as it is available in both major DBAL versions
+         * and falls back to the deprecated `PostgreSqlPlatform` class to account for older PostgreSQL versions with DBAL 2.x
+         */
         // try to look at postgres casting
```
phansys, line 466: IMO, we should avoid block comments inside a method's body.
```suggestion
        // doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier, the below check prefers the 
        // `PostgreSQL94Platform` class as it is available in both major DBAL versions and falls back to the 
        // deprecated `PostgreSqlPlatform` class to account for older PostgreSQL versions with DBAL 2.x.
        // Try to look at PostgreSQL casting.
```

tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
franmomu: ah ok, didn't see that 👍

src/Translatable/Query/TreeWalker/TranslationWalker.php
```diff
@@ -457,8 +458,11 @@ private function getCastedForeignKey($component, $typeFK, $typePK)
             return $component;
         }
 
-        // try to look at postgres casting
-        if ($this->platform instanceof PostgreSqlPlatform) {
+        // doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier, the below check prefers the 
+        // `PostgreSQL94Platform` class as it is available in both major DBAL versions and falls back to the
```
phansys, line 462: Please, check the CI status.
```suggestion
        // doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier, the below check prefers the
        // `PostgreSQL94Platform` class as it is available in both major DBAL versions and falls back to the
```

phansys:
@mbabker, @franmomu; do you think we could have this PR merged soon? I'm deferring the next release in order to include these changes, if possible.
Please, tell me if we should wait a bit longer, otherwise we can create a release without this PR, and make a new release when it is finished.
I'll be waiting for your feedback. Thank you.

mbabker:
I just fixed the merge conflicts, so we should be good here.

phpstan.neon.dist
```diff
@@ -8,3 +8,8 @@ parameters:
         - tests
     bootstrapFiles:
         - tests/bootstrap.php
+
+    ignoreErrors:
+        # Ignore errors about classes removed from doctrine/dbal 3.x
+        - '/Class Doctrine\\DBAL\\Platforms\\(MySqlPlatform|PostgreSqlPlatform) not found\./'
+        - '/Instantiated class Doctrine\\DBAL\\Driver\\PDOException not found\./'
```
mbabker, line 15: `MySqlPlatform` had its casing changed to `MySQLPlatform` in DBAL 3.  That one's pretty much a non-issue with PHP being case insensitive with class names, but PHPStan flags it (could probably fix it to the updated casing and be done with it).  The other two classes are fully removed in DBAL 3.  Unless the PHPStan config is going to get crazy and a run is done with both DBAL 2 and DBAL 3 (some of the Doctrine repos are doing this), just ignoring these would be the best way to deal with it.  I didn't put it in the baseline just because I did this commit through the GitHub UI and don't have the time right now to do a full clone, run PHPStan, and re-generate the baseline file, but it's easy to move these ignores if need be.

phansys, line 15:
```suggestion
        - '#^Class Doctrine\\DBAL\\Platforms\\(MySqlPlatform|PostgreSqlPlatform) not found\.$#'
        - '#^Instantiated class Doctrine\\DBAL\\Driver\\PDOException not found\.$#'
```
I don't have time right now to check, but I'll try to take a look soon.
If we must keep these errors ignored, I agree they should be included in the baseline.

phansys: If these classes are not present in doctrine / dbal 3.x and our code is relying on them without any BC layer, I think we can't allow the version bumping until this situation is resolved.

mbabker: Of the 6 reported errors, 4 of them come from the class casing and 2 are B/C conditionals.  PHPStan just can't work that out on its own.

The 3 class not found errors from `TranslationWalker.php` are a mix of the mixed `MySqlPlatform` versus `MySQLPlatform` casing or the B/C layer applied in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2284/files#diff-07d23b78d99c87e3121e586d76ea31fb9f3a5f2eb97602a123c6811074f5d379R464-R468

The class not found error from `CustomerType.php` has the B/C layer at https://github.com/doctrine-extensions/DoctrineExtensions/pull/2284/files#diff-367f5ccd73a06127b2dcc52d40484b3f73bb19fd10ac5c1e93266f04f2ab68e5L91-L94 covering it

And the 2 class not found errors from `BaseTestCaseOM.php` and `BaseTestCaseORM.php` are again the mixed MySQL platform cases.

tests/Gedmo/Sortable/Fixture/CustomerType.php
```diff
@@ -88,9 +91,20 @@ public function removeCustomer(Customer $customer)
...
         if ($this->getCustomers()->count() > 0) {
-            // we imitate an foreign key constraint exception, because doctrine
-            // does not support sqlite constraints, which must be tested, too.
+            // we imitate a foreign key constraint exception because Doctrine
+            // does not support SQLite constraints, which must be tested, too.
+
             $pdoException = new \PDOException('SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails', '23000');
+
+            // doctrine/dbal 3.x changes the signature of the `DriverException::__construct()`.
+            $refl = new \ReflectionMethod(DriverException::class, '__construct');
```
phansys, line 100: Isn't simpler to check against the existence of the `Doctrine\DBAL\Driver\PDOException` class in order to determine the installed `doctrine/dbal` version?

phansys:
@mbabker, @franmomu; thank you for your effort and patience.
I'll release `3.3.0` at the end of the day. If this PR can't be merged yet, we'll have to wait for the next release.
Not really a big deal, as we have many PRs to close and many improvements to do, so the next release will be soon.

CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ a release.
 
 ### Added
 - Support to use Translatable annotations as attributes on PHP >= 8.0.
+- Support for doctrine/dbal 3.x
+
+### Changes
+- Dropped support for doctrine/dbal < 2.13
```
stof, line 31: should say `<2.13.1`

src/Translatable/Query/TreeWalker/TranslationWalker.php
```diff
@@ -460,8 +461,11 @@ private function getCastedForeignKey($component, $typeFK, $typePK)
             return $component;
         }
 
-        // try to look at postgres casting
-        if ($this->platform instanceof PostgreSqlPlatform) {
+        // doctrine/dbal 3.x drops support for PostgreSQL 9.3 and earlier, the below check prefers the
+        // `PostgreSQL94Platform` class as it is available in both major DBAL versions and falls back to the
+        // deprecated `PostgreSqlPlatform` class to account for older PostgreSQL versions with DBAL 2.x.
```
stof, line 466: Note that DBAL 3.2 reintroduces the `PostgreSqlPlatform` class as the base class for postgresql platforms (still supporting only PG 9.4+)

mbabker: Well, ain't that something.  Either way, the second `instanceof` check is going to be needed for DBAL 3.1 I guess.

CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- Support for doctrine/dbal 3.x
+
+### Changes
+- Dropped support for doctrine/dbal < 2.13.1
+- The third argument of `Gedmo\SoftDeleteable\Query\TreeWalker\Exec\MultiTableDeleteExecutor` now requires a `Doctrine\ORM\Mapping\ClassMetadata` instance
```
phansys, line 27:
```suggestion
- The third argument of `Gedmo\SoftDeleteable\Query\TreeWalker\Exec\MultiTableDeleteExecutor::__construct()` requires a `Doctrine\ORM\Mapping\ClassMetadata` instance.
```

phpstan-baseline.neon
```diff
@@ -1,11 +1,17 @@
 parameters:
-    ignoreErrors:
-        -
-            message: "#^Class Doctrine\\\\ODM\\\\PHPCR\\\\DocumentManager not found\\.$#"
-            count: 2
-            path: src/References/Mapping/Event/Adapter/ORM.php
+	ignoreErrors:
```
phansys, line 2: Please, keep the previous indentation.

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
```diff
@@ -89,8 +132,8 @@ private function getSoftDeleteableListener()
         if (null === $this->listener) {
             $em = $this->getEntityManager();
 
-            foreach ($em->getEventManager()->getListeners() as $event => $listeners) {
-                foreach ($listeners as $hash => $listener) {
+            foreach ($em->getEventManager()->getListeners() as $listeners) {
+                foreach ($listeners as $listener) {
                     if ($listener instanceof SoftDeleteableListener) {
                         $this->listener = $listener;
```
phansys, line 139: Couldn't we use `break 2` in the next line and remove the extra `break` condition in the first `foreach`?

phpstan-baseline.neon
mbabker: That's how PHPStan dumps it.  The previous indention requires manual changes after `--generate-baseline`, and I'd say that's not a good practice for a machine-generated file.

src/SoftDeleteable/Query/TreeWalker/SoftDeleteableWalker.php
mbabker: Probably.  But, I'd do this change as a separate PR because there are a bunch of classes with a very similar loop.

franmomu (approved):

phansys:
Thank you @mbabker!

PR #2285, opened 2021-11-08
Use `::class`

franmomu:


tests/Gedmo/SoftDeleteable/SoftDeleteableDocumentTest.php
```diff
@@ -27,8 +30,8 @@ class SoftDeleteableDocumentTest extends BaseTestCaseMongoODM
     public const MODULE_CLASS = 'Gedmo\Tests\SoftDeleteable\Fixture\Document\Module';
```
phansys, line 30: IMO, these constants that were not replaced can be safely removed, as these classes don't exist.

phansys: By now, I'll merge this as is. Later, we can create a new PR for these removals.

phansys:
Thank you @franmomu!

PR #2286, opened 2021-11-08
Avoid using the reflection API

phansys:
Follows #2225.

### DIFF
--- a/src/Uploadable/Mapping/Validator.php
+++ b/src/Uploadable/Mapping/Validator.php
@@ -195,21 +195,8 @@ class Validator
             case self::FILENAME_GENERATOR_NONE:
                 break;
             default:
-                $ok = false;
-
-                if (class_exists($config['filenameGenerator'])) {
-                    $refl = new \ReflectionClass($config['filenameGenerator']);
-
-                    if ($refl->implementsInterface(FilenameGeneratorInterface::class)) {
-                        $ok = true;
-                    }
-                }
-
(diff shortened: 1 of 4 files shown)


franmomu (approved):

src/Uploadable/Mapping/Validator.php
```diff
@@ -195,21 +195,8 @@ public static function validateConfiguration(ClassMetadata $meta, array &$config
...
-                    }
-                }
-
-                if (!$ok) {
-                    $msg = 'Class "%s" needs a valid value for filenameGenerator. It can be: SHA1, ALPHANUMERIC, NONE or ';
-                    $msg .= 'a class implementing FileGeneratorInterface.';
-
-                    throw new InvalidMappingException(sprintf($msg, $meta->name));
+                if (!class_exists($config['filenameGenerator']) || !is_subclass_of($config['filenameGenerator'], FilenameGeneratorInterface::class)) {
+                    throw new InvalidMappingException(sprintf('Class "%s" needs a valid value for filenameGenerator. It can be: SHA1, ALPHANUMERIC, NONE'.' or a class implementing %s.', $meta->name, FilenameGeneratorInterface::class));
```
franmomu, line 199: should we use multilines? to be more readable

phansys: I think yes, but currently they are not allowed by our CS rules.
We should create a separate PR for this matter.

PR #2291, opened 2021-11-09
Fix value passed in the `--config` option to `fix-cs` Composer script

phansys:


composer.json
```diff
@@ -84,7 +84,7 @@
         }
     },
     "scripts": {
-        "fix-cs": "php bin/php-cs-fixer fix --config=.php_cs.dist"
+        "fix-cs": "php-cs-fixer fix --config=.php-cs-fixer.dist.php"
```
franmomu, line 87:
```suggestion
        "fix-cs": "bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php"
```
To use the local `php-cs-fixer`

phansys: It should use the local `php-cs-fixer`: https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands.

franmomu: Didn't know about that 👍

franmomu (approved):

franmomu:
Thanks @phansys!

PR #2295, opened 2021-11-09
Apply "header_comment" CS rule

phansys:


franmomu (approved):
since there are some open PR, I'll leave you to merge them in the order you prefer

PR #2301, opened 2021-11-12
Closing the API by marking classes as final and internal

phansys:
**TODO**:
- [x] Resolve the cases where the test classes are extending classes marked as final.

### DIFF
--- a/src/Blameable/BlameableListener.php
+++ b/src/Blameable/BlameableListener.php
@@ -18,6 +18,8 @@ use Gedmo\Exception\InvalidArgumentException;
  * dates on creation and update.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ *
+ * @final since gedmo/doctrine-extensions 3.x
  */
 class BlameableListener extends AbstractTrackingListener
 {
--- a/src/Blameable/Mapping/Driver/Annotation.php
+++ b/src/Blameable/Mapping/Driver/Annotation.php
@@ -20,6 +20,8 @@ use Gedmo\Mapping\Driver\AbstractAnnotationDriver;
  * extension.
  *
  * @author David Buchmann <mail@davidbu.ch>
+ *
+ * @internal
  */
 class Annotation extends AbstractAnnotationDriver
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ a release.
 - Sortable: Fix return value check of Comparable interface (#2541)
 
 ### Deprecated
+- In order to close the API, `@final` and `@internal` annotations were added to all non base classes, which means that extending
+  these classes is deprecated and can not be inherited in version 4.0.
 - Sortable: Accepting a return type other than "integer" from `Comparable::compareTo()` is deprecated in `SortableListener::postFlush()`. This will not be accepted in version 4.0.
 
 ## [3.10.0] - 2022-11-14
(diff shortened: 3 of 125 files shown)


src/Blameable/BlameableListener.php
```diff
@@ -11,6 +11,8 @@
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
+ *
+ * @final since gedmo/doctrine-extensions 3.x
```
stof, line 15: should mention the actual minor version rather than saying `3.x`

phansys: I'm performing these replacements during the release process, as we don't know yet if the next minor release will include this change. Take a look at #2296 for an example.

src/Blameable/Mapping/Driver/Annotation.php
```diff
@@ -20,6 +20,8 @@
  * extension.
  *
  * @author David Buchmann <mail@davidbu.ch>
+ *
+ * @final since gedmo/doctrine-extensions 3.x
  */
 class Annotation extends AbstractAnnotationDriver
```
franmomu, line 26: When adding support for attributes, this `Annotation` class would be extended by an `Attribute` class to not have duplicated code, maybe we should mark as internal some of them? like the drivers?

src/Loggable/Document/LogEntry.php
```diff
@@ -24,6 +24,8 @@
  *         @MongoODM\Index(keys={"username"="asc"})
  *     }
  * )
+ *
+ * @final since gedmo/doctrine-extensions 3.x
```
franmomu, line 28: I think this one shouldn't be `final` since is a Document and it could be proxied

src/Translatable/Document/Translation.php
```diff
@@ -27,6 +27,8 @@
  *      "object_class" = "asc",
  *      "foreign_key" = "asc"
  * })
+ *
+ * @final since gedmo/doctrine-extensions 3.x
```
franmomu, line 31: Same here

src/Tree/Strategy/ODM/MongoDB/MaterializedPath.php
```diff
@@ -21,6 +21,8 @@
  *
  * @author Gustavo Falco <comfortablynumb84@gmail.com>
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ *
+ * @todo Make this class final.
```
franmomu, line 25: this line is different

franmomu:
@phansys do you want to take a look at this before releasing next minor version or do we defer this to the next one?

phansys:
> @phansys do you want to take a look at this before releasing next minor version or do we defer this to the next one?

I'd like to merge this PR this week, but as I'm not able to confirm this expectation, I'd say you should not wait.
Thank you!

src/Sluggable/SluggableListener.php
```diff
@@ -27,6 +27,8 @@
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  * @author Klein Florian <florian.klein@free.fr>
+ *
+ * @internal
```
stof, line 31: should not be internal, as this class is expected to be instantiated by downstream code to use the library

franmomu (approved):

franmomu (approved):

franmomu (approved):

franmomu (approved):

PR #2310, opened 2021-11-15
Normalize usages of `instanceof` operator

phansys:


franmomu:
What's the reason behind this change? I remember reading that `instanceof` had better performance.

phansys:
>What's the reason behind this change? I remember reading that instanceof had better performance.

We need to replace cases like `$node instanceof $meta->name` with `$node instanceof $meta->getName()` (as the `name` property is not guaranteed by the interface), and this syntax is not valid (see https://3v4l.org/Dq4mU).

franmomu (approved):
Thanks for the explanation!

PR #2311, opened 2021-11-15
Fix PHPDoc syntax in `@param`, `@return` and `@throws` docblocks

phansys:


src/Translatable/Mapping/Event/Adapter/ORM.php
```diff
@@ -101,26 +101,26 @@ public function loadTranslations($object, $translationClass, $locale, $objectCla
...
 
     /**
-     * Transforms foreigh key of translation to appropriate PHP value
+     * Transforms foreing key of translation to appropriate PHP value
      * to prevent database level cast
      *
-     * @param $key - foreign key value
-     * @param $className - translation class name
+     * @param mixed  $key       foreign key value
+     * @param string $className translation class name
```
franmomu, line 108: we may also add a `phpstan-param class-string $className` (it will be required when reaching higher levels of PHPStan)

franmomu (approved):

src/Translatable/Mapping/Event/Adapter/ORM.php
phansys: Yes, I agree. My intention is to bump the PHPStan level in separate PRs later.

PR #2312, opened 2021-11-16
Fix `Gedmo\Mapping\Driver\Xml::_loadMappingFile()` behavior

phansys:
Fixes `Gedmo\Mapping\Driver\Xml::_loadMappingFile()` behavior in scenarios where `libxml_disable_entity_loader(true)` was previously called.

Closes #1945,


franmomu (approved):

franmomu:
thanks @phansys and @Piskvor!

PR #2322, opened 2021-11-18
[Translatable] Fix mixing attributes and annotations

### LINKED ISSUE #2321: Translatable with attributes and annotations
In https://github.com/doctrine-extensions/DoctrineExtensions/pull/2253 I added support for attributes for Translatable extension, but I missed mixing attributes and annotations.

The problem I've found is that the driver used (`Annotation`, `Attribute`, etc) is based on the doctrine configuration:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/46e102535be5eb978549520cc3645a1571cb4c03/src/Mapping/ExtensionMetadataFactory.php#L141-L196

So if someone has defined its mapping using attributes, it uses the `AttributeReader` and it won't be able to read our annotations.

We could maybe create another reader with both `AttributeReader` and `AnnotationReader` and try with attributes and then fallback to annotations.

I'll try to implement it, but not sure if I have time this week.

The solution for someone facing these issues is to use attributes for the Translatable extension in the meantime.

franmomu:
Fixes #2321

It adds a new `AttributeAnnotationReader` which tries to read attributes and if it fails, it reads annotations.

### DIFF
--- a/src/Mapping/Driver/AttributeAnnotationReader.php
+++ b/src/Mapping/Driver/AttributeAnnotationReader.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Gedmo\Mapping\Driver;
+
+use Doctrine\Common\Annotations\Reader;
+use Gedmo\Mapping\Annotation\Annotation;
+use ReflectionClass;
+use ReflectionMethod;
+
+/**
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ * @license MIT License (http://www.opensource.org/licenses/mit-license.php)
+ *
+ * @internal
+ */
+final class AttributeAnnotationReader implements Reader
+{
+    /**
+     * @var Reader
+     */
+    private $annotationReader;
+
+    /**
+     * @var AttributeReader
+     */
+    private $attributeReader;
+
+    public function __construct(AttributeReader $attributeReader, Reader $annotationReader)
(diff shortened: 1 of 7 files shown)


src/Mapping/Driver/AttributeAnnotationReader.php
```diff
@@ -0,0 +1,93 @@
...
+        if (null !== $annotation) {
+            return $annotation;
+        }
+
+        return $this->annotationReader->getPropertyAnnotation($property, $annotationName);
+    }
+
+    public function getMethodAnnotations(ReflectionMethod $method)
+    {
+        throw new \LogicException('Not implemented');
```
phansys, line 86:
```suggestion
        throw new \BadMethodCallException('Not implemented.');
```

src/Mapping/Driver/AttributeAnnotationReader.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
+    {
+        throw new \LogicException('Not implemented');
```
phansys, line 91:
```suggestion
        throw new \BadMethodCallException('Not implemented.');
```

phansys (approved):

phansys:
Thank you @franmomu!

PR #2323, opened 2021-11-18
3.3.1

phansys:


franmomu:
Thanks! I was going to do it, I think we have to also change https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/DoctrineExtensions.php#L29

franmomu (approved):

PR #2327, opened 2021-11-18
Apply "php_unit_strict" CS rule

phansys:


franmomu (approved):
Haven't check all the files, but since it's automated 👍

phansys:
>Haven't check all the files, but since it's automated +1

The only file where I made manual changes is `tests/Gedmo/References/ReferencesListenerTest.php`, as there was a comparison between 2 objects.

PR #2332, opened 2021-11-18
Apply "psr_autoloading" CS rule

phansys:


tests/Gedmo/Mapping/Mock/Extension/Encoder/Mapping/Annotations.php
```diff
@@ -1,15 +1,13 @@
...
-
 namespace Gedmo\Tests\Mapping\Mock\Extension\Encoder\Mapping;
 
 use Doctrine\Common\Annotations\Annotation;
 
 /**
  * @Annotation
  */
-final class Encode extends Annotation
+final class Annotations extends Annotation
```
franmomu, line 10: Should we change the file name to `Encode` instead?  sounds better IMHO

phansys: I agree. Done.

franmomu (approved):

franmomu:
thanks @phansys!

PR #2335, opened 2021-11-20
Refresh the example app code

### LINKED ISSUE #2262: Update em.php example to >v3.0
Is it possible to update the [file](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/example/em.php)?
I'm upgrading from 2.4 and ArrayCache, ApcuCache, CachedReader and such are all removed and i'd love a new example

mbabker:
Fixes #2262

This refreshes the example app code to work against the current ORM, DBAL, and extensions versions while doing its best to avoid deprecated APIs and using the tools that the DBAL and ORM provide to set things up efficiently.

This also moves the `example/run.php` logic into a console command in the example application, with the thinking here being that since there is already a console application being set up with all the DBAL/ORM utilities, that can also be used to create more examples than just the existing tree/translation demo if someone really wanted to.


franmomu (approved):
nice clean up 👍

phansys (commented):
Good job!
My only concern is that some of these changes make use of classes declared in packages not required in `composer.json` (`symfony/console`).

mbabker:
> My only concern is that some of these changes make use of classes declared in packages not required in `composer.json` (`symfony/console`).

The console component was never in the dev dependencies to begin with (even though most of the example code uses it).  I've added it.

example/em.php
```diff
@@ -9,123 +9,126 @@
  * file that was distributed with this source code.
  */
 
-// This entity manager configuration works with doctrine 2.1.x and 2.2.x
-// versions. Regarding AnnotationDriver setup it most probably will be changed into
-// xml. Because annotation driver fails to read other classes in same namespace.
+/**
```
phansys, line 12: IMO, we should not use a block comment here.

franmomu (approved):

phansys:
Thank you @mbabker!

PR #2342, opened 2021-11-24
Apply "declare_strict_types" CS rule

phansys:


franmomu (approved):
Should we mention this in the changelog or something? I guess someone using this package may get errors using the wrong type from now on

phansys:
> Should we mention this in the changelog or something? I guess someone using this package may get errors using the wrong type from now on

I think you're right. There are some cases where the codebase is using user provided options (which were not filtered by typed arguments) as arguments in function calls.

AkenRoberts:
This would probably have to be a major release, then, eh? I don't think there's a way to mark this as incompatible like we would using Composer's `conflict` or similar version bump.

phansys:
>This would probably have to be a major release, then, eh? I don't think there's a way to mark this as incompatible like we would using Composer's conflict or similar version bump.

Yes, I'll add the `declare` construct only for the scripts under `tests/` directory by now, without the CS rule.
Then, we can defer the rule for the next major :+1:

PR #2345, opened 2021-11-26
Add type declarations in private functions

phansys:


src/Loggable/Mapping/Driver/Xml.php
```diff
@@ -85,10 +85,8 @@ public function readExtendedMetadata($meta, array &$config)
 
     /**
      * Searches mappings on element for versioned fields
-     *
-     * @param object $meta
      */
-    private function inspectElementForVersioned(\SimpleXMLElement $element, array &$config, $meta)
+    private function inspectElementForVersioned(\SimpleXMLElement $element, array &$config, object $meta): void
```
franmomu, line 89:
```suggestion
    private function inspectElementForVersioned(\SimpleXMLElement $element, array &$config, ClassMetadata $meta): void
```
importing `Doctrine\Persistence\Mapping\ClassMetadata`

src/Translatable/Mapping/Driver/Yaml.php
```diff
@@ -80,7 +80,7 @@ protected function _loadMappingFile($file)
         return \Symfony\Component\Yaml\Yaml::parse(file_get_contents($file));
     }
 
-    private function buildFieldConfiguration($field, $fieldMapping, array &$config)
+    private function buildFieldConfiguration(string $field, $fieldMapping, array &$config): void
```
franmomu, line 83:
```suggestion
    private function buildFieldConfiguration(string $field, array $fieldMapping, array &$config): void
```

based on 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/477dca8ca53a418c04a46dad6c30253837a04312/src/Sluggable/Mapping/Driver/Yaml.php#L90

src/Translatable/TranslatableListener.php
```diff
@@ -555,24 +555,19 @@ protected function validateLocale($locale)
      * Check if the given locale is valid
      *
      * @param string $locale locale to check
```
franmomu, line 557:
```suggestion
```

src/Translator/TranslationProxy.php
```diff
@@ -153,10 +153,8 @@ public function setTranslatedValue($property, $value)
      *
      * @param string $property object property name
      * @param string $locale   locale name
```
franmomu, line 155:
```suggestion
```

src/Translator/TranslationProxy.php
```diff
@@ -153,10 +153,8 @@ public function setTranslatedValue($property, $value)
...
-     *
-     * @return Translation
      */
-    private function findOrCreateTranslationForProperty($property, $locale)
+    private function findOrCreateTranslationForProperty(string $property, string $locale): Translation
```
franmomu, line 157:
```suggestion
    private function findOrCreateTranslationForProperty(string $property, string $locale): TranslationInterface
```

tests/Gedmo/Mapping/Mock/Extension/Encoder/EncoderListener.php
```diff
@@ -65,7 +65,7 @@ protected function getNamespace()
         return __NAMESPACE__;
     }
 
-    private function encode(EventAdapterInterface $ea, $object, $config)
+    private function encode(EventAdapterInterface $ea, $object, $config): void
```
franmomu, line 68:
```suggestion
    private function encode(EventAdapterInterface $ea, object $object, array $config): void
```

src/Translatable/Mapping/Driver/Yaml.php
phansys: But in this case, there is a check against `is_array()` for the argument 2. If the check is not satisfied, currently the function does nothing.

franmomu: yeah, sorry, what I meant is that those methods are called with the same context:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/477dca8ca53a418c04a46dad6c30253837a04312/src/Sluggable/Mapping/Driver/Yaml.php#L54-L64

and

https://github.com/doctrine-extensions/DoctrineExtensions/blob/477dca8ca53a418c04a46dad6c30253837a04312/src/Translatable/Mapping/Driver/Yaml.php#L56-L66

So `$fieldMapping` should be always an array (if in one place is typed as `array`).

franmomu (approved):

src/Translator/TranslationProxy.php
```diff
@@ -18,25 +18,41 @@
...
     protected $properties = [];
+    /**
+     * @var string
+     *
+     * @phpstan-var class-string<TranslationInterface>
+     */
     protected $class;
     /**
-     * @var Collection|TranslationInterface[]
+     * @var Collection<TranslationInterface>
```
franmomu, line 40: For `Collection` I think you must specify also the key type: https://github.com/doctrine/collections/blob/42a111afab917a328fdc38a35101c089ce52b580/lib/Doctrine/Common/Collections/Collection.php#L27-L32.

Since here I think it is only used `$this->coll->add($translation);` I would use `int` as key type.

franmomu (approved):

franmomu:
thanks @phansys!

PR #2348, opened 2021-11-29
Add support for attribute mapping for `doctrine/mongodb-odm`

franmomu:
`doctrine/mongodb-odm` will support attributes on the next minor version, this PR adds compatibility for users using attribute mapping with mongodb.

~I'll leave it as draft until that version is released.~


phansys:
Thank you @franmomu!

PR #2349, opened 2021-11-29
Test with PHP 8.1 on CI

franmomu:
Tests were changed because of https://github.com/php/php-src/blob/dab6226cbe247fe926e43a1230ead15b8309be4c/UPGRADING#L130-L134


phansys:
Thanks @franmomu!

PR #2352, opened 2021-11-30
Apply "no_null_property_initialization" CS rule

phansys:


src/Mapping/Annotation/Slug.php
```diff
@@ -30,7 +30,7 @@ final class Slug extends Annotation
     /** @var bool */
     public $unique = true;
     /** @var string */
-    public $unique_base = null;
+    public $unique_base;
```
franmomu, line 33: I'm not sure about if this rule can be applied in general (at least without fixing the type declarations).

In this case `$unique_base` looks more like `string|null` and when we add property type declarations I guess this will become:

`public ?string $unique_base = null` instead of `public string $unique_base =''`

based on the usage: https://github.com/doctrine-extensions/DoctrineExtensions/search?q=unique_base

phansys: I think this change should be done for consistency.
When we can fix the type hints or add the type declarations, I guess we should evaluate all the properties, as most of them are allowing null values or even (implicitly) using `null` as default value.

franmomu (approved):

PR #2357, opened 2021-12-04
Allow to use Loggable extension with attributes

franmomu:


tests/Gedmo/Loggable/Fixture/Entity/Geo.php
```diff
@@ -21,86 +22,69 @@
  *
  * @ORM\Embeddable()
  */
+#[ORM\Embeddable]
 class Geo
 {
     /**
-     * @var string
+     * @var float
```
phansys, line 29: Is this really using `float`? https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#decimal.

phansys (approved):

tests/Gedmo/Loggable/Fixture/Entity/Geo.php
mbabker: No, it is a string.  If you want a float exposed to the application, you need to handle type conversions in the getter/setter or everything should use the value as a string.  This is what I do in my apps:

```php
#[ORM\Entity]
class Entity
{
    /**
     * @phpstan-var numeric-string|null
     */
    #[ORM\Column(type: Types::DECIMAL, scale: 7, nullable: true)]
    private ?string $latitude = null;

    public function setLatitude(?float $latitude): void
    {
        if ($latitude !== null) {
            /** @phpstan-var numeric-string $formattedLatitude */
            $formattedLatitude = sprintf('%.7f', $latitude);

            $this->latitude = $formattedLatitude;
        } else {
            $this->latitude = null;
        }
    }

    public function getLatitude(): ?float
    {
        return \is_string($this->latitude) ? (float) $this->latitude : $this->latitude;
    }
}
```

franmomu: Thank you for the reviews! I was adding these types declarations in a second iteration and I just used `float` because in the tests a float is being used as a argument, but didn't paid attention to the mapping type.

phansys:
Thanks @franmomu!

PR #2360, opened 2021-12-05
Expand type info on tree annotations, deprecate unused `@Tree::$identifierMethod`

mbabker:
Adds `@phpstan-var` annotations when specific values are expected to improve static analysis

Deprecates the unused `@Tree::$identifierMethod` property, this is actually read from the `@TreeRoot` annotation


phansys:
Forcing a fresh build.

phansys (approved):

franmomu (approved):

franmomu:
Thanks you @mbabker!

franmomu:
maybe an entry in the changelog should be added for the deprecated property?

PR #2361, opened 2021-12-05
Dropped support for PHP 7.2 and 7.3

stephanvierkant:
In a few hours, PHP 7.3 becomes End of Life. That's why I think it's safe to drop support for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ a release.
 
 ## [Unreleased]
 
+### Changed
+- Dropped support for PHP 7.2 and 7.2, which are End of Life (EOL)
+
 ## [3.4.0] - 2021-12-05
 ### Added
 - PHP 8 Attributes support for Doctrine MongoDB to document & traits.
--- a/.docker/php/Dockerfile
+++ b/.docker/php/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2-cli
+FROM php:7.4-cli
 
 COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,8 +23,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.2"
-          - "7.3"
           - "7.4"
           - "8.0"
         deps:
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ a release.
 
 ## [Unreleased]
 
+### Changed
+- Dropped support for PHP 7.2 and 7.2, which are End of Life (EOL)
```
franmomu, line 24:
```suggestion
- Dropped support for PHP 7.2 and 7.3, which are End of Life (EOL)
```

franmomu (approved):

phansys (commented):
My first opinion about this change is that we should include it in a MAJOR release.
Otherwise, I think we should wait for the fix of the bugs that are already reported, as after this change the users experiencing these bugs and using PHP < 7.4 will be not allowed to update.

mbabker:
Based on https://packagist.org/packages/gedmo/doctrine-extensions/php-stats#3 it already looks like about 90% of 3.x installs are coming from PHP 7.4+ users.  So dropping 7.2 and 7.3 wouldn’t be majorly disruptive, but definitely annoying if you’re in that 10%.

stephanvierkant:
My opinion is that dropping support for an EOL-version of PHP/Symfony isn't a breaking change and can be done in a minor release.

And because we add this requirement to composer.json, you can't even update this library so you'll stay on the previous release anyway. Breaking changes in a minor release are annoying because you can break applications by just running `composer update`, in this case you can't.

(and don't ask for new bugfix releases from this library if you are using a dangerous EOL-version of PHP)

tacman:
I think we're doing people a favor dropping support for EOL versions of php (and Symfony).  

You shouldn't be using PHP 7.2.  If you're using it, you should lock to a pre-EOL commit, but really you should use this as an excuse to talk to whatever powers-that-be that require you to use outdated version of PHP and convince them to update.

Using 7.4 gives us lots of opportunities to code better (e.g. typehints).  We shouldn't be catering to outdated versions.  Sure, make it a minor release, but everything we can do to stop people from using outdated versions of PHP helps the ecosystem as a whole.

AkenRoberts:
My opinion is to do it on a major version.

@stephanvierkant is correct in that it doesn't _have_ to be done in a major version because of Composer's platform checks. However, I agree with @phansys — we can fix bugs and improve the code base without forcing PHP upgrades or alienating users.

There are bigger fish to fry and other significant breaking changes that will be needed. Bumping the minimum PHP version at that time would be trivial.

tacman:
Perhaps when Doctrine ORM 3 comes out?

Issue #2363, opened 2021-12-06
Support of PHP 8 attributes


franmomu:
This is an issue to keep track the work needed to fully support PHP 8 attributes.

In `3.2.0` we added support to use PHP 8 attributes with Doctrine ORM mapping, in `3.3.0` we added support to use Translatable annotations as attributes and in `3.4.0` we added support to use PHP 8 attributes with Doctrine MongoDB and also added support to use Timestampable and Loggable annotations as attributes.

This is the list of extensions:

- [x] [Blameable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2375)
- [x] [Loggable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2357)
- [x] [Sluggable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2379)
- [x] [Timestampable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2325)
- [x] [Translatable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2253)
- [x] [Tree](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2388)
- [x] [IpTraceable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2377)
- [x] [SoftDeleteable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2374)
- [x] [Sortable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2380)
- [x] [Uploadable](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2386)
- [x] [References](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2389)
- [x] [ReferenceIntegrity](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2389)

~We may need to change the `AttributeReader` to support repeatable attributes.~

tacman:
I can work on the Tree extension, I'll submit a PR this afternoon.

tacman:
I think Tree is almost done, but I need some help bringing it across the finish line.  

I can take a look at Slug next unless someone else is already working on it.

franmomu:
> I can take a look at Slug next unless someone else is already working on it.

@tacman have you started with Slug? I think I can try that one.

tacman:
No, I haven't.  Thanks!

vstelmakh:
I prepared the PR for Sortable #2380. Please add it to issue description.

[...]

PR #2366, opened 2021-12-07
allow php8 attributes for Tree extension

tacman:
This PR follows the pattern in Loggable to allow for PHP 8 attributes, #2363

However, it doesn't quite work yet, when trying to use a class that implements this I get

    Tree object class: App\Entity\Location must have tree metadata at this point

I imagine that in src/Tree/Mapping/Driver/Annotation.php, it is not loading the attributes, but I can't figure out where that's happening.  Can someone point me in the right direction?

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - support-attributes
   pull_request:
 
 jobs:
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - support-attributes
   pull_request:
 
 jobs:
--- a/src/Mapping/Annotation/Slug.php
+++ b/src/Mapping/Annotation/Slug.php
@@ -9,7 +9,9 @@
 
 namespace Gedmo\Mapping\Annotation;
 
+use Attribute;
 use Doctrine\Common\Annotations\Annotation;
+use Gedmo\Mapping\Annotation\Annotation as GedmoAnnotation;
 
 /**
  * Slug annotation for Sluggable behavioral extension
@@ -19,7 +21,8 @@ use Doctrine\Common\Annotations\Annotation;
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  */
-final class Slug extends Annotation
+#[Attribute(Attribute::TARGET_PROPERTY)]
+final class Slug implements GedmoAnnotation
 {
     /** @var array<string> @Required */
     public $fields = [];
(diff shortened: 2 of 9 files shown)


tacman:
I added the Attribute class, and am doing some basic testing now.

src/Mapping/Annotation/Tree.php
```diff
@@ -42,4 +45,21 @@ final class Tree extends Annotation
      * @deprecated to be removed in 4.0, unused, configure the property on the TreeRoot annotation instead
      */
     public $identifierMethod;
+
+    /**
+     * @phpstan-param class-string|null $type
+     */
+    public function __construct(array $data = [], ?string $type = null)
```
franmomu, line 52: All the public fields should be in the constructor, see

https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/Mapping/Annotation/Timestampable.php#L26-L48

franmomu (commented):
thanks for starting this PR!

tacman:
Thanks, I've fixed that and pushed a change.

My tests are failing, but they seem to be related to Mongo.  I don't have
mongo (or the driver) installed locally, but the github actions failure
seem to have nothing to do with my changes.

tacman:
I'm stuck. :-(

It _appears_ that the failures have something to do with mongodb.

src/Mapping/Annotation/Slug.php
```diff
@@ -41,4 +44,38 @@ final class Slug extends Annotation
...
+
+    /**
+     * @phpstan-param class-string|null $type
+     */
+    public function __construct(array $data = [],
+                                array $fields = [],
+                                bool $updatable = true,
+                                string $style = 'default', // or "camel"
+                                bool $unique = true,
+                                string $unique_base = '',
```
franmomu, line 56:
```suggestion
                                ?string $unique_base = null,
```

franmomu:
> It _appears_ that the failures have something to do with mongodb.

Looking at the errors:

https://github.com/doctrine-extensions/DoctrineExtensions/runs/4459312433?check_suite_focus=true#step:8:34

it is not only mongodb, and looking at the line

https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/Sluggable/SluggableListener.php#L431

something was not right with the `unique_base` (the phpdoc is not right, it should be `string|null`)

franmomu (changes requested):
thanks @tacman! but there is still some work needed, basically testing.

We need to add support for these attributes in our models like:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/05e0780273389764dc6dcebfcea34bbac84ff6f0/src/Tree/Traits/NestedSetEntity.php#L23-L29

And then, for testing purposes we have to modify the tests in https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/tests/Gedmo/Tree and replace the calls:

`$this->getMockSqliteEntityManager($evm)` -> `$this->getDefaultMockSqliteEntityManager($evm)`
`$this->getMockDocumentManager($evm)` -> `$this->getDefaultDocumentManager($evm)`

This will use AttributeDriver when using PHP >= 8

And for fully test it we should add attributes to the fixtures: https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/tests/Gedmo/Tree/Fixture. This is quite some work that maybe could be done using [rector](https://getrector.org/blog/how-to-upgrade-annotations-to-attributes) (not removing the existing annotation), but I haven't tried it (so far, I've always done it by hand because I was also adding types).

.github/workflows/coding-standards.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - support-attributes
```
franmomu, line 7: Please remove this line

tacman:
Sure.  Question -- how do I trigger github actions on my branch?  Or should
I be making changes on my fork's main branch?

franmomu:
> Sure. Question -- how do I trigger github actions on my branch? Or should I be making changes on my fork's main branch?

hmmm I don't know if you can see the actions here: https://github.com/tacman/DoctrineExtensions/actions or you can create also a PR in you fork

franmomu:
> > Sure. Question -- how do I trigger github actions on my branch? Or should I be making changes on my fork's main branch?
> 
> hmmm I don't know if you can see the actions here: https://github.com/tacman/DoctrineExtensions/actions or you can also create a PR in you fork

franmomu:
@tacman if you rebase with the `main` branch, you _only_ have to add the attributes of the Tree extension to tests, the mapping is already done.

franmomu:
@tacman if you give permissions to push in your repository I'll try to work on this

franmomu:
Thanks @tacman for starting this PR! I've used your code and added the missing parts in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2388

PR #2368, opened 2021-12-07
Fix using Loggable extension with only attributes

franmomu:
I missed implementing `AttributeDriverInterface`. 

I've also marked `Attribute` as internal as the other existing `Attribute` classes. I know it has been released, but IMO it has been there just a couple of days and I don't think someone would have used directly the `Attribute` class from the `Loggable` extension.


phansys:
Thank you @franmomu!

Issue #2367, opened 2021-12-07
Timestampable on="change" field="integerValue" broken in 3.4.0


holtkamp:
I did not have time to try to reproduce it but it seems the following configuration broke when upgrading from `3.3.1` => `3.4.0`:

```xml
<field name="someTimestampPopulatedWhenFieldChangesTo10" type="datetime_immutable" precision="6" column="some_timestamp_populated_when_field_changes_to_10" nullable="true">
    <gedmo:timestampable on="change" field="state" value="10"/>
</field>
```

Note that the "state" field of the involved entity is an integer. 

The column `some_timestamp_populated_when_field_changes_to_10` is not populated anymore when the `state` field changes to 10. 

Maybe this is because the integer `10` does not equal the configured value `"10"` anymore...

Reverting back to `3.3.1` fixed it. 

Will try to pinpoint the cause, but wanted to report early to warn others.

franmomu:
Based on the description looks like it could be something related to https://github.com/doctrine-extensions/DoctrineExtensions/commit/d573d269c8b3eef72553a0011d377c4eb690d682, the only thing I see at first sight is this line:

https://github.com/doctrine-extensions/DoctrineExtensions/commit/d573d269c8b3eef72553a0011d377c4eb690d682#diff-25844f6a0588b21dbad934d2925e310a4977be5595f9db4fcd29ef0dae538ddeR140

can you please try to see if it's that one?

holtkamp:
Thanks @franmomu for the quick follow up, that strict comparison of `in_array()` seems as a potential cause indeed.

Will try to come up with a failing test tomorrow/soon to actually "prove" it.

holtkamp:
Ok, I can confirm that the strict comparison of `in_array()` in line https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/AbstractTrackingListener.php#L140 causes the problem with not detecting field changes to a specific integer value when using XML annotations on `3.4.0`.

Disabling the strict comparison causes the change of the field to an integer value like `10` work again.

The root cause seems to be the use of XML as annotations format in which it is not possible to indicate that the 'value' is an integer: 

```XML
<gedmo:timestampable on="change" field="state" value="10"/>
```

Note that removing the double quotes for the value attribute results in an invalid mapping file: 

```XML
<gedmo:timestampable on="change" field="state" value=10/>
```

**Test**
[...]

PR #2369, opened 2021-12-08
Detect and respect integer values for onChange

holtkamp:
As suggested in https://github.com/doctrine-extensions/DoctrineExtensions/issues/2367#issuecomment-988683978


phansys:
I'm fine with this change if you are able to cover all the possible cases and (ideally) delegating the type casting responsibility to Doctrine:
* https://github.com/doctrine/dbal/blob/2aec3f8da5f601462162d9b90d013554b240292b/src/Types/Types.php#L12-L36
* https://github.com/doctrine/mongodb-odm/blob/a520aaec89e7c12537f7bc56942976f35feaa4d4/lib/Doctrine/ODM/MongoDB/Types/Type.php#L27-L54

holtkamp:
@phansys thanks for the input. Not sure whether I am up to this task, but first things first: when taking this path, probably the XML drivers of all extensions should be checked, right?

A [quick search](https://github.com/doctrine-extensions/DoctrineExtensions/blob/daebbcf27d909b5b1e8d394ec0e12bd181dcae85/src/Blameable/Mapping/Driver/Xml.php) gives me: 

- Blameable
- IpTraceable
- Timestampeable

Which seems reasonable... Do you think these 3 are all extensions that are affected?

phansys:
Thank you for the analysis @holtkamp.

>Not sure whether I am up to this task

I'm sure you're up for this task. We are here to bring support if you need some help.

>first things first: when taking this path, probably the XML drivers of all extensions should be checked, right?

Yes, I think all the extensions hitting the same issue should share the same fix. Even if some case is missed in this iteration, we could create a new PR for border cases later.

holtkamp:
As mentioned in https://github.com/doctrine-extensions/DoctrineExtensions/issues/2367#issuecomment-992601242 I still think it would be best to resolve https://github.com/doctrine-extensions/DoctrineExtensions/issues/2367 by making the mapping driver detect integers, but did not have any time / focus to work on this...

Issue #58, opened 2011-05-26
[Tree - Closure] Issue enabling APC cache 


comfortablynumb:
Hi,

Using extensions from master branch (updated a week  ago or so) and enabling APC Cache I get the following exception, which is not thrown if I disable it:

[Semantical Error] line 0, col 103 near 'descendant WHERE': Error: Class Entity\EntityClosure has no association named descendant

Am I doing something wrong here? or maybe I'm missing something?

Thanks in advance.

PR #2370, opened 2021-12-09
Remove unused cache

franmomu:
I was taking a look at not using cache from metadata factory and bump into this.

It was added https://github.com/doctrine-extensions/DoctrineExtensions/commit/56e4f69f5a7dcd0bace35332cbce4b95e9207ba3 because of https://github.com/doctrine-extensions/DoctrineExtensions/issues/58. I guess that was added because at that time the key for storing metadata was created like that based on:

https://github.com/doctrine/orm/blob/b67d400e940b7bd0c99e37b2f6112f8dc1ac7710/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L165-L174

The way the key is generated has been changed lately [when supporting PRS-6](https://github.com/doctrine/persistence/pull/144) because the cache key can't contain `\`, so not sure if this is needed anymore.

I'll try to see if I can find out if this is really needed, but since I don't use the extension any comment is appreciated.

### DIFF
--- a/src/Tree/Strategy/ORM/Closure.php
+++ b/src/Tree/Strategy/ORM/Closure.php
@@ -9,7 +9,6 @@
 
 namespace Gedmo\Tree\Strategy\ORM;
 
-use Doctrine\Common\Cache\Cache;
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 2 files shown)


franmomu:
Closing in favor of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2390

PR #2372, opened 2021-12-11
Fix type handling for the tracked field values configured in the timestampable field

### LINKED ISSUE #2367: Timestampable on="change" field="integerValue" broken in 3.4.0

phansys:
Fixes #2367.

**TODO**:
- [x] Cover the case where the configured value is an array.

Even if these changes could work, I'll be trying to get this solved from the mapping instead, as I think that will be the proper fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,10 @@ a release.
 ## [Unreleased]
 ### Added
 - SoftDeleteable: Support to use annotations as attributes on PHP >= 8.0.
-
-### Added
 - Blameable: Support to use annotations as attributes on PHP >= 8.0.
 
 ### Fixed
+- Blameable, IpTraceable, Timestampable: Type handling for the tracked field values configured in the origin field.
 - Loggable: Using only PHP 8 attributes.
 
 ## [3.4.0] - 2021-12-05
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ a release.
 ## [Unreleased]
 
 ### Fixed
+- Timestampable: Type handling for the tracked field values configured in the timestampable field.
```
franmomu, line 24: this also fixes it in `Blameable` and `IpTraceable` extensions which use `AbstractTracking`.

franmomu (approved):

CHANGELOG.md
phansys: Good catch.

franmomu:
Thanks @phansys!

Issue #2365, opened 2021-12-06
Use our cache pool to store configurations


franmomu:
Right now we are using the cache from `ClassMetadataFactory`:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/Mapping/MappedEventSubscriber.php#L104-L108

https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/Mapping/ExtensionMetadataFactory.php#L113-L120

https://github.com/doctrine-extensions/DoctrineExtensions/blob/81458019f09ae999c4509e670f64594c6d8f7e3c/src/Tree/Strategy/ORM/Closure.php#L159-L163

`AbstractMetadataFactory::getCacheDriver()` [was deprecated](https://github.com/doctrine/persistence/pull/144) some time ago ~, so it is likely that this method is returning `null`~.

Instead of rely on this cache we should use our own cache, the problem is that `MappedEventSubscriber` is the base class of all event listeners so passing an instance of `CacheItemPoolInterface` to `MappedEventSubscriber` would mean that we need to pass it to all listeners.

Any other options?

Issue #2227, opened 2021-05-18
Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheDriver() is deprecated


holtkamp:
As of Doctrine/Persistence 2.2 getCacheDriver was deprecated:
https://github.com/doctrine/persistence/blob/8b420f7065c2d6c0149ff523816f6b28391581d1/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L105-L110

It is used by a number of extensions: https://github.com/doctrine-extensions/DoctrineExtensions/search?q=getCacheDriver

I created this issue to keep track of this deprecation.

A potential solution was already mentioned here: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2226#pullrequestreview-660767646

franmomu:
👍 Fixed in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2373

PR #2373, opened 2021-12-12
Allow to set a cache to store configuration

franmomu:
Fixes https://github.com/doctrine-extensions/DoctrineExtensions/issues/2365, https://github.com/doctrine-extensions/DoctrineExtensions/issues/2227

Instead of relying on the metadata cache from the orm/odm, this PR adds a `MappedEventSubscriber::setCacheItemPool()` method to configure the cache service, in case it is not provided it uses an `ArrayAdapter` (this will invalidate existing caches).

For the bundle, a service can be created (allowing users to configure their own) and call `setCacheItemPool` in every definition as `setAnnotationReader` (https://github.com/stof/StofDoctrineExtensionsBundle/blob/a2bffca41974d1c968557b343e269a60a8d5ffa4/src/Resources/config/blameable.xml#L13-L15)

- [x] There is still one call to `AbstractClassMetadataFactory::getCacheDriver()` ~that we have to figure out what to do~ (https://github.com/doctrine-extensions/DoctrineExtensions/pull/2390).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,9 +38,14 @@ a release.
 - References: Avoid deprecations using LazyCollection with PHP 8.1
 - Tree: Association mapping problems using Closure tree strategy (by manually defining mapping on the closure entity).
 - Wrong PHPDoc type declarations.
+- Avoid calling deprecated `AbstractClassMetadataFactory::getCacheDriver()` method.
 
 ### Deprecated
 - Tree: When using Closure tree strategy, it is deprecated not defining the mapping associations of the closure entity.
+
+### Changed
+- In order to use a custom cache for storing configuration of an extension, the user has to call `setCacheItemPool()`
+  on the extension listener passing an instance of `Psr\Cache\CacheItemPoolInterface`.
 
 ## [3.4.0] - 2021-12-05
 ### Added
--- a/src/Mapping/ExtensionMetadataFactory.php
+++ b/src/Mapping/ExtensionMetadataFactory.php
@@ -10,7 +10,6 @@
 namespace Gedmo\Mapping;
 
 use Doctrine\Bundle\DoctrineBundle\Mapping\MappingDriver as DoctrineBundleMappingDriver;
-use Doctrine\Common\Cache\Cache;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\Mapping\Driver\DefaultFileLocator;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
@@ -22,6 +21,7 @@ use Gedmo\Mapping\Driver\AttributeAnnotationReader;
 use Gedmo\Mapping\Driver\AttributeDriverInterface;
 use Gedmo\Mapping\Driver\AttributeReader;
 use Gedmo\Mapping\Driver\File as FileDriver;
+use Psr\Cache\CacheItemPoolInterface;
 
 /**
  * The extension metadata factory is responsible for extension driver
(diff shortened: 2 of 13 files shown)


src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -201,4 +194,18 @@ protected function getDriver($omDriver)
 
         return $driver;
     }
+
+    private function storeConfiguration(string $className, array $config): void
+    {
+        // Cache the result, even if it's empty, to prevent re-parsing non-existent annotations.
+        $cacheId = self::getCacheId($className, $this->extensionNamespace);
+
+        if (null === $this->cachePoolItem) {
```
mbabker, line 203: Move this up?  No point in computing a cache ID if the cache pool isn't set.

franmomu: thanks! in a previous version I was setting the configuration to the metadata cache when `$this->cachePoolItem` was null I forgot to move it again.

CHANGELOG.md
```diff
@@ -36,10 +36,15 @@ a release.
 - References: Avoid deprecations using LazyCollection with PHP 8.1
 - Tree: Association mapping problems using Closure tree strategy (by manually defining mapping on the closure entity).
 - Wrong PHPDoc type declarations.
+- Avoid calling deprecated `AbstractClassMetadataFactory::getCacheDriver()` method.
 
 ### Deprecated
 - Tree: When using Closure tree strategy, it is deprecated not defining the mapping associations of the closure entity.
 
+### Changed
+- In order to use a custom cache for storing configuration of an extension, the user has to call `setCacheItemPool`
```
phansys, line 45:
```suggestion
- In order to use a custom cache for storing configuration of an extension, the user has to call `setCacheItemPool()`
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -60,18 +60,18 @@ class ExtensionMetadataFactory
...
 
     /**
-     * Initializes extension driver
-     *
-     * @param string $extensionNamespace
-     * @param object $annotationReader
+     * @var CacheItemPoolInterface|null
      */
-    public function __construct(ObjectManager $objectManager, $extensionNamespace, $annotationReader)
+    private $cachePoolItem;
```
phansys, line 65: Just for consistency:
```suggestion
    private $cacheItemPool;
```

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -60,18 +60,18 @@ class ExtensionMetadataFactory
...
+
+    public function __construct(ObjectManager $objectManager, string $extensionNamespace, object $annotationReader, ?CacheItemPoolInterface $cachePoolItem = null)
```
phansys, line 67:
```suggestion
    public function __construct(ObjectManager $objectManager, string $extensionNamespace, object $annotationReader, ?CacheItemPoolInterface $cacheItemPool = null)
```

src/Mapping/MappedEventSubscriber.php
```diff
@@ -269,4 +282,13 @@ private function getDefaultAnnotationReader(): Reader
 
         return self::$defaultAnnotationReader;
     }
+
+    private function getCacheItemPool(): CacheItemPoolInterface
+    {
+        if (null === $this->cacheItemPool) {
+            $this->cacheItemPool = new ArrayAdapter();
```
phansys, line 289: Maybe we could make this assignment by default at `__construct()`. Is there a performance gain making it lazy here?

phansys (approved):

phansys:
Thanks @franmomu!

PR #2378, opened 2021-12-15
sluggable: Add mapping attributes to tests

franmomu:
this PR adds mapping attributes to sluggable tests to when adding the ability of using Slug as attribute it would be easy to review


phansys:
Thanks @franmomu!

PR #2379, opened 2021-12-16
Allow to use Slug as attribute

franmomu:
Part of https://github.com/doctrine-extensions/DoctrineExtensions/issues/2363

This is a tricky one, because `SlugHandler` and `SlugHandlerOption` are nested annotations, so what this PR does is to allow `SlugHandler` as independent (repeatable) attribute and its options are an array of key-value.

This means that this previous code:

```php
    /**
     * @Gedmo\Slug(handlers={
     *      @Gedmo\SlugHandler(class="Gedmo\Sluggable\Handler\TreeSlugHandler", options={
     *          @Gedmo\SlugHandlerOption(name="parentRelationField", value="parent"),
     *          @Gedmo\SlugHandlerOption(name="separator", value="/")
     *      }),
     *      @Gedmo\SlugHandler(class="Gedmo\Sluggable\Handler\InversedRelativeSlugHandler", options={
     *          @Gedmo\SlugHandlerOption(name="relationClass", value="Gedmo\Tests\Sluggable\Fixture\Handler\People\Person"),
     *          @Gedmo\SlugHandlerOption(name="mappedBy", value="occupation"),
     *          @Gedmo\SlugHandlerOption(name="inverseSlugField", value="slug")
     *      })
     * }, fields={"title"})
     * @ORM\Column(length=64, unique=true)
     */
    private ?string $slug;
```

would look like:

```php
    #[Gedmo\Slug(fields: ['title'])]
    #[Gedmo\SlugHandler(class: TreeSlugHandler::class, options: [
        'parentRelationField' => 'parent', 
        'separator' => '/',
    ])]
    #[Gedmo\SlugHandler(class: InversedRelativeSlugHandler::class, options: [
        'relationClass' => Person::class, 
        'mappedBy' => 'occupation', 
        'inverseSlugField' => 'slug',
    ])]
    #[ORM\Column(length: 64, unique: true)]
    private ?string $slug;
```

I think this is the easiest way to solve it, but I'm open to other options.


src/Mapping/Annotation/Slug.php
```diff
@@ -41,4 +45,40 @@ final class Slug extends Annotation
...
+     * @param string[]      $fields
+     * @param SlugHandler[] $handlers
+     */
+    public function __construct(
+        array $data = [],
+        array $fields = [],
+        bool $updatable = true,
+        string $style = 'default',
+        bool $unique = true,
+        ?string $unique_base = null,
```
phansys, line 59: I'd try to use camel case where possible.

src/Sluggable/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,58 @@
...
+        $attributeHandlers = $this->reader->getPropertyAnnotation($property, self::HANDLER);
+
+        if (null === $attributeHandlers) {
+            return [];
+        }
+
+        $handlers = [];
+
+        foreach ($attributeHandlers as $handler) {
+            if (!strlen($handler->class)) {
```
phansys, line 42: Could we use a stricter check here?

franmomu: sure!

src/Mapping/Annotation/Slug.php
franmomu: Here the problem is that this name is used also for annotations (as name parameter), so we need to keep it for now. 

I agree about using camel case, but that can be done in another PR deprecating the snake case.

src/Sluggable/Mapping/Driver/Annotation.php
```diff
@@ -86,91 +86,114 @@ public function readExtendedMetadata($meta, array &$config)
...
+        $handlers = [];
+
+        foreach ($slug->handlers as $handler) {
+            if (!$handler instanceof SlugHandler) {
+                throw new InvalidMappingException("SlugHandler: {$handler} should be instance of SlugHandler annotation in entity - {$meta->getName()}");
             }
-            // process slug fields
-            if (empty($slug->fields) || !is_array($slug->fields)) {
-                throw new InvalidMappingException("Slug must contain at least one field for slug generation in class - {$meta->getName()}");
+            if (!strlen($handler->class)) {
```
phansys, line 105:
```suggestion
            if (!class_exists($handler->class)) {
```

src/Sluggable/Mapping/Driver/Annotation.php
```diff
@@ -86,91 +86,114 @@ public function readExtendedMetadata($meta, array &$config)
...
-                    throw new InvalidMappingException("Unable to find slug [{$slugFieldWithPrefix}] as mapped property in entity - {$meta->getName()}");
+            $class = $handler->class;
+            $handlers[$class] = [];
+            foreach ((array) $handler->options as $option) {
+                if (!$option instanceof SlugHandlerOption) {
+                    throw new InvalidMappingException("SlugHandlerOption: {$option} should be instance of SlugHandlerOption annotation in entity - {$meta->getName()}");
                 }
-                if (!$this->isValidField($meta, $slugFieldWithPrefix)) {
-                    throw new InvalidMappingException("Cannot use field - [{$slugFieldWithPrefix}] for slug storage, type is not valid and must be 'string' or 'text' in class - {$meta->getName()}");
+                if (!strlen($option->name)) {
```
phansys, line 114: Could we use a stricter comparison?

src/Sluggable/Mapping/Driver/Annotation.php
```diff
@@ -86,91 +86,114 @@ public function readExtendedMetadata($meta, array &$config)
...
+            if (!strlen($handler->class)) {
+                throw new InvalidMappingException("SlugHandler class: {$handler->class} should be a valid class name in entity - {$meta->getName()}");
             }
-            foreach ($slug->fields as $slugField) {
-                $slugFieldWithPrefix = null !== $fieldNamePrefix ? ($fieldNamePrefix.'.'.$slugField) : $slugField;
-                if (!$meta->hasField($slugFieldWithPrefix)) {
-                    throw new InvalidMappingException("Unable to find slug [{$slugFieldWithPrefix}] as mapped property in entity - {$meta->getName()}");
+            $class = $handler->class;
+            $handlers[$class] = [];
+            foreach ((array) $handler->options as $option) {
```
phansys, line 110: Do we have a way to avoid the type casting?

phansys (changes requested):

src/Sluggable/Mapping/Driver/Annotation.php
franmomu: since we there is the `NamedArgumentConstructor` annotation, I think there is no need for the type casting anymore.

phansys (approved):

doc/sluggable.md
```diff
@@ -94,38 +94,51 @@ on how to setup and use the extensions in most optimized way.
...
+## Sluggable mapping:
 
 ### Sluggable annotations:
 
 - **@Gedmo\Mapping\Annotation\Slug** it will use this column to store **slug** generated
 **fields** option must be specified, an array of field names to slug
 
+### Sluggable attributes:
+
+- **\#[Gedmo\Mapping\Annotation\Slug]** it will use this column to store **slug** generated
```
phansys, line 106: I know this text is copied from the previous item, but it doesn't seem reasonable to me. Is there a missing full stop after "generated"? Otherwise, I don't get the message in this sentence.

doc/sluggable.md
```diff
@@ -94,38 +94,51 @@ on how to setup and use the extensions in most optimized way.
...
+**fields** option must be specified, an array of field names to slug
+
+**Note:** the examples showed here are using annotations and attributes for mapping, you should use
```
phansys, line 109:
```suggestion
**Note:** the examples shown here are using annotations and attributes for mapping, you should use
```

doc/sluggable.md
```diff
@@ -645,7 +737,7 @@ is a company and user name is **Gedi**. In this case relation has a path separat
 
 User entity example:
 
-``` php
+```php
 <?php
```
phansys, line 741:
```suggestion
<?php

```

doc/sluggable.md
```diff
@@ -720,7 +812,7 @@ class User
 
 Company entity example:
 
-``` php
+```php
 <?php
```
phansys, line 816:
```suggestion
<?php

```

phansys (changes requested):

src/Sluggable/Mapping/Driver/Annotation.php
```diff
@@ -86,91 +86,114 @@ public function readExtendedMetadata($meta, array &$config)
...
+            throw new InvalidMappingException("Unable to find slug [{$fieldName}] as mapped property in entity - {$meta->getName()}");
+        }
+        if (!$this->isValidField($meta, $fieldName)) {
+            throw new InvalidMappingException("Cannot use field - [{$fieldName}] for slug storage, type is not valid and must be 'string' or 'text' in class - {$meta->getName()}");
+        }
+        // process slug handlers
+        $handlers = $this->getSlugHandlers($property, $slug, $meta);
+
+        // process slug fields
+        if (empty($slug->fields) || !is_array($slug->fields)) {
```
phansys, line 151:
```suggestion
        if ([] === $slug->fields || !is_array($slug->fields)) {
```

phansys (approved):

PR #2392, opened 2021-12-30
Fix some PHPDoc types

franmomu:
I found that I added some wrong phpdoc in the annotation classes (not released yet) and ended up modifying a bit more.

Maybe we should create a baseline for PHPStan level 5 or 6 to catch these things and eventually we will be removing errors from the baseline.

### DIFF
--- a/src/Mapping/ExtensionMetadataFactory.php
+++ b/src/Mapping/ExtensionMetadataFactory.php
@@ -11,6 +11,7 @@ namespace Gedmo\Mapping;
 
 use Doctrine\Bundle\DoctrineBundle\Mapping\MappingDriver as DoctrineBundleMappingDriver;
 use Doctrine\Common\Cache\Cache;
+use Doctrine\Persistence\Mapping\ClassMetadata;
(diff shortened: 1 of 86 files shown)


src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -154,7 +155,7 @@ protected function getDriver($omDriver)
         $driver = null;
         $className = get_class($omDriver);
         $driverName = substr($className, strrpos($className, '\\') + 1);
-        if ($omDriver instanceof MappingDriverChain || 'DriverChain' === $driverName) {
```
franmomu, line 157: `DriverChain`: https://github.com/doctrine/orm/blob/40d1e7bbfc62df08406c91b12c6107a815a8d3cf/lib/Doctrine/ORM/Mapping/Driver/DriverChain.php implements `MappingDriverChain`.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2394, opened 2021-12-31
Add missing is_bundle tag in mapping configuration

pietereggink:
Add missing is_bundle tag in mapping configuration

### DIFF
--- a/doc/symfony4.md
+++ b/doc/symfony4.md
@@ -54,6 +54,7 @@ doctrine:
 # only these lines are added additionally
         mappings:
             translatable:
+                is_bundle: false
                 type: annotation # or attribute
                 alias: Gedmo
                 prefix: Gedmo\Translatable\Entity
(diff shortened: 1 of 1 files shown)


franmomu:
Thanks for the PR! is it necessary to specify this option? 

I remember using that option when Symfony encouraged people to organize code in bundles (Symfony < 4) and the mapping was outside a bundle, but since Symfony 4 I guessed that is not used anymore.

Not related to this PR, but just saw that we are using `kernel.root_dir` and we should change it to `kernel.project_dir`.

pietereggink:
> Thanks for the PR! is it necessary to specify this option?
> 
> I remember using that option when Symfony encouraged people to organize code in bundles (Symfony < 4) and the mapping was outside a bundle, but since Symfony 4 I guessed that is not used anymore.
> 
> Not related to this PR, but just saw that we are using `kernel.root_dir` and we should change it to `kernel.project_dir`.

You are right!

I was getting the following error:

Bundle "loggable" does not exist or it is not enabled.  

But this has nothing todo with is_bundle, is_bundle is default true and not required to set.

The error came because of setting an incorrect path, the provided example from the documentation was ORM based and not ODM based.

PR #2364, opened 2021-12-06
Add Symfony PHPUnit Bridge

franmomu:
This also removes deprecation warnings

Needs: 
- [x] https://github.com/doctrine-extensions/DoctrineExtensions/pull/2373

PR #2395, opened 2022-01-01
Fix ReferenceIntegrity using annotations

franmomu:
This fixes when using `ReferenceIntegrity` as annotation passing the value directly (extracted from https://github.com/doctrine-extensions/DoctrineExtensions/pull/2364).


phansys (approved):

phansys:
Thanks @franmomu!

Issue #2318, opened 2021-11-17
XML mappings should stop extending the Doctrine mapping file


stof:
Currently, the XML mapping driver relies on adding custom elements inside the Doctrine XML mapping file itself. This relies on the fact that the Doctrine XML schema allows to add any node there. However, this is something that the ORM maintainers want to stop supporting. The 2.10.0 release removed that from the XSD. When I made them aware that `gedmo/doctrine-extensions` was impacted by that BC break, they agreed on reverting it in 2.10.2. However, I suspect that ORM 3.0 will still perform that BC break again.

It would be wise for  `gedmo/doctrine-extensions` to migrate away from this usage (in a backward-compatible way of course), to avoid being totally blocked by the time ORM 3.0 comes out.

stof:
I see 2 solutions there:

- deprecate support for XML, pushing the extension mapping to be done with annotations all the time (or later PHP attributes once support is added)
- define a standalone XML mapping format that puts the extension mapping into a dedicated XML file separate from the ORM mapping (no idea about how this file would look like for now)

In any case, there needs to be a way to opt-in for the new driver behavior (as the drivers are currently inferred from the ORM driver being used)

stof:
And while this is done, it probably makes sense to do the same for the Yaml driver (but there, as the ORM Yaml driver is deprecated, it might make sense to go for just deprecating the Yaml driver too)

stof:
As the ORM has added a flat to turn on XSD validation in their 2.14 release and has deprecated the fact of keeping it disabled, this issue becomes much more relevant, as the current way to put the extension mapping inside the XML file of the ORM mapping would then fail with a XSD validation error.

eisberg:
> copy this comment from https://github.com/doctrine-extensions/DoctrineExtensions/issues/2613#issuecomment-2188009498

Many projects use XML for ORM, and there are no plans to translate this into attributes. In the PR (https://github.com/doctrine-extensions/DoctrineExtensions/pull/2657) I proposed, there are two options for setting up Gedmo:

1. Use separate XML files for Gedmo ( MyEntity.gedmo.xml )
2. Use attributes, even if the project uses XML (or annotations) for the ORM.
Without these options, GEDMO takes Doctrine settings, and as a result, it is impossible to use attributes for GEDMO if Doctrine uses xml

stof:
[...]

PR #2396, opened 2022-01-01
Deprecate Yaml mapping

franmomu:
ORM Yaml driver was deprecated long time ago and for MongoDB it was already removed for the version we are currently supporting.

[So it makes sense to deprecate Yaml mapping](https://github.com/doctrine-extensions/DoctrineExtensions/issues/2318#issuecomment-971740588).

I've only added the `@deprecated` annotation to the drivers since I guess the user is already receiving deprecation warnings from the ORM, but let me know if you think we should also trigger another deprecation warning.

I'll wait to https://github.com/doctrine-extensions/DoctrineExtensions/pull/2364 get merged so we can see where no add the `@group legacy` annotation in tests.


CHANGELOG.md
```diff
@@ -47,6 +47,9 @@ a release.
 - In order to use a custom cache for storing configuration of an extension, the user has to call `setCacheItemPool()`
   on the extension listener passing an instance of `Psr\Cache\CacheItemPoolInterface`.
 
+### Deprecated
+- Using YAML mapping is deprecated, you SHOULD migrate to attributes or annotations.
```
phansys, line 51: IMO, as we aren't deprecating the XML mapping yet, we should also add this option to the alternatives list.

phansys (approved):

phansys:
Thanks @franmomu!

PR #2397, opened 2022-01-01
Deprecate `QueryAnalyzer`

franmomu:
`QueryAnalyzer` is only used for testing purposes and it implements the deprecated (in `doctrine/dbal` 3) `Doctrine\DBAL\Logging\SQLLogger`.

This PR deprecates the one in `src` and creates a basic `QueryAnalyzer` for testing purposes under the `tests` namespace that won't be necessary when dropping support of `doctrine/dbal` 2.

I've tried to implement `LoggerInterface` in the new `QueryAnalyzer` to try to use it with the new `Middleware` way of logging, but the messages are different in version 2 and 3, so I think it's not worth it and I used `if-else` structures for supporting both.

In `SortableGroupTest` the queries were logged, but they weren't used.

Fixes https://github.com/doctrine-extensions/DoctrineExtensions/pull/2233 by deprecating the class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ a release.
 
 ### Deprecated
 - Tree: When using Closure tree strategy, it is deprecated not defining the mapping associations of the closure entity.
+- `Gedmo\Tool\Logging\DBAL\QueryAnalizer` class without replacement.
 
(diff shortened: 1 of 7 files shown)


phansys:
Thank you @franmomu!

PR #2399, opened 2022-01-02
Add index for translation search

franmomu:
supersedes https://github.com/doctrine-extensions/DoctrineExtensions/pull/2144

The idea is that in 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/adcedc11c6dd9d419e01f84dd95fc3c599137722/src/Translatable/Entity/Repository/TranslationRepository.php#L141-L144

only `foreignKey` and `objectClass` are used in the `WHERE` part and there was no index for that searches.


phansys (approved):

phansys:
Thanks @franmomu and @cbclicer!

PR #2402, opened 2022-01-04
Improve phpdoc by using a more specific types

franmomu:
This solves some PHPStan errors in level 2.

### DIFF
--- a/src/Loggable/LoggableListener.php
+++ b/src/Loggable/LoggableListener.php
@@ -10,6 +10,7 @@
 namespace Gedmo\Loggable;
 
 use Doctrine\Common\EventArgs;
+use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
 use Gedmo\Loggable\Mapping\Event\LoggableAdapter;
 use Gedmo\Mapping\MappedEventSubscriber;
 use Gedmo\Tool\Wrapper\AbstractWrapper;
@@ -99,6 +100,8 @@ class LoggableListener extends MappedEventSubscriber
 
     /**
      * Maps additional metadata
+     *
+     * @param LoadClassMetadataEventArgs $eventArgs
      *
      * @return void
      */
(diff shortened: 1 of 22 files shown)


src/Loggable/LoggableListener.php
```diff
@@ -100,6 +101,8 @@ public function getSubscribedEvents()
     /**
      * Maps additional metadata
      *
+     * @param LoadClassMetadataEventArgs $eventArgs
```
phansys, line 104: I think we should not support this type here, as it is breaking the contravariance.

franmomu: What do you mean?

The thing here is that is actually receiving a `LoadClassMetadataEventArgs` instance, since we are still supporting PHP 7.2 (with partial support of contravariance) we cannot change it in the method signature.

phansys: I mean this situation is not allowed even in PHP 7.2, where a warning is emitted if we could replace `EventArgs` by `LoadClassMetadataEventArgs`. In PHP 8, it results in a fatal error: https://3v4l.org/bvZdv.
I think we should work in the method's body instead, avoiding calls for methods not defined in `EventArgs` or adding a call to `assert($eventArgs instanceof LoadClassMetadataEventArgs)`.
IMO, those are the only ways we have to avoid breaking the contravariance in a safer way, without bringing a wrong message through the docblocks: https://phpstan.org/r/5fa3b930-245b-461e-9077-a38a625c9b6b.

franmomu: I think here the case is different, in these classes the `loadClassMetadata` method is not inherited. It is more this case: https://3v4l.org/E3vU8.

Anyway, I could add `assert($eventArgs instanceof LoadClassMetadataEventArgs)`.

phansys: Forget about the comment then. I thought these methods were overriding this one:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/bcc142745aa97a77d58af591a72d4341b2615ac7/src/AbstractTrackingListener.php#L53-L57

but checking the ancestors, I see there is no inheritance for this.
Thank you for the clarification.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2407, opened 2022-01-08
Use Doctrine MetadataCache if a cache is not set

franmomu:
I'm a bit concern about performance after https://github.com/doctrine-extensions/DoctrineExtensions/pull/2373 because before that, the cache was shared among the listeners and now it should be set externally.

I've created https://github.com/stof/StofDoctrineExtensionsBundle/pull/436 to try to solve it for the bundle.

In the meantime, instead of creating a new `ArrayAdapter` as fallback, I think we can go back to doctrine metadata cache so everything should work as before.

~I'll try later today to add a test and~ after this I'm fine releasing `3.5.0`.

### DIFF
--- a/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
+++ b/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Tests\Mapping;
+
+use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\Common\EventManager;
+use Doctrine\Deprecations\Deprecation;
+use Doctrine\Deprecations\PHPUnit\VerifyDeprecations;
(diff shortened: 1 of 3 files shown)


tests/Gedmo/Mapping/MappingEventSubscriberTest.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        // doctrine/persistence changed from trigger_error to doctrine/deprecations in 2.2.1. In 2.2.2 this trait was
+        // added, this is used to know if the doctrine/persistence version is using trigger_error or
+        // doctrine/deprecations. This "if" check can be removed once we drop support for doctrine/persistence < 2.2.1
+        if (trait_exists(TypedNoDefaultReflectionPropertyBase::class)) {
+            Deprecation::enableWithTriggerError();
+
+            $this->expectDeprecationWithIdentifier('https://github.com/doctrine/persistence/issues/184');
+        } else {
+            $this->expectDeprecation('Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheDriver is deprecated. Use getCache() instead.');
```
stof, line 63: why not use `getCache` instead of the deprecated API when it is available ?

franmomu: because `AbstractClassMetadataFactory::getCache()`is protected: https://github.com/doctrine/persistence/blob/f8af155c1e7963f3d2b4415097d55757bbaa53d8/lib/Doctrine/Persistence/Mapping/AbstractClassMetadataFactory.php#L119

phansys (approved):

PR #2409, opened 2022-01-09
Deprecate `AdapterInterface::__call()` method

franmomu:
Apparently (based on tests) we were calling `AdapterInterface::__call()` to transform the `getObject()` call to `getEntity()` or `getDocument()`, instead of that this PR adds a `AdapterInterface::getObject()` method.


src/Mapping/Event/Adapter/ODM.php
```diff
@@ -36,6 +36,11 @@ class ODM implements AdapterInterface
 
     public function __call($method, $args)
     {
+        @trigger_error(sprintf(
+            'Using "%s()" method is deprecated.',
```
phansys, line 40:
```suggestion
            'Using "%s()" method is deprecated since gedmo/doctrine-extensions 3.x and will be removed in version 4.0.',
```

src/Mapping/Event/Adapter/ODM.php
```diff
@@ -86,7 +91,20 @@ public function getObjectManager()
             return $this->dm;
         }
 
-        return $this->__call('getDocumentManager', []);
+        if (null === $this->args) {
+            throw new RuntimeException(sprintf('Event args must be set before calling "%s()"', __METHOD__));
```
phansys, line 95:
```suggestion
            throw new RuntimeException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
```
BTW, it seems more like a logic exception than a runtime one to me.

phansys (approved):

phansys:
Thanks @franmomu!

PR #2410, opened 2022-01-10
Deprecate `AbstractWrapper::clear()` and `WrapperInterface::populate()` methods

franmomu:
`AbstractWrapper::clear()` method was reseting `$wrappedObjectReferences` which was an unused private variable and `WrapperInterface::populate()` method is was only used in tests.

In another PR I think we can deprecate the whole `AbstractWrapper` and use a `FactoryWrapper` with the `wrap` method.


phansys (approved):

phansys:
Thank you @franmomu!

PR #2414, opened 2022-01-12
[Uploadable] Fix `FileInfoInterface::getSize()` return type declaration

### LINKED ISSUE #2413: Possibly wrong typehint on UploadedFileInfo
The other side of issue https://github.com/antishov/StofDoctrineExtensionsBundle/issues/32

Here https://github.com/doctrine-extensions/DoctrineExtensions/commit/907cd6e11f933ef03093220b93bb48e1b0c25484 FileInfoInterface got some new typehints with the latest release, and I think that the getSize method should return int (or ?int) instead of string|null.

In the docs the interface is used with $_FILES global (https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/doc/uploadable.md#usage) which has a data type of int at the size key.

```php
array(1) {
     ["file"]=> array(5) {
         ["name"]=> string(10) "138728.jpg"
         ["type"]=> string(10) "image/jpeg"
         ["tmp_name"]=> string(14) "/tmp/php1FHOKM"
         ["error"]=> int(0)
         ["size"]=> int(504984)
     }
 } 
```

franmomu:
Fixes #2413

I've added a conversion that is not really needed, but IMO looks better. It is not needed because apparently you can set an `int` to a `string` property (using strict types) through reflection: https://3v4l.org/jLBuI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Fixed
+- Uploadable: `FileInfoInterface::getSize()` return type declaration (#2413).
 
 ## [3.5.0] - 2022-01-10
 ### Added
--- a/src/Uploadable/FileInfo/FileInfoInterface.php
+++ b/src/Uploadable/FileInfo/FileInfoInterface.php
@@ -28,7 +28,7 @@ interface FileInfoInterface
     public function getName();
 
     /**
-     * @return string|null
+     * @return int|null
      */
     public function getSize();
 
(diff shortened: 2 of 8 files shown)


phansys:
>It is not needed because apparently you can set an int to a string property (using strict types) through reflection: https://3v4l.org/jLBuI

I think we could report the issue to [php/php-src](https://github.com/php/php-src).
Even if the reflection API allows to access private properties, I don't know if violating the type declaration should be allowed.

phansys (approved):

phansys:
Thanks @franmomu!

franmomu:
For the record: https://github.com/php/php-src/issues/7936#issuecomment-1011417421

PR #2390, opened 2021-12-27
Make users add mapping to tree closure entities

franmomu:
In https://github.com/doctrine-extensions/DoctrineExtensions/issues/58 one of the solution was to make the user to map the associations manually.

Fixes https://github.com/doctrine-extensions/DoctrineExtensions/issues/2236 and probably https://github.com/doctrine-extensions/DoctrineExtensions/issues/2240

This PR supersedes https://github.com/doctrine-extensions/DoctrineExtensions/pull/2370

PR #2416, opened 2022-01-13
Closure: mirror Doctrine cache key

Slamdunk:
Hi, I had this bug for a while: yesterday I found the issue and the solution, but then I found https://github.com/doctrine-extensions/DoctrineExtensions/pull/2390 that basically wipes the bug when you forward upgrade the code as documented.

Still I think it's worth fixing it: when running `./cli orm:schema-tool:update`, the order in which classes are loaded into metadatas changes the outcome of the cache. In particluar, [this file iterator](https://github.com/doctrine/persistence/blob/f8af155c1e7963f3d2b4415097d55757bbaa53d8/lib/Doctrine/Persistence/Mapping/Driver/AnnotationDriver.php#L211-L218) could read the Closure class before or after the reference class.

If the Closure class is read first, than the reference class loads `ancestor` and `descendant` properties only afterward, which must be updated in the cache too, but the key is different so it gets stored in a cache key that will never be read/used.

Using explicit fields like pointed out in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2390 fixes the issue, but still it's better to also fix this until `v4` is out.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/persistence": "^1.3.3 || ^2.0",
+        "doctrine/persistence": "^2.0",
         "psr/cache": "^1 || ^2 || ^3",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/persistence": "^2.0",
+        "doctrine/persistence": "^2.2",
         "psr/cache": "^1 || ^2 || ^3",
         "symfony/cache": "^4.4 || ^5.3 || ^6.0"
     },
--- a/src/Tree/Strategy/ORM/Closure.php
+++ b/src/Tree/Strategy/ORM/Closure.php
@@ -201,10 +201,6 @@ class Closure implements Strategy
                 /* @see https://github.com/doctrine/persistence/pull/144 */
                 /* @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey */
                 $cacheDriver->save(
-                    $closureMetadata->getName().'$CLASSMETADATA',
-                    $closureMetadata
-                );
-                $cacheDriver->save(
                     str_replace('\\', '__', $closureMetadata->getName()).'__CLASSMETADATA__',
                     $closureMetadata
                 );
--- a/src/Tree/Strategy/ORM/Closure.php
+++ b/src/Tree/Strategy/ORM/Closure.php
@@ -198,7 +198,11 @@ class Closure implements Strategy
             $cacheDriver = $cmf->getCacheDriver();
 
             if ($cacheDriver instanceof Cache) {
-                $cacheDriver->save($closureMetadata->getName().'$CLASSMETADATA', $closureMetadata);
+                /* @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey */
+                $cacheDriver->save(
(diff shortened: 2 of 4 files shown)


franmomu:
thanks for the PR! yeah, it make sense to update that code. 

The change of how the key is generated was introduced in https://github.com/doctrine/persistence/pull/144, since we are supporting `"doctrine/persistence": "^1.3.3 || ^2.0"` I think we should only change the key using `doctrine/persistence` >= 2.2.0.

And if you could add an entry in the changelog it would be really awesome.

Slamdunk:
> I think we should only change the key when using `doctrine/persistence` >= 2.2.0.

Ok, and how should we accomplish this goal? Bumping the dep version in the `composer.json`?

franmomu:
hmmm we could maybe check if `method_exists(AbstractClassMetadataFactory::class, 'setCache')` (which was introduced in that PR) then use this new way, otherwise stick to the old one.

Slamdunk:
Mmm, that seems a brittle solution to me: what about setting **both** cache keys anyway?

franmomu:
Can you please add an entry in the changelog? https://github.com/doctrine-extensions/DoctrineExtensions/blob/27f0c13e7b7a6dc1d14ef119df0d58c6dd7fcdce/CHANGELOG.md#fixed

Adding two keys doesn't seem a proper solution to me either, maybe we can just bump to `doctrine/persistence: ^2.2` then, for this package to work it needs either `doctrine/orm` or `doctrine/mongodb-odm` AFAIK.

`doctrine/orm` 2.10.2 (the minimum we require) already requires `doctrine/persistence: ^2.2` and our tests are not installing `doctrine/persistence` 1 anymore because of that.

I don't think there is a problem dropping support for version 1, what do you think @phansys?

Slamdunk:
Done

Slamdunk:
CS Fix fails after rebase: ever though I could easily fix those errors, I prefer not to because the affected code isn't related to this PR

composer.json
```diff
@@ -45,7 +45,7 @@
         "doctrine/collections": "^1.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.0",
-        "doctrine/persistence": "^1.3.3 || ^2.0",
+        "doctrine/persistence": "^2.0",
```
franmomu, line 48:
```suggestion
        "doctrine/persistence": "^2.2",
```
Since it's the version when the key was changed

franmomu (approved):

phansys:
>I don't think there is a problem dropping support for version 1, what do you think @phansys?

I think there is no problem with this change.
BTW, sorry for the delay.

src/Tree/Strategy/ORM/Closure.php
```diff
@@ -198,7 +198,12 @@ public function processMetadataLoad($em, $meta)
             $cacheDriver = $cmf->getCacheDriver();
 
             if ($cacheDriver instanceof Cache) {
-                $cacheDriver->save($closureMetadata->getName().'$CLASSMETADATA', $closureMetadata);
+                /* @see https://github.com/doctrine/persistence/pull/144 */
+                /* @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey */
```
phansys, line 202:
```suggestion
                // @see https://github.com/doctrine/persistence/pull/144
                // @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey()
```

tests/Gedmo/Mapping/TreeMappingTest.php
```diff
@@ -77,15 +77,19 @@ protected function setUp(): void
 
     /**
      * @group legacy
+     *
+     * @see https://github.com/doctrine/persistence/pull/144
+     * @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey
```
phansys, line 82:
```suggestion
     * @see \Doctrine\Persistence\Mapping\AbstractClassMetadataFactory::getCacheKey()
```

phansys:
I'm closing and reopening this PR in order to trigger an up to date build.

phansys (approved):

phansys:
Thank you @Slamdunk.

PR #2417, opened 2022-01-14
Fix phpcs

franmomu:


franmomu:
Closing as intersection type is the right one, we should probably skip  php-cs-fixer `3.5`

PR #2355, opened 2021-12-03
[#2351] Add `@return` docblock in methods implementing `EventSubscriber::getSubscribedEvents()`

phansys:
Thank you for reporting this issue @jokaorgua.
Could you please confirm if these changes are enough to avoid the deprecation message?

Closes #2351.

PR #2422, opened 2022-02-03
Fix Deprecation notice `add "array" as a native return type` on the getSubscribedEvents()

damienfa:
On Symfony 5.4, I have dozens of Deprecation notice.  e.g. : 

`
php.INFO: User Deprecated: Method "Doctrine\Common\EventSubscriber::getSubscribedEvents()" might add "array" as a native return type declaration in the future. Do the same in implementation "Gedmo\Loggable\LoggableListener" now to avoid errors or add an explicit @return annotation to suppress this message. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Method \"Doctrine\\Common\\EventSubscriber::getSubscribedEvents()\" might add \"array\" as a native return type declaration in the future. Do the same in implementation \"Gedmo\\Loggable\\LoggableListener\" now to avoid errors or add an explicit @return annotation to suppress this message. at /app/vendor/symfony/error-handler/DebugClassLoader.php:325)"} []`

This pull request fix these deprecations adding a `: array` to every `getSubscribedEvents()` method declared in the Listeners classes.

### DIFF
--- a/src/AbstractTrackingListener.php
+++ b/src/AbstractTrackingListener.php
@@ -32,9 +32,9 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
     /**
      * Specifies the list of events to listen on.
      *
-     * @return string[]
+     * @return array
      */
-    public function getSubscribedEvents()
+    public function getSubscribedEvents():array
     {
         return [
             'prePersist',
--- a/src/Loggable/LoggableListener.php
+++ b/src/Loggable/LoggableListener.php
@@ -89,7 +89,7 @@ class LoggableListener extends MappedEventSubscriber
     /**
      * @return string[]
      */
-    public function getSubscribedEvents()
+    public function getSubscribedEvents():array
     {
         return [
             'onFlush',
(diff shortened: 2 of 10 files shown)


franmomu:
Hi @damienfa thanks for the PR, but these changes break BC and AFAIK there are not necessary, those deprecations were fixed in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2355, maybe you're using an old version of this package?

damienfa:
@franmomu , ha ok. I understand for the BC.
I'm on the last version (v3.5.0), but I have a lot of these deprecation notices in my logs. I'm on PHP 7.4. 
Next week I will try to find some time to investigate.

franmomu:
@damienfa I think it shouldn't be related to PHP 7.4, do you still get these deprecations?

damienfa:
@franmomu , ok.
 No, I think it's now ok 👍🏽

franmomu:
ok! I'll close the issue then 👍

PR #2425, opened 2022-02-24
Mark Cache dependencies as optional

Slamdunk:
None of the actual Cache usages are required, as they are checked by runtime if exist:
1. https://github.com/doctrine-extensions/DoctrineExtensions/blob/2def628a9a8304c1e7f56847a0b21778a3a9ed61/src/DoctrineExtensions.php#L119-L123
1. https://github.com/doctrine-extensions/DoctrineExtensions/blob/2def628a9a8304c1e7f56847a0b21778a3a9ed61/src/Tree/Strategy/ORM/Closure.php#L197-L200

Marking those as optional allow to get rid of unused (an thus unwanted) dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,6 @@
         "doctrine/event-manager": "^1.2 || ^2.0",
         "doctrine/persistence": "^2.2 || ^3.0",
         "psr/cache": "^1 || ^2 || ^3",
-        "symfony/cache": "^4.4 || ^5.3 || ^6.0",
         "symfony/deprecation-contracts": "^2.1 || ^3.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


franmomu:
There is another usage (it was defined as required because of this):

https://github.com/doctrine-extensions/DoctrineExtensions/blob/ddc468bcf40045a5b1606f62296b75c7c2a6a42c/src/Mapping/MappedEventSubscriber.php#L283-L301

Slamdunk:
You're right: we still can mark `symfony/cache` as optional, and throw an Exception if both are absent.
I'll update the PR as soon as possible

Slamdunk:
Huh, I completely forgot to update this PR :open_mouth:

Slamdunk:
@franmomu rebased and updated as requested

Slamdunk:
Also ping @phansys as I've just seen this comment: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2564#issuecomment-1370155413

phansys:
I guess we can merge this PR before the new release, as it is not narrowing an existing dependency or adding new ones. :+1:

PR #2426, opened 2022-02-28
[Bleameable, Loggable] Update user identifier determination logic

### LINKED ISSUE #2424: BlameableListener not contemplating getUserIdentifier method availability (still getUsername)
I think it would make sense that when the user is an object, the listener should expect / try to use a getUserIdentifier method on it

currently it's not: 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/2def628a9a8304c1e7f56847a0b21778a3a9ed61/src/Blameable/BlameableListener.php#L48-L57

I didn't have a chance at using Blameable myself yet, so I'm not 100% sure if this actually makes sense to you, so, just ignore if I'm wrong...

but, btw, I also noticed Loggable received a similar update, in 396faefb9eef46ebdf56a9b146abae50b66b73f5 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/ca2849656df9f5097a7a92fba2162fdba34c504a/src/Loggable/LoggableListener.php#L76-L87

AlexBevilacqua:
Closes #2424
with the addition of `getUserIdentifier()` support to `BlameableListener`

additionally to that, I imagined it didn't make a lot of sense for this logic to be different between the two
(hope I'm not mistaken.. - or could it be an idea to move it somewhere, re-usable?)

so, in `LoggableListener` I've also ported the support for stringable objects (addition of `__toString` support) already available in  BlameableListener

### DIFF
--- a/src/Blameable/BlameableListener.php
+++ b/src/Blameable/BlameableListener.php
@@ -44,8 +44,11 @@ class BlameableListener extends AbstractTrackingListener
             return $this->user;
         }
 
-        // ok so its not an association, then it is a string
+        // ok so it's not an association, then it is a string, or an object
         if (is_object($this->user)) {
+            if (method_exists($this->user, 'getUserIdentifier')) {
+                return (string) $this->user->getUserIdentifier();
+            }
             if (method_exists($this->user, 'getUsername')) {
                 return (string) $this->user->getUsername();
             }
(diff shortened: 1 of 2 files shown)


franmomu (approved):

franmomu:
Thanks @AlexBevilacqua! 

I think this if fine for now, since this is kind of related to Symfony `UserInterface`, I have to take a better look, but if the listener is storing a `string` we should probably only accept `string` and the framework dependent listener should be the one transforming the user object into `string`.

PR #2436, opened 2022-03-17
Feature/serializer support for traits

johnkhansrc:
Hello

I propose this PR to allow the use of Traits in contexts using Symfony Serializer,
  for example when using API Platform.
  
I made sure that all the tests passed, and I added tests to validate the concept.


franmomu:
Hi @johnkhansrc, thanks for submitting the PR, but I don't think we should add support for these groups. 

IMHO these traits are basic code that you can more or less use, but once you need to change something you should use your own traits. It's always better if your model data does not depend on an external package.

Having said that, IMO we should keep traits simple.

PR #2441, opened 2022-04-18
Add fractional seconds support

samuel4x4:
https://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html

Without this, if you do a GET in the same second you marked an item as soft deleted, you'll still get it as not deleted.


phansys:
@samuel4x4, thank you so much for this feature!
Could you please add tests for this change and a changelog note?

PR #2446, opened 2022-04-29
[Tree] [NestedSet] [ClosureTree] childrenQueryBuilder() to allow specifying sort order separately for each field

pmishev:
When sorting by more than one field, before you could only specify the sort order of the primary field. Now you can specify the order for each field you're sorting on

### DIFF
--- a/src/Tree/Entity/Repository/ClosureTreeRepository.php
+++ b/src/Tree/Entity/Repository/ClosureTreeRepository.php
@@ -101,11 +101,11 @@ class ClosureTreeRepository extends AbstractTreeRepository
     }
 
     /**
-     * @param object|null $node        if null, all tree nodes will be taken
-     * @param bool        $direct      true to take only direct children
-     * @param string      $sortByField field name to sort by
-     * @param string      $direction   sort direction : "ASC" or "DESC"
-     * @param bool        $includeNode Include the root node in results?
+     * @param object|null          $node        If null, all tree nodes will be taken
+     * @param bool                 $direct      True to take only direct children
+     * @param string|string[]|null $sortByField Field name or array of fields names to sort by
+     * @param string|string[]      $direction   Sort order ('ASC'|'DESC'). If $sortByField is an array, this may also be an array with matching number of elements
+     * @param bool                 $includeNode Include the root node in results?
      *
      * @return QueryBuilder QueryBuilder object
      */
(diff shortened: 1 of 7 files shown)


shafan (changes requested):
The param $sortByField of childrenQueryBuilder function must be type of : string|string[]|null

shafan:
For me there is still a problem. NestedTreeRepository extend AbsractTreeRepository which defines the childrenQueryBuilder function with $sortByField as string.
So we have to change the phpDoc of AbstractTreeRepository but also for ClosureTreeRepository.

shafan:
Thank you for your work @pmishev

src/Tree/Entity/Repository/ClosureTreeRepository.php
```diff
@@ -150,7 +150,11 @@ public function childrenQueryBuilder($node = null, $direct = false, $sortByField
         }
 
         if ($sortByField) {
-            if ($meta->hasField($sortByField) && in_array(strtolower($direction), ['asc', 'desc'], true)) {
+            if (is_array($sortByField)) {
+                foreach ($sortByField as $key => $field) {
+                    $qb->addOrderBy('node.'.$field, is_array($direction) ? $direction[$key] : $direction);
```
franmomu, line 155:
```suggestion
                    $qb->addOrderBy('node.'.$field, $direction[$key] ?? $direction);
```
Maybe? to handle possible array size missmatch

pmishev: It wouldn't work exactly like that, because if `$direction` is a string, `$direction[$key]` would return its first letter.

franmomu (approved):

franmomu:
Thanks @pmishev! We can maybe deprecate the second parameter and use an associative array for both.

src/Tree/Entity/Repository/ClosureTreeRepository.php
```diff
@@ -150,10 +150,21 @@ public function childrenQueryBuilder($node = null, $direct = false, $sortByField
         }
 
         if ($sortByField) {
-            if ($meta->hasField($sortByField) && in_array(strtolower($direction), ['asc', 'desc'], true)) {
-                $qb->orderBy('node.'.$sortByField, $direction);
+            if (is_array($sortByField)) {
+                foreach ($sortByField as $key => $field) {
+                    $fieldDirection = strtolower(is_array($direction) ? ($direction[$key] ?? 'asc') : $direction);
+                    if ($meta->hasField($field) && in_array($fieldDirection, ['asc', 'desc'], true)) {
```
hacfi, line 156: This breaks my proejct as the `root` is a reference and not a field. The error I get is `Invalid sort options specified: field - root, direction - ASC`

franmomu: @pmishev can you please take a look at this? (also https://github.com/doctrine-extensions/DoctrineExtensions/issues/2492)

pmishev:
I have already fixed this in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2447/commits/926fcb90b374d5f5a7c8e9d305fc2d4ab20b606c, which I was hoping to be merged as part of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2447, but it looks like it may take some time to be approved.
Perhaps the commit can be cherry-picked.

pmishev:
Actually, I just notice the problem is in ClosureTree as well. Will have a better look.

PR #2447, opened 2022-04-29
[Tree] [NestedSet] New additions

pmishev:
- [Tree] [NestedSet] Added "base" property for tree level annotation
- [Tree] [NestedSet] Added 'includeNode' option in getPathQueryBuilder() to specify whether you want the starting node included or not
- [Tree] [NestedSet] Added getPathAsString() method to entity repository
- [Tree] [NestedSet] Added 'treeRootNode' options in verify() in case you want to verify a single tree in a forest
- [Tree] [NestedSet] Added recoverFast() method for where speed is more important than safety and entity manager state
- [Tree] [NestedSet] Added options to recover() for sibling order, tree root in a forest, verification skip and auto-flushing
- [Tree] [NestedSet] Verify and recover wrong levels in nested set
- [Tree] [NestedSet] Added test for removing parent node from tree

### DIFF
--- a/src/Tree/Entity/Repository/NestedTreeRepository.php
+++ b/src/Tree/Entity/Repository/NestedTreeRepository.php
@@ -9,10 +9,12 @@
 
 namespace Gedmo\Tree\Entity\Repository;
 
+use Doctrine\ORM\Exception\ORMException;
 use Doctrine\ORM\Proxy\Proxy;
 use Doctrine\ORM\Query;
 use Doctrine\ORM\QueryBuilder;
 use Gedmo\Exception\InvalidArgumentException;
+use Gedmo\Exception\RuntimeException;
 use Gedmo\Exception\UnexpectedValueException;
 use Gedmo\Tool\Wrapper\EntityWrapper;
 use Gedmo\Tree\Strategy;
@@ -61,7 +63,7 @@ class NestedTreeRepository extends AbstractTreeRepository
     {
         if ('persistAs' === substr($method, 0, 9)) {
             if (!isset($args[0])) {
-                throw new \Gedmo\Exception\InvalidArgumentException('Node to persist must be available as first argument');
+                throw new InvalidArgumentException('Node to persist must be available as first argument.');
             }
             $node = $args[0];
             $wrapped = new EntityWrapper($node, $this->_em);
@@ -70,7 +72,7 @@ class NestedTreeRepository extends AbstractTreeRepository
             $position = substr($method, 9);
             if ('Of' === substr($method, -2)) {
                 if (!isset($args[1])) {
-                    throw new \Gedmo\Exception\InvalidArgumentException('If "Of" is specified you must provide parent or sibling as the second argument');
+                    throw new InvalidArgumentException('If "Of" is specified you must provide parent or sibling as the second argument.');
                 }
                 $parentOrSibling = $args[1];
                 if (strstr($method, 'Sibling')) {
@@ -134,13 +136,27 @@ class NestedTreeRepository extends AbstractTreeRepository
      * Get the Tree path query builder by given $node
      *
      * @param object $node
+     * @phpstan-param array{includeNode?: bool} $options
      *
-     * @throws InvalidArgumentException if input is not valid
+     * options:
+     * - includeNode: (bool) Whether to include the node itself. Defaults to true.
      *
      * @return QueryBuilder
+     *
+     * @throws InvalidArgumentException if input is not valid
      */
-    public function getPathQueryBuilder($node)
+    public function getPathQueryBuilder($node/* , array $options = [] */) // @phpstan-ignore-line
     {
+        $options = func_get_args()[1] ?? [];
+        if (!\is_array($options)) {
+            throw new \TypeError('Argument 2 MUST be an array.');
+        }
+
+        $defaultOptions = [
+            'includeNode' => true,
+        ];
+        $options += $defaultOptions;
+
         $meta = $this->getClassMetadata();
         if (!is_a($node, $meta->getName())) {
             throw new InvalidArgumentException('Node is not related to this repository');
@@ -155,10 +171,15 @@ class NestedTreeRepository extends AbstractTreeRepository
         $qb = $this->getQueryBuilder();
         $qb->select('node')
             ->from($config['useObjectClass'], 'node')
-            ->where($qb->expr()->lte('node.'.$config['left'], $left))
-            ->andWhere($qb->expr()->gte('node.'.$config['right'], $right))
             ->orderBy('node.'.$config['left'], 'ASC')
         ;
+        if ($options['includeNode']) {
+            $qb->where($qb->expr()->lte('node.'.$config['left'], $left))
+               ->andWhere($qb->expr()->gte('node.'.$config['right'], $right));
+        } else {
+            $qb->where($qb->expr()->lt('node.'.$config['left'], $left))
+               ->andWhere($qb->expr()->gt('node.'.$config['right'], $right));
+        }
         if (isset($config['root'])) {
             $rootId = $wrapped->getPropertyValue($config['root']);
             $qb->andWhere($qb->expr()->eq('node.'.$config['root'], ':rid'));
@@ -172,24 +193,82 @@ class NestedTreeRepository extends AbstractTreeRepository
      * Get the Tree path query by given $node
      *
      * @param object $node
+     * @phpstan-param array{includeNode?: bool} $options
      *
-     * @return \Doctrine\ORM\Query
+     * options:
+     * - includeNode: (bool) Whether to include the node itself. Defaults to true.
+     *
+     * @return Query
      */
-    public function getPathQuery($node)
+    public function getPathQuery($node/* , array $options = [] */) // @phpstan-ignore-line
     {
-        return $this->getPathQueryBuilder($node)->getQuery();
+        $options = func_get_args()[1] ?? [];
+        if (!\is_array($options)) {
+            throw new \TypeError('Argument 2 MUST be an array.');
+        }
+
+        return $this->getPathQueryBuilder($node, $options)->getQuery();
     }
 
     /**
      * Get the Tree path of Nodes by given $node
      *
      * @param object $node
+     * @phpstan-param array{includeNode?: bool} $options
+     *
+     * options:
+     * - includeNode: (bool) Whether to include the node itself. Defaults to true.
      *
      * @return array list of Nodes in path
      */
-    public function getPath($node)
+    public function getPath($node/* , array $options = [] */) // @phpstan-ignore-line
     {
-        return $this->getPathQuery($node)->getResult();
+        $options = func_get_args()[1] ?? [];
+        if (!\is_array($options)) {
+            throw new \TypeError('Argument 2 MUST be an array.');
+        }
+
+        return $this->getPathQuery($node, $options)->getResult();
+    }
+
+    /**
+     * Get the Tree path of Nodes by given $node as a string
+     *
+     * @phpstan-param array{
+     *     includeNode?: bool,
+     *     separator?: string,
+     *     stringMethod?: string
+     * } $options
+     *
+     * options:
+     * - includeNode:  (bool)   Whether to include the node itself. Defaults to true.
+     * - separator:    (string) The string separating the nodes of the tree. Defaults to ' > '.
+     * - stringMethod: (string) Entity method returning its displayable name. Defaults to '__toString'.
+     *
+     * @throws InvalidArgumentException
+     */
+    public function getPathAsString(object $node, array $options = []): string
+    {
+        $defaultOptions = [
+            'includeNode' => true,
+            'separator' => ' > ',
+            'stringMethod' => '__toString',
+        ];
+        $options += $defaultOptions;
+
+        if (!is_string($options['stringMethod'])) {
+            throw new InvalidArgumentException(sprintf('"stringMethod" option passed in argument 2 to %s must be a valid string.', __METHOD__));
+        }
+        if (!method_exists($node, $options['stringMethod'])) {
+            throw new InvalidArgumentException(sprintf('%s must implement method "%s".', get_class($node), $options['stringMethod']));
+        }
+
+        $path = [];
+        foreach ($this->getPath($node, $options) as $pathNode) {
+            $path[] = $pathNode->{$options['stringMethod']}();
+        }
+
+        return implode($options['separator'], $path);
     }
 
     /**
@@ -381,7 +460,7 @@ class NestedTreeRepository extends AbstractTreeRepository
      * @param string $sortByField field name to sort by
      * @param string $direction   sort direction : "ASC" or "DESC"
      *
-     * @return \Doctrine\ORM\Query
+     * @return Query
      */
     public function getLeafsQuery($root = null, $sortByField = null, $direction = 'ASC')
     {
@@ -408,9 +487,9 @@ class NestedTreeRepository extends AbstractTreeRepository
      * @param object $node
      * @param bool   $includeSelf include the node itself
      *
-     * @throws \Gedmo\Exception\InvalidArgumentException if input is invalid
-     *
      * @return QueryBuilder
+     *
+     * @throws InvalidArgumentException if input is invalid
      */
     public function getNextSiblingsQueryBuilder($node, $includeSelf = false)
     {
@@ -462,7 +541,7 @@ class NestedTreeRepository extends AbstractTreeRepository
      * @param object $node
      * @param bool   $includeSelf include the node itself
      *
-     * @return \Doctrine\ORM\Query
+     * @return Query
      */
     public function getNextSiblingsQuery($node, $includeSelf = false)
     {
@@ -488,9 +567,9 @@ class NestedTreeRepository extends AbstractTreeRepository
      * @param object $node
      * @param bool   $includeSelf include the node itself
      *
-     * @throws \Gedmo\Exception\InvalidArgumentException if input is invalid
-     *
      * @return QueryBuilder
+     *
+     * @throws InvalidArgumentException if input is invalid
      */
     public function getPrevSiblingsQueryBuilder($node, $includeSelf = false)
     {
@@ -539,9 +618,9 @@ class NestedTreeRepository extends AbstractTreeRepository
      * @param object $node
      * @param bool   $includeSelf include the node itself
      *
-     * @throws \Gedmo\Exception\InvalidArgumentException if input is invalid
+     * @return Query
      *
-     * @return \Doctrine\ORM\Query
+     * @throws InvalidArgumentException if input is invalid
      */
     public function getPrevSiblingsQuery($node, $includeSelf = false)
     {
@@ -752,7 +831,7 @@ class NestedTreeRepository extends AbstractTreeRepository
                 $this->_em->close();
                 $this->_em->getConnection()->rollback();
 
-                throw new \Gedmo\Exception\RuntimeException('Transaction failed', $e->getCode(), $e);
+                throw new RuntimeException('Transaction failed', $e->getCode(), $e);
             }
         } else {
             throw new InvalidArgumentException('Node is not related to this repository');
@@ -814,10 +893,25 @@ class NestedTreeRepository extends AbstractTreeRepository
      * If any error is detected it will return an array
      * with a list of errors found on tree
      *
+     * @phpstan-param array{treeRootNode?: object} $options
+     *
+     * options:
+     * - treeRootNode: (object) Optional tree root node to verify, if not the whole forest (only available for forests, not for single trees).
+     *
      * @return array|bool true on success,error list on failure
      */
-    public function verify()
+    public function verify(/* array $options = [] */) // @phpstan-ignore-line
     {
+        $options = func_get_args()[0] ?? [];
+        if (!\is_array($options)) {
+            throw new \TypeError('Argument 1 MUST be an array.');
+        }
+
+        $defaultOptions = [
+            'treeRootNode' => null,
+        ];
+        $options += $defaultOptions;
+
         if (!$this->childCount()) {
             return true; // tree is empty
         }
@@ -828,6 +922,10 @@ class NestedTreeRepository extends AbstractTreeRepository
         if (isset($config['root'])) {
             $trees = $this->getRootNodes();
             foreach ($trees as $tree) {
+                // if a root node is specified, verify only it
+                if (null !== $options['treeRootNode'] && $options['treeRootNode'] !== $tree) {
+                    continue;
+                }
                 $this->verifyTree($errors, $tree);
             }
         } else {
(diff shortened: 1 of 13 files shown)


src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -134,13 +134,21 @@ public function getRootNodes($sortByField = null, $direction = 'asc')
      * Get the Tree path query builder by given $node
      *
      * @param object $node
+     * @param array  $options
```
franmomu, line 137:
```suggestion
     * @phpstan-param array{includeNode?: bool}  $options
```

Can we please specify these options with array shapes? so PHPStan understands it

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -800,10 +849,20 @@ public function reorderAll($sortByField = null, $direction = 'ASC', $verify = tr
...
      *
+     * @param array $options
+     *
+     * $options['treeRootNode']  = (object) Optional tree root node to verify, if not the whole forest (default: null)
+     *                             Option is only available for forests, not for single trees
+     *
      * @return array|bool true on success,error list on failure
      */
-    public function verify()
+    public function verify(array $options = [])
```
franmomu, line 859: I'm not using the Tree behaviour, but is it likely to pass more options? if it's just one we could maybe use `$treeRootNode`

pmishev: I agree that in this case a single `$treeRootNode` argument would be better, however I just wanted to keep consistency with other places where I added extra options. For example in the `recover()` below there are a lot optional options and one of them is again `treeRootNode`. 
In the latter case, I'd argue that it's better to have an `$options` array, at least for PHP7 (there are also other precedents in the bundle where such an `$options` array is used), but that is probably no longer the case with the named arguments in PHP8, so maybe I should use separate arguments everywhere? 
On the other hand, in the case of `getPath -> getPathQuery -> getPathQueryBuilder`, it's maybe easier to maintain with an array of options, as if you add another a new option, you don't have to change the signature of all 3 methods. WDYT?

franmomu: Hey @pmishev sorry for the late reply, yeah, I guess it's fine to keep these `$options`, please specify all these options in PHPDoc as `array{option1: type, ...}`

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -134,13 +134,21 @@ public function getRootNodes($sortByField = null, $direction = 'asc')
...
+     * @throws InvalidArgumentException if input is not valid
+     *
      */
-    public function getPathQueryBuilder($node)
+    public function getPathQueryBuilder($node, array $options = [])
     {
+        $defaultOptions = [
+            'includeNode' => true,
+        ];
+        $options = array_merge($defaultOptions, $options);
```
phansys, line 150:
```suggestion
        $options += $defaultOptions;
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +185,60 @@ public function getPathQueryBuilder($node)
...
+     * @param object $node
+     * @param array  $options
+     *
+     * $options['includeNode']  = (bool) Whether to include the node itself (default: true)
+     * $options['separator']    = (string) The string separating the nodes of the tree
+     * $options['stringMethod'] = Entity method returning its displayable name.
+     *                            If not provided, entity must have __toString method
+     * @return string
+     */
+    public function getPathAsString($node, array $options = [])
```
phansys, line 226:
```suggestion
     */
    public function getPathAsString($node, array $options = []): string
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +185,60 @@ public function getPathQueryBuilder($node)
...
+    {
+        $defaultOptions = [
+            'includeNode'  => true,
+            'separator'    => ' > ',
+            'stringMethod' => null,
+        ];
+        $options = array_merge($defaultOptions, $options);
```
phansys, line 233:
```suggestion
        $options += $defaultOptions;
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +185,60 @@ public function getPathQueryBuilder($node)
...
+
+        $path = [];
+        $pathNodes = $this->getPath($node, $options);
+        foreach ($pathNodes as $pathNode) {
```
phansys, line 237:
```suggestion
        foreach ($this->getPath($node, $options) as $pathNode) {
```

phansys (changes requested):

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -824,25 +887,120 @@ public function verify()
...
+    {
+        $defaultOptions = [
+            'sortByField'   => null,
+            'sortDirection' => 'ASC',
+        ];
+        $options = array_merge($defaultOptions, $options);
+
+        $meta = $this->getClassMetadata();
+        $config = $this->listener->getConfiguration($this->getEntityManager(), $meta->name);
+        $self = $this;
```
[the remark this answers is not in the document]
pmishev: We don't. I've just done it the same way as in the existing `recover()` method where it's not necessary either. I'll remove it in both places.

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -824,25 +887,120 @@ public function verify()
...
+                       ->set('node.'.$config['left'], ':left')
+                       ->set('node.'.$config['right'], ':right')
+                       ->where('node.id = :id');
+        if (isset($config['level'])) {
+            $updateQb->set('node.'.$config['level'], ':level');
+        }
+
+        $doRecover = function ($root, &$count, $level) use ($meta, $config, $self, $em, $options, $updateQb, &$doRecover) {
+            $rootEntity = $em->getReference($meta->getName(), $root['node_id']);
+            $lft = $count++;
```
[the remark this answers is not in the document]
pmishev: It is the `left` value of the node being processed

src/Tree/Mapping/Driver/Annotation.php
```diff
@@ -180,7 +180,7 @@ public function readExtendedMetadata($meta, array &$config)
                 $config['root'] = $field;
             }
             // level
-            if ($this->reader->getPropertyAnnotation($property, self::LEVEL)) {
+            if ($levelAnnotation = $this->reader->getPropertyAnnotation($property, self::LEVEL)) {
```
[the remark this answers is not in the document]
pmishev: Do you mean to separate the assignment with the condition on 2 separate lines?
Generally I've tried to keep to the existing code style and you can see the following `if` on line 195 is done the same way.

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -825,6 +894,10 @@ public function verify()
         if (isset($config['root'])) {
             $trees = $this->getRootNodes();
             foreach ($trees as $tree) {
+                // if a root node is specified, verify only it
+                if (is_object($options['treeRootNode']) && $options['treeRootNode'] !== $tree) {
```
phansys, line 898: If `$options['treeRootNode']` can only hold an object or `null`, I think `null !== $options['treeRootNode']` is cheaper.

src/Tree/Mapping/Driver/Annotation.php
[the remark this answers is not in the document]
phansys: Yes, but I haven't a strong argument here. Generally, the assignment inside a condition leads more changes in case of refactoring. Additionally, it makes harder to annotate the variables (when it is strictly required).

src/Tree/Entity/Repository/NestedTreeRepository.php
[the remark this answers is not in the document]
phansys: I think we can avoid this kind of abbreviations that don't save too much.

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -835,25 +908,124 @@ public function verify()
...
+        $updateQb = $em->createQueryBuilder()
+                       ->update($meta->getName(), 'node')
+                       ->set('node.'.$config['left'], ':left')
+                       ->set('node.'.$config['right'], ':right')
+                       ->where('node.id = :id');
+        if (isset($config['level'])) {
+            $updateQb->set('node.'.$config['level'], ':level');
+        }
+
+        $doRecover = function ($root, &$count, $level) use ($meta, $em, $options, $updateQb, &$doRecover) {
```
franmomu, line 946:
```suggestion
        $doRecover = function ($root, &$count, $level) use ($meta, $em, $options, $updateQb, &$doRecover): void {
```
We could also add parameter type declarations

src/Tree/Mapping/Driver/Annotation.php
```diff
@@ -180,7 +180,8 @@ public function readExtendedMetadata($meta, array &$config)
                 $config['root'] = $field;
             }
             // level
-            if ($this->reader->getPropertyAnnotation($property, self::LEVEL)) {
+            $levelAnnotation = $this->reader->getPropertyAnnotation($property, self::LEVEL);
+            if ($levelAnnotation) {
```
phansys, line 184: Could we use a stricter condition?

tests/Gedmo/Tree/NestedTreeRootRepositoryTest.php
```diff
@@ -238,8 +240,36 @@ public function testShouldRemoveRootNodeFromTree(): void
...
+        $repo = $this->em->getRepository(self::CATEGORY);
+        $carrots = $repo->findOneBy(['title' => 'Carrots']);
+
+        $this->expectException(InvalidArgumentException::class);
+        $repo->getPathAsString($carrots, [
+            'stringMethod' => $stringMethod,
+        ]);
+    }
+
+    public function invalidStringMethods(): array
```
phansys, line 260:
```suggestion
    public function invalidStringMethods(): iterable
```
You can also yield these values instead of put an array in memory.

phansys (changes requested):

tests/Gedmo/Tree/NestedTreeRootRepositoryTest.php
```diff
@@ -257,9 +257,9 @@ public function testGetPathAsStringWithInvalidStringMethod($stringMethod): void
         ]);
     }
 
-    public function invalidStringMethods(): array
+    public function invalidStringMethods(): iterable
     {
-        return [
+        yield from [
```
phansys, line 262: If you use `yield from`, the array in the `from` part is already in memory. The real advantage of `yield` can be leveraged by yielding every single item. By instance:
```php
yield [null];
yield [123];
yield ['nonExistingMethod'];
yield [''];
```

tests/Gedmo/Tree/NestedTreeRootRepositoryTest.php
```diff
@@ -157,7 +159,7 @@ public function testShouldSupportChildrenHierarchyAsHtml(): void
         $childOpen = '';
         $childClose = '';
         $nodeDecorator = static function ($node) {
-            return str_repeat('-', $node['level']).$node['title']."\n";
+            return str_repeat('-', $node['level'] - 1).$node['title']."\n";
```
franmomu, line 162: is this change needed because the level of `RootCategory` has been changed from 0 to 1? if that's the case I think we should change the expected output instead of this or do not change `RootCategory` level and create another fixture to test this new feature.

tests/Gedmo/Tree/NestedTreeRootRepositoryTest.php
```diff
@@ -174,7 +176,7 @@ public function testShouldSupportChildrenHierarchyAsHtml(): void
         // check support of the closures in rootClose
         $rootClose = static function () {return '</ul><!--rootCloseClosure-->'; };
         $childOpen = static function (&$node) {
-            return '<li class="depth'.$node['level'].'">';
+            return '<li class="depth'.($node['level'] - 1).'">';
```
franmomu, line 179: Here the same thing than https://github.com/doctrine-extensions/DoctrineExtensions/pull/2447/files#r922696534

tests/Gedmo/Tree/NestedTreeRootRepositoryTest.php
pmishev: Yes, that's the case. Changing the root level to 1 would change the test output from:
```
-Fruits
-Vegitables
--Carrots
--Potatoes
```
to:
```
--Fruits
--Vegitables
---Carrots
---Potatoes
```
I decided to change the node decorator to keep the old output as it looks more sensible. Other than that, I don't think it really matters if I changed the output instead. But, if you prefer that, I can change the output.
As for creating another fixture, I agree that would be ideal, but I've had to change the expected level in multiple tests that use the RootCategory fixture and if I had to to another fixture, this would result in a lot of duplicated tests. It just seemed a bit too much. If that's the way you prefer to go though, I'll need some time to work on it.

pmishev: Just to give you some context on this feature and the point of it - it was necessary because we were incrementally migrating from another framework, working with a legacy DB that we couldn't change at the time.

franmomu: It's not a blocker, I was just a bit concerned about someone reading the code and do not understand that that `-1` inside the method is because `RootCategory` has level 1, I guess is like in the other changes it would be:

```diff
- static::assertSame(1, $node->getLevel());
+ static::assertSame(1, $node->getLevel() - 1);
```
Anyway, it's fine by me when it's rebased.

franmomu (approved):

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +187,69 @@ public function getPathQueryBuilder($node)
      * Get the Tree path query by given $node
      *
      * @param object $node
+     * @phpstan-param array{includeNode?: bool} $options
+     *
+     * $options['includeNode'] = (bool) Whether to include the node itself (default: true)
```
phansys, line 192: If you need to add a description for the method arguments, please use the conventional PHPDoc `@param` annotation:
```suggestion
     * @param array $options
     *    bool "includeNode": Whether to include the node itself (default: true)
     *
     * @phpstan-param array{includeNode?: bool} $options
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +187,69 @@ public function getPathQueryBuilder($node)
...
      *
-     * @return \Doctrine\ORM\Query
+     * @return Query
      */
-    public function getPathQuery($node)
+    public function getPathQuery($node, array $options = [])
```
phansys, line 196: This is a BC break.

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -61,7 +63,7 @@ public function __call($method, $args)
     {
         if ('persistAs' === substr($method, 0, 9)) {
             if (!isset($args[0])) {
-                throw new \Gedmo\Exception\InvalidArgumentException('Node to persist must be available as first argument');
+                throw new InvalidArgumentException('Node to persist must be available as first argument');
```
phansys, line 66:
```suggestion
                throw new InvalidArgumentException('Node to persist must be available as first argument.');
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -70,7 +72,7 @@ public function __call($method, $args)
             $position = substr($method, 9);
             if ('Of' === substr($method, -2)) {
                 if (!isset($args[1])) {
-                    throw new \Gedmo\Exception\InvalidArgumentException('If "Of" is specified you must provide parent or sibling as the second argument');
+                    throw new InvalidArgumentException('If "Of" is specified you must provide parent or sibling as the second argument');
```
phansys, line 75:
```suggestion
                    throw new InvalidArgumentException('If "Of" is specified you must provide parent or sibling as the second argument.');
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +187,69 @@ public function getPathQueryBuilder($node)
...
+    {
+        $defaultOptions = [
+            'includeNode' => true,
+            'separator' => ' > ',
+            'stringMethod' => '__toString',
+        ];
+        $options += $defaultOptions;
+
+        if (!is_string($options['stringMethod'])) {
+            throw new InvalidArgumentException(sprintf('"stringMethod" parameter passed to %s must be a valid string', __METHOD__));
```
phansys, line 241:
```suggestion
            throw new InvalidArgumentException(sprintf('"stringMethod" option passed in argument 2 to %s must be a valid string.', __METHOD__));
```

src/Tree/Entity/Repository/NestedTreeRepository.php
```diff
@@ -172,24 +187,69 @@ public function getPathQueryBuilder($node)
...
+        }
+        if (!method_exists($node, $options['stringMethod'])) {
+            throw new InvalidArgumentException(sprintf('%s must implement method "%s"', get_class($node), $options['stringMethod']));
```
phansys, line 244:
```suggestion
            throw new InvalidArgumentException(sprintf('%s must implement method "%s".', get_class($node), $options['stringMethod']));
```

phansys (changes requested):

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: Ok, I wasn't aware there was a convention in documenting these. For example here they are described a bit differently: https://github.com/doctrine-extensions/DoctrineExtensions/blob/c51e05ebbeccdabe9d0f09cc264c1657880a2bfa/src/Tree/RepositoryUtilsInterface.php#L47
Your proposed format looks nice, so I can change my comments like that. Do you mind that there will still be a discrepancy with the above example?

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: I'm sorry, how is this a BC break?

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: Doesn't this qualify as a BC change, in the unlikely scenario someone is checking for the exception message text?

src/Tree/Entity/Repository/NestedTreeRepository.php
phansys: This is adding a new parameter to a method that (currently) can be extended by users.
See:
* https://3v4l.org/ldmIb;
* https://en.wikipedia.org/wiki/Liskov_substitution_principle.

src/Tree/Entity/Repository/NestedTreeRepository.php
phansys: I honestly hope not. But if you think it may be the case, please dismiss my suggestion.

src/Tree/Entity/Repository/NestedTreeRepository.php
phansys: I think the only implicit convention is to leverage the PHPDoc annotations when possible. We have a hard normalizing the existing codebase. By now, we haven't a rule for the indentation other than consistency, so feel free to come up the way you think is a better fit.

pmishev: in this case, I'd like to suggest:
```
     * @param array $options
     *     includeNode => (bool) Whether to include the node itself (default: true),
```
I think it's a bit more readable having the param name first, not the type, if you have many params and you're trying to scan through them. WDYT?

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: Well yes, you have a point, but if we're going to consider this a breaking change, I have also added new methods, such as `getPathAsString()` and `recoverFast()`, so they are then breaking changes as well. What do you suggest to do about that?

src/Tree/Entity/Repository/NestedTreeRepository.php
phansys: I have no personal preference here. The only thing I can use as base for the syntax I've suggested is the structure used in the PHPDoc [`@param`](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/param.html) tag:
```
@param [<Type>] [name] [<description>]
```

src/Tree/Entity/Repository/NestedTreeRepository.php
phansys: AFAIK, adding new methods in a class that is not part of the current API is not a BC break.

franmomu: you can do something similar to: 

```php
$options = func_get_args()[1] ?? [];

if (!\is_array($options)) {
    throw new TypeError('Argument 2 MUST be an array');
}
```
And we'll add the parameter in the next major version.

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: I updated the phpdocs according to this https://stackoverflow.com/a/51032174/2737626

src/Tree/Entity/Repository/NestedTreeRepository.php
pmishev: Are you suggesting that I create a separate PR where the only changes are adding the above snippet in the following methods?
```
    public function getPathQueryBuilder($node)
    public function getPathQuery($node)
    public function getPath($node)
    public function verify()
    public function recover()
```
...and then this PR will just wait for a next major version?

TBH, I'm not sure what help will this snippet do. Someone might have extended the method and added a second array argument, in which case they won't get a TypeError and in the next major version, their code will still be broken.

pmishev: Apologies, I've been away for a while. I've made the requested changes now.

pmishev:
Just checking in to see if you need anything more fixed in this PR?

franmomu:
> Just checking in to see if you need anything more fixed in this PR?

Just a rebase to fix the conflict in the `CHANGLOG:md` file from my side, let see if @phansys can give another review.

phansys (commented):
I'm not convinced about the `@param` annotations like this:
```
 * @param array  $options = [
 *                        'includeNode' => true, // Whether to include the node itself
 *                        ]
 *
```
but I don't want to block the additions introduced in this PR.

pmishev:
Did the rebase. 
As far as the `@param` are concerned, we had a discussion about those above. I did some research then and didn't find any consensus on what's the best way to document those, so I picked one that seemed the least bad. I don't really care much about those, so if you prefer another one, I can change them.

phansys:
>I don't really care much about those, so if you prefer another one, I can change them.

For lack of a better example, I think [this one](https://github.com/aws/aws-sdk-php/blob/5791e496e3fc0847f40c86515ed8659b0ecd5991/src/S3/S3Client.php#L297-L345) a bit tidier.

pmishev:
So, instead of this:
```
     * @param object $node
     * @param array  $options = [
     *                        'includeNode' => true, // Whether to include the node itself
     *                        ]
     *
     * @phpstan-param array{includeNode?: bool} $options
```
it would be this:
```
     * @param object $node
     * @phpstan-param array{includeNode?: bool} $options
     *
     * options:
     * - includeNode: (bool) Whether to include the node itself. Defaults to true.
```

Just to point out the cons I see in this format:
- The default values are not immediately obvious.
- The `$option` will no longer have a `@param` annotation

Please confirm and I'll change them.

phansys:
Thanks @pmishev.
As I mentioned, I don't want to block this PR.
Regarding your points, I guess we can keep the `@param` tag if it adds some value, like `@param array<string, mixed> $options`.
About the default value, I find the example explicit enough.

phansys (approved):

franmomu:
thanks for keeping working on this @pmishev! and sorry for taking that long

PR #2454, opened 2022-05-08
Allow doctrine/persistence 3

### LINKED ISSUE #2450: doctrine/persistence v3 is not allowed
This repository is restricting `doctrine/persistence` to version 2.
Unless some changes introduced in v3 are blocking, the new version should be allowed as well.

franmomu:
Apparently the removal of `LifecycleEventArgs::getEntity()` should not be an issue since it uses the `LifecycleEventArgs::getEntity()` from doctrine/orm.

It would be nice to test this on real projects just in case there is something missing.

Closes #2450

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 ---
 
 ## [Unreleased]
+## Added
+- Add support for doctrine/persistence 3
+
 ## Changed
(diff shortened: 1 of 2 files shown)


franmomu:
Since apparently there are no issues with tests, let's merge it so it would be easier to test in real projects.

TomAtom:
And now new release please :-)

PR #2456, opened 2022-05-13
Make doctrine/cache optional

franmomu:
Since https://github.com/doctrine-extensions/DoctrineExtensions/pull/2451 we don't need `doctrine/cache` to be required.


jcglombard:
Very nice!

TamasSzigeti (approved):

PR #2462, opened 2022-05-23
[Sortable] Resolve Column Type when fetching Max Position from Database

dritter:
Hi there!

In SQLite it makes a difference if date columns are queried with '00:00:00' as timestring or without. MySQL seems to ignore the time completely. Unfortunately there is an asymmetry between SQL Types and PHP Objects: 'date' and 'datetime' Column point both to a DateTime Object in PHP. Therefore, we cannot derive the column type from the PHP Object. Luckily we can ask the ClassMetadata in Doctrine for the right type.

IMHO, this change belongs in Doctrine ORM, but it is hard to implement there. Hence we have to work around that limitation.
There are a lot of PRs/Issues on Doctrine side:
- https://github.com/doctrine/orm/pull/1383
- https://github.com/doctrine/orm/pull/7236

to name a few.

### DIFF
--- a/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
+++ b/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
@@ -82,34 +82,22 @@ class ItemWithDateColumn
         $this->position = $position;
     }
 
-    /**
-     * @return mixed
-     */
-    public function getDate()
+    public function getDate(): ?\DateTime
     {
         return $this->date;
     }
 
-    /**
-     * @param mixed $date
-     */
-    public function setDate($date): void
+    public function setDate(?\DateTime $date): void
     {
         $this->date = $date;
     }
 
-    /**
-     * @return mixed
-     */
-    public function getUserId()
+    public function getUserId(): ?string
     {
         return $this->userId;
     }
 
-    /**
-     * @param mixed $userId
-     */
-    public function setUserId($userId): void
+    public function setUserId(?string $userId): void
     {
         $this->userId = $userId;
     }
--- a/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
+++ b/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
@@ -45,7 +45,7 @@ class ItemWithDateColumn
      * @ORM\Column(type="date")
      */
     #[Gedmo\SortableGroup]
-    #[ORM\Column(type: 'date')]
+    #[ORM\Column(type: Types::DATE_MUTABLE)]
     private $date;
 
     /**
--- a/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
+++ b/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
@@ -33,6 +33,8 @@ class ItemWithDateColumn
     private $id;
 
     /**
+     * @var int
+     *
      * @Gedmo\SortablePosition
      * @ORM\Column(type="integer")
      */
--- a/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
+++ b/tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace Gedmo\Tests\Sortable\Fixture;
+
+use Doctrine\DBAL\Types\Types;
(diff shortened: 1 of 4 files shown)


dritter:
@franmomu I fixed the last code style issue and rebased to fix the phpstan error (hopefully). Could you trigger a CI-Build?

tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
```diff
@@ -0,0 +1,110 @@
...
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: Types::INTEGER)]
+    private $id;
+
+    /**
+     * @Gedmo\SortablePosition
+     * @ORM\Column(type="integer")
+     */
+    #[Gedmo\SortablePosition]
+    #[ORM\Column(type: 'integer')]
```
franmomu, line 40:
```suggestion
    #[ORM\Column(type: Types::INTEGER)]
```

tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
```diff
@@ -0,0 +1,110 @@
...
+    private $position = 0;
+
+    /**
+     * @Gedmo\SortableGroup
+     * @ORM\Column(type="date")
+     */
+    #[Gedmo\SortableGroup]
+    #[ORM\Column(type: 'date')]
```
franmomu, line 48:
```suggestion
    #[ORM\Column(type: Types::DATE_MUTABLE)]
```

tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
```diff
@@ -0,0 +1,110 @@
...
+     * @ORM\Column(type="integer")
+     */
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: Types::INTEGER)]
+    private $id;
+
+    /**
+     * @Gedmo\SortablePosition
+     * @ORM\Column(type="integer")
```
franmomu, line 37:
```suggestion
    /**
     * @var int
     *
     * @Gedmo\SortablePosition
     * @ORM\Column(type="integer")
```

tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    #[Gedmo\SortablePosition]
+    #[ORM\Column(type: 'integer')]
+    private $position = 0;
+
+    /**
+     * @Gedmo\SortableGroup
+     * @ORM\Column(type="date")
+     */
```
franmomu, line 46:
```suggestion
    /**
     * @var \DateTime|null
     *
     * @Gedmo\SortableGroup
     * @ORM\Column(type="date")
     */
```
Can you please update the getter and setter to use `\DateTime|null` instead of `mixed`?

tests/Gedmo/Sortable/Fixture/ItemWithDateColumn.php
```diff
@@ -0,0 +1,110 @@
...
+    #[Gedmo\SortableGroup]
+    #[ORM\Column(type: 'date')]
+    private $date;
+
+    /**
+     * @Gedmo\SortableGroup
+     * @ORM\Column(nullable=false)
+     */
```
franmomu, line 54:
```suggestion
    /**
     * @var string|null
     *
     * @Gedmo\SortableGroup
     * @ORM\Column(nullable=false)
     */
```
Can you please update the getter and setter to use `string|null` instead of `mixed`?

franmomu:
And can you also add an entry in `CHANGELOG.md` please? thanks

dritter:
@franmomu All done.

franmomu (commented):
hi @dritter, sorry for not paying attention before, but now looking at the test I'm not sure if it's easy to understand the purpose of that test.

Can it be simplified? I don't know if it's necessary to create 100 instances or with just 3 is enough

dritter:
@franmomu No worries! Happy to save some CPU cycles. I reduced the amount of generated test data to six and removed the unnecessary Group `userId`. I think that makes the test better understandable.

franmomu (approved):

franmomu:
Thanks @dritter!

PR #2477, opened 2022-06-22
[Loggable] Support embeddable xml mapping

eytanProxi:


src/Loggable/Mapping/Driver/Xml.php
```diff
@@ -35,7 +35,7 @@ public function readExtendedMetadata($meta, array &$config)
 
         $xml = $xml->children(self::GEDMO_NAMESPACE_URI);
 
-        if ('entity' === $xmlDoctrine->getName() || 'document' === $xmlDoctrine->getName() || 'mapped-superclass' === $xmlDoctrine->getName()) {
+        if (in_array($xmlDoctrine->getName(), ['entity', 'document', 'mapped-superclass', 'embeddable'])) {
```
phansys, line 38:
```suggestion
        if (in_array($xmlDoctrine->getName(), ['entity', 'document', 'mapped-superclass', 'embeddable'], true)) {
```

phansys (changes requested):

src/Loggable/Mapping/Driver/Xml.php
eytanProxi: It's done

pierre-H:
@phansys  : what should I do ?

phansys:
Hi @pierre-H!
You should add a changelog note with the related changes ([ref.](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/CONTRIBUTING.md#changelog)) and the required tests to cover this feature.

PR #2479, opened 2022-06-27
Specify array configuration in extensions

franmomu:
I was trying to see errors using PHPStan level 5 and most of the are because when calling `SomeListener::getConfiguration()` it doesn't know what that array contains and also calling `SomeListener::getEventAdapter()` it doesn't know that if it's a custom event adapter.

This PR tries to let PHPStan understand these calls overriding PHPDoc, ideally it should infer it from the code, but I can't think of an easy way to do that.

In level 5 it goes from 75 errors to 33 (I probably have missed some, I'll check the other extensions if this is a right approach).

### DIFF
--- a/src/References/ReferencesListener.php
+++ b/src/References/ReferencesListener.php
@@ -21,6 +21,24 @@ use Gedmo\Mapping\MappedEventSubscriber;
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  * @author Bulat Shakirzyanov <mallluhuct@gmail.com>
  * @author Jonathan H. Wage <jonwage@gmail.com>
+ *
+ * @phpstan-type ReferenceConfiguration = array{
+ *   field?: string,
+ *   type?: string,
+ *   class?: class-string,
+ *   identifier?: string,
+ *   mappedBy?: string,
+ *   inversedBy?: string,
+ * }
+ *
+ * @phpstan-type ReferencesConfiguration = array{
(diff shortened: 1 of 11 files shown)


phansys:
>ideally it should infer it from the code, but I can't think of an easy way to do that.

If this is a workaround, I think we should be clear about that through a `@todo` comment or a similar approach.

franmomu:
There are quite dynamic things around events, I guess it would be nice to change

https://github.com/doctrine-extensions/DoctrineExtensions/blob/f3a912950c3eaadbe0b4eac69c9424e62176f902/src/Mapping/MappedEventSubscriber.php#L216-L233

but I have no idea how to deal with that right now, I can open an issue about that maybe? not sure what to write in the `@todo`

src/References/ReferencesListener.php
```diff
@@ -21,6 +21,24 @@
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  * @author Bulat Shakirzyanov <mallluhuct@gmail.com>
  * @author Jonathan H. Wage <jonwage@gmail.com>
+ *
+ * @phpstan-type ReferenceConfiguration = array{
+ *     field?: string,
```
phansys, line 26: In some docblocks you're using 4 spaces for indentation, and 2 in others.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2482, opened 2022-07-01
handle null   abs($levelDelta)

tacman:
passing null to abs is deprecated, this handles the null case.  

Alternatively, the default value in the signature could be 0, and type-hinted as an int.  This PR is the smallest change, though, and gets rid of the 

06:00:24 INFO      [php] Deprecated: abs(): Passing null to parameter #1 ($num) of type int|float is deprecated ["exception" => ErrorException { …}]

messages.

### DIFF
--- a/src/Tree/Strategy/ORM/Nested.php
+++ b/src/Tree/Strategy/ORM/Nested.php
@@ -647,7 +647,7 @@ class Nested implements Strategy
         $sign = ($delta >= 0) ? ' + ' : ' - ';
         $absDelta = abs($delta);
         $levelSign = ($levelDelta >= 0) ? ' + ' : ' - ';
-        $absLevelDelta = $levelDelta ? abs($levelDelta) : 0;
+        $absLevelDelta = null !== $levelDelta ? abs($levelDelta) : 0;
 
         $qb = $em->createQueryBuilder();
         $qb->update($config['useObjectClass'], 'node')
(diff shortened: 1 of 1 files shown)


src/Tree/Strategy/ORM/Nested.php
```diff
@@ -647,7 +647,7 @@ public function shiftRangeRL(EntityManagerInterface $em, $class, $first, $last,
         $sign = ($delta >= 0) ? ' + ' : ' - ';
         $absDelta = abs($delta);
         $levelSign = ($levelDelta >= 0) ? ' + ' : ' - ';
-        $absLevelDelta = abs($levelDelta);
+        $absLevelDelta = $levelDelta ? abs($levelDelta) : 0;
```
franmomu, line 650:
```suggestion
        $absLevelDelta = null !== $levelDelta ? abs($levelDelta) : 0;
```
To use a stricter comparison

src/Tree/Strategy/ORM/Nested.php
```diff
@@ -647,7 +647,7 @@ public function shiftRangeRL(EntityManagerInterface $em, $class, $first, $last,
         $sign = ($delta >= 0) ? ' + ' : ' - ';
         $absDelta = abs($delta);
         $levelSign = ($levelDelta >= 0) ? ' + ' : ' - ';
-        $absLevelDelta = abs($levelDelta);
+        $absLevelDelta = null !== $levelDelta ? abs($levelDelta) : 0;
```
phansys, line 650: As the type hint for the `$levelDelta` argument only allows `int`, I think we should:
* change its default value to `0`;
* add a deprecation message if other type is passed and cast it to `int` or throw a `\TypeError` error (this is a BC break).

phansys:
Thank you @tacman!

Issue #2484, opened 2022-07-02
Breaking BC when using with Doctrine Bundle since version 3.7.0


ossinkine:
Since version 3.7.0 this library [uses cache from persisting configuration](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2451) and this is a bug and I will describe below.

Doctrine Bundle [uses PhpArrayAdapter](https://github.com/doctrine/DoctrineBundle/blob/59c101a5e80078648b5f558db9d0dd40fe91e147/DependencyInjection/DoctrineExtension.php#L1010-L1017) by default as metadata cache as the fastest one. This adapter requires cache warming up before using, this is doing by [DoctrineMetadataCacheWarmer](https://github.com/doctrine/DoctrineBundle/blob/59c101a5e80078648b5f558db9d0dd40fe91e147/CacheWarmer/DoctrineMetadataCacheWarmer.php). It set up `ArrayAdapter` to `MetadataFactory` and get metadata for all classes, and then dump all cached metadata to php file. So if this library wants to use the same cache from persisting layer configuration it should fill the same `ArrayAdapter` before but it doesn't, so configuration isn't warmed up and can't be received from cache in the future.

The best way would be to get cache not from configuration but from metadata factory like this
```diff
    private function getCacheItemPool(ObjectManager $objectManager): CacheItemPoolInterface
    {
        if (null !== $this->cacheItemPool) {
            return $this->cacheItemPool;
        }

        if ($objectManager instanceof EntityManagerInterface || $objectManager instanceof DocumentManager) {
-           $metadataCache = $objectManager->getConfiguration()->getMetadataCache();
+           $metadataCache = $objectManager->getMetadataFactory()->getCache();

            if (null !== $metadataCache) {
                $this->cacheItemPool = $metadataCache;

                return $this->cacheItemPool;
            }
        }

        $this->cacheItemPool = new ArrayAdapter();

        return $this->cacheItemPool;
    }
```
but `getCache` method is protected, so I don't how to fix it.

franmomu:
Thanks @ossinkine for reporting, I didn't know that `DoctrineMetadataCacheWarmer` was modifying the cache of `MetadataFactory`, I thought it was only set from the `Configuration` class.

This explains https://github.com/doctrine-extensions/DoctrineExtensions/issues/2463#issuecomment-1137797451.

[...]

Issue #2463, opened 2022-05-24
#[Gedmo\Loggable] on parent entity?


stephanvierkant:
I've got something like this:

```
class User extends BaseEntity {}

#[Gedmo\Loggable]
class BaseEntity {}

```

This worked, but on 3.7 I got this error message:

> Uncaught PHP Exception Gedmo\Exception\InvalidMappingException: "Class must be annotated with Loggable annotation in order to track versioned fields in class - App\Entity\User"

I sticked to `"doctrine/persistence": "2.5.3"` to found out if this related to this dependency, but as far as I can see now, the problem started when updating to 3.7 of this library.

My test suite didn't fail on 3.7, but in production I've got this problem. So it might be related to caching. I'll try to create a reproducible example, but please let me know your thoughts.

franmomu:
hmm that seems to be fine, that commit it's the only one modifying things related to cache, I would try to see the difference between the previous version of the commit

stephanvierkant:
The previous version returns this:

```
^ Symfony\Component\Cache\Adapter\PhpArrayAdapter {#337 ▼
  -file: "/home/stephan/PhpstormProjects/qurplus/var/cache/prod/doctrine/orm/default_metadata.php"
  -keys: array:898 [▶]
  -values: array:443 [▶]
  -computing: []
  -pool: Symfony\Component\Cache\Adapter\ArrayAdapter {#338 ▶}
}
```

So far more cache items, including my entities:

```
-values: array:443 [▼
    0 => Closure() {#410 ▶}
    1 => []
    2 => array:3 [▼
      "create" => array:1 [ …1]
      "update" => array:1 [ …1]
      "useObjectClass" => "App\Entity\User\User"
    ]
```

stephanvierkant:
The previous version returns this:

```
^ Symfony\Component\Cache\Adapter\PhpArrayAdapter {#337 ▼
  -file: "{project}/var/cache/prod/doctrine/orm/default_metadata.php"
  -keys: array:898 [▶]
  -values: array:443 [▶]
  -computing: []
  -pool: Symfony\Component\Cache\Adapter\ArrayAdapter {#338 ▶}
}
```

So far more cache items, including my entities:

```
-values: array:443 [▼
    0 => Closure() {#410 ▶}
    1 => []
    2 => array:3 [▼
      "create" => array:1 [ …1]
      "update" => array:1 [ …1]
      "useObjectClass" => "App\Entity\User\User"
[...]

PR #2485, opened 2022-07-02
Get cache from MetadataFactory

franmomu:
Fixes https://github.com/doctrine-extensions/DoctrineExtensions/issues/2484 and https://github.com/doctrine-extensions/DoctrineExtensions/issues/2463#issuecomment-1137797451

I don't know other way to access the cache, ideally the cache should be set by the user and this won't be needed, when using the bundle there is https://github.com/stof/StofDoctrineExtensionsBundle/pull/436 to automatically set this cache, but the PR hasn't been reviewed yet, my idea after finishing that PR was to deprecate not setting the cache pool.

cc @ossinkine @dpfaffenbauer

### DIFF
--- a/src/Mapping/MappedEventSubscriber.php
+++ b/src/Mapping/MappedEventSubscriber.php
@@ -21,6 +21,7 @@ use Doctrine\Common\EventArgs;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
 use Gedmo\Mapping\Event\AdapterInterface;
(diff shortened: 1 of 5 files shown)


dpfaffenbauer:
Wow, will give it a try next week 💪💪💪🚀🚀🚀

dpfaffenbauer:
@franmomu This fixes it for me. 🚀

ossinkine:
This fixes it for me too 👍

src/Mapping/MappedEventSubscriber.php
```diff
@@ -291,7 +292,12 @@ private function getCacheItemPool(ObjectManager $objectManager): CacheItemPoolIn
         }
 
         if ($objectManager instanceof EntityManagerInterface || $objectManager instanceof DocumentManager) {
-            $metadataCache = $objectManager->getConfiguration()->getMetadataCache();
+            $metadataFactory = $objectManager->getMetadataFactory();
+            $getCache = \Closure::bind(static function (AbstractClassMetadataFactory $metadataFactory): ?CacheItemPoolInterface {
```
phansys, line 296: At first sight, violating the encapsulation seems a workaround.
Do we have any chance to add a TODO explaining how we should resolve this problem in a proper way?

franmomu: I've added a comment to try to explain it, for the CS fail I've created https://github.com/doctrine-extensions/DoctrineExtensions/pull/2488

phansys: Thank you! #2488 is merged. Let's rebase this :rocket:

dpfaffenbauer: 🚀🚀

dpfaffenbauer:
@franmomu thank you very much for the fix. 🍻🍻

phansys:
Thnaks @franmomu!

Issue #2492, opened 2022-07-21
Issue with checking using $meta->hasField


yellow1912:
In the file: ClosureTreeRepository.php

```
if ($sortByField) {
            if (is_array($sortByField)) {
                foreach ($sortByField as $key => $field) {
                    $fieldDirection = strtolower(is_array($direction) ? ($direction[$key] ?? 'asc') : $direction);
                    if ($meta->hasField($field) && in_array($fieldDirection, ['asc', 'desc'], true)) {
                        $qb->addOrderBy('node.'.$field, $fieldDirection);
                    } else {
                        throw new InvalidArgumentException(sprintf('Invalid sort options specified: field - %s, direction - %s', $field, $fieldDirection));
                    }
                }
            } else {
                if ($meta->hasField($sortByField) && in_array(strtolower($direction), ['asc', 'desc'], true)) {
                    $qb->orderBy('node.'.$sortByField, $direction);
                } else {
                    throw new InvalidArgumentException(sprintf('Invalid sort options specified: field - %s, direction - %s', $sortByField, $direction));
                }
            }
        }
```

It seems that with the latest SF5 and Doctrine version root is NOT a field, it is an association instead. So the sort by root throws an exception.

yellow1912:
Proposed fix:
```
if (!$sortByField) {
            $qb->orderBy('node.'.$config['left'], 'ASC');
        } elseif (is_array($sortByField)) {
            foreach ($sortByField as $key => $field) {
                $fieldDirection = strtolower(is_array($direction) ? ($direction[$key] ?? 'asc') : $direction);
                if (($meta->hasAssociation($field) || $meta->hasField($field)) && in_array($fieldDirection, ['asc', 'desc'], true)) {
                    $qb->addOrderBy('node.'.$field, $fieldDirection);
                } else {
                    throw new InvalidArgumentException(sprintf('Invalid sort options specified: field - %s, direction - %s', $field, $fieldDirection));
                }
            }
        } else {
            if (($meta->hasAssociation($sortByField) || $meta->hasField($sortByField)) && in_array(strtolower($direction), ['asc', 'desc'], true)) {
                $qb->orderBy('node.'.$sortByField, $direction);
            } else {
                throw new InvalidArgumentException(sprintf('Invalid sort options specified: field - %s, direction - %s', $sortByField, $direction));
            }
        }
```

This should be applied for NestedTree and ClosureTree Repository

PR #2493, opened 2022-07-22
[Tree] [NestedSet] [ClosureTree] Allow sorting children by a ManyToOne relation

pmishev:
Fixes https://github.com/doctrine-extensions/DoctrineExtensions/issues/2492
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2446#pullrequestreview-1046194534

### DIFF
--- a/src/Tree/Entity/Repository/ClosureTreeRepository.php
+++ b/src/Tree/Entity/Repository/ClosureTreeRepository.php
@@ -153,14 +153,14 @@ class ClosureTreeRepository extends AbstractTreeRepository
             if (is_array($sortByField)) {
                 foreach ($sortByField as $key => $field) {
                     $fieldDirection = strtolower(is_array($direction) ? ($direction[$key] ?? 'asc') : $direction);
-                    if ($meta->hasField($field) && in_array($fieldDirection, ['asc', 'desc'], true)) {
+                    if (($meta->hasField($field) || $meta->isSingleValuedAssociation($field)) && in_array($fieldDirection, ['asc', 'desc'], true)) {
                         $qb->addOrderBy('node.'.$field, $fieldDirection);
                     } else {
                         throw new InvalidArgumentException(sprintf('Invalid sort options specified: field - %s, direction - %s', $field, $fieldDirection));
                     }
                 }
             } else {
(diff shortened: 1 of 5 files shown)


src/Tree/Entity/Repository/ClosureTreeRepository.php
```diff
@@ -153,14 +153,14 @@ public function childrenQueryBuilder($node = null, $direct = false, $sortByField
             if (is_array($sortByField)) {
                 foreach ($sortByField as $key => $field) {
                     $fieldDirection = strtolower(is_array($direction) ? ($direction[$key] ?? 'asc') : $direction);
-                    if ($meta->hasField($field) && in_array($fieldDirection, ['asc', 'desc'], true)) {
+                    if (($meta->hasField($field) || $meta->isSingleValuedAssociation($field)) && in_array($fieldDirection, ['asc', 'desc'], true)) {
```
phansys, line 156: IMO, we should move the cheaper check first:
```suggestion
                    if (in_array($fieldDirection, ['asc', 'desc'], true) && ($meta->hasField($field) || $meta->isSingleValuedAssociation($field))) {
```

yellow1912:
Great fix, I hope this gets merged soon because it will fix the issue immediately.

src/Tree/Entity/Repository/ClosureTreeRepository.php
pmishev: Those function calls are not really expensive I think - just checking if array keys exist and doing bitwise operation.
I've moved the `strtolower()` call in the `if`, so it's now not always called, but when it is, it would probably be the most expensive call, so I've left the order as it is.

franmomu (commented):
Thanks for the fix, can you please add a changelog entry?

phansys (approved):

franmomu (approved):

phansys:
Thank you @pmishev!

yellow1912:
@franmomu can we create a new tag version for this change?

franmomu:
yeah, since we introduced a deprecation (in another PR) we need to create a minor version, I think it would be nice to include https://github.com/doctrine-extensions/DoctrineExtensions/pull/2301 and https://github.com/doctrine-extensions/DoctrineExtensions/pull/2447 which seem to be almost ready, let me see if we can move those forward

PR #2522, opened 2022-09-29
fix on documentation about slug attribute and unique property

baptistepottier:
on slug annotation, when property "unique" is true, it become suffixed and not prefixed


phansys (approved):

franmomu (approved):

franmomu:
Thanks @baptistepottier!

PR #2509, opened 2022-08-30
blameable.md - wrong join column name on createdBy

DennisdeBest:
When copy pasting the docs without paying attention (it happens ....) the `bin/console make:migrations` generated a migration file which ended up throwing an error : `(errno: 150 "Foreign key constraint is incorrectly formed") `

My entity had the `TimestampableEntity` trait and I wanted to add the blameable fields but with an association not just a string. Therefore I added the lines 
```php 
    #[ORM\ManyToOne]
    #[ORM\JoinColumn(name: 'created_by', referencedColumnName: 'id')]
    #[Gedmo\Blameable(on: 'create')]
    private ?User $createdBy = null;

    #[ORM\ManyToOne]
    #[ORM\JoinColumn(name: 'updated_by', referencedColumnName: 'id')]
    #[Gedmo\Blameable(on: 'update')]
    private ?User $updatedBy = null;
```

But the JoinColumn name of `createdBy` was set to `create_at` which made the migration fail. A simple change to the documentation for when you do not pay attention to the copy paste habits ...

PR #2523, opened 2022-10-15
Update blameable.md

kamil-karkus:
typo


franmomu:
Thanks @kamil-karkus! it was done in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2509

PR #2531, opened 2022-11-03
Make TimestampableEntity getter type hints nullable

villermen:
Signals nullability for return types of `TimestampableEntity::getCreatedAt()` and `TimestampableEntity::getUpdatedAt()`. They may actually return `null` when they are not yet persisted.

This helps static type checkers understand this trait better.

### DIFF
--- a/src/Timestampable/Traits/TimestampableEntity.php
+++ b/src/Timestampable/Traits/TimestampableEntity.php
@@ -53,7 +53,7 @@ trait TimestampableEntity
     /**
      * Returns createdAt.
      *
-     * @return \DateTime
+     * @return \DateTime|null
      */
     public function getCreatedAt()
     {
@@ -75,7 +75,7 @@ trait TimestampableEntity
     /**
      * Returns updatedAt.
      *
-     * @return \DateTime
+     * @return \DateTime|null
      */
     public function getUpdatedAt()
     {
--- a/src/Timestampable/Traits/TimestampableEntity.php
+++ b/src/Timestampable/Traits/TimestampableEntity.php
@@ -21,7 +21,7 @@ use Gedmo\Mapping\Annotation as Gedmo;
 trait TimestampableEntity
 {
     /**
-     * @var \DateTime
+     * @var \DateTime|null
      * @Gedmo\Timestampable(on="create")
      * @ORM\Column(type="datetime")
      */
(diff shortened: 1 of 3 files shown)


src/Timestampable/Traits/TimestampableEntity.php
```diff
@@ -53,7 +53,7 @@ public function setCreatedAt(\DateTime $createdAt)
     /**
      * Returns createdAt.
      *
-     * @return \DateTime
+     * @return \DateTime|null
```
phansys, line 56: If we are going to allow `null` as return type for these methods, I guess we should first update the hints for the properties that are being returned here.

phansys (changes requested):
Hi @villermen! Thank you for your first contribution.
Please, be aware about the PR's labels. It currently requires a rebase.

villermen:
@phansys Thanks for the review! I have rebased the PR against `main`, and applied nullable hints to all timestampable traits.

Side question: Is there anything against adding them as native return type? E.g., `getCreatedAt(): ?\DateTime`? I noticed the minimum required PHP version is 7.2 where this is supported, and it's not technically a BC break right?

phansys:
>Side question: Is there anything against adding them as native return type? E.g., getCreatedAt(): ?\DateTime?

The reason behind that decision is to avoid issues for users that are not respecting the type hints declared in our API. Because its nature, I guess changes like this will be part of the next major release.

phansys (approved):

phansys:
Thank you @villermen!

nvdbeek:
It is technically correct that these properties may be null when retrieved before being persisted. However in which situations would you actually call these methods before then? Is it not improper to call this method when they are nullable, wouldn't be better to throw exceptions in the situation when this data is not available yet. Making these return types nullables forces us to either throw exceptions or add assertions which are unnecessary in most cases.

villermen:
@nvdbeek I created this PR to have the documentation match the behavior, mainly to correct static type check inconsistencies I was experiencing. I would be in favor of a non-null property too, but I don't think an exception will help prevent undesirable situations. `null` conveys better that the value has not yet been set and allows creating simple fallback logic with `??`, `?->` and the likes.

Forcing initialization of the properties upon construction (not on get) so that they are always available may also prove difficult when using traits. I wouldn't know how to approach that, but if you have an idea there a PR is probably welcome =)

PR #2532, opened 2022-11-06
3.10.0

phansys:


franmomu (approved):
Thanks @phansys! The composer.json file should be updated as well

phansys:
> Thanks @phansys! The composer.json file should be updated as well

Good catch! It's updated.

PR #2537, opened 2022-11-21
Support doctrine/collections 2

### LINKED ISSUE #2528: allow for doctrine/collections:^2.0
I see that doctrine/collections 2.0 is out now, but doctrine/orm doesn't yet allow it.

Nonetheless, eventually this package will need to support it.

oojacoboo:
doctrine/collections 2 has a number of updates that should be supported by this lib.

Closes #2528.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "php": "^7.2 || ^8.0",
         "behat/transliterator": "~1.2",
         "doctrine/annotations": "^1.13",
-        "doctrine/collections": "^1.0",
+        "doctrine/collections": "^1.0 || ^2.0",
         "doctrine/common": "^2.13 || ^3.0",
         "doctrine/event-manager": "^1.2",
         "doctrine/persistence": "^2.2 || ^3.0",


phansys:
Thank you for the proposal @oojacoboo.
I'm afraid we aren't able to allow this version yet:
```
composer why-not doctrine/collections 2

gedmo/doctrine-extensions  3.11.x-dev  requires  doctrine/collections (^1.0)         
doctrine/mongodb-odm       2.4.2       requires  doctrine/collections (^1.5)         
doctrine/orm               2.13.3      requires  doctrine/collections (^1.5) 
```
You can check this in the build results of the current pipeline. The resolved version is always taken from the 1.x branch.

oojacoboo:
@phansys I don't see the issue with including 2.0 with the pipe, `||`, meaning OR.  I was able to fork this lib with the requested changes and `doctrine/collections` properly resolves and does install 2.0.  There are other dependency concerns with 1.0 support.  Adding the `||` doesn't have any BC issues.  If your requirements need 1.0, so be it - it'll get installed.

If there are `doctrine/collections` 2.0 changes that break current support within this extension, it shouldn't be included.  But that doesn't seem to be the case.  So, I really fail to see the issue here.

phansys:
>@phansys I don't see the issue with including 2.0 with the pipe, ||, meaning OR. I was able to fork this lib with the requested changes and doctrine/collections properly resolves and does install 2.0. There are other dependency concerns with 1.0 support.

Could you please share the job(s) in the [CI pipeline of this PR](https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/3511692571) where the evidence about the resolution of `doctrine/collections ^2.0` can be found?

>Adding the || doesn't have any BC issues. If your requirements need 1.0, so be it - it'll get installed.
>
>If there are doctrine/collections 2.0 changes that break current support within this extension, it shouldn't be included. But that doesn't seem to be the case. So, I really fail to see the issue here.

IMO, there is no sense in widening this constraint if it can not be leveraged currently.

oojacoboo:
@phansys 

>Could you please share the job(s) in the [CI pipeline of this PR](https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/3511692571) where the evidence about the resolution of doctrine/collections ^2.0 can be found?

There isn't one to be found since it's installing the dev dependencies.  That's the issue here.  The dependencies are only requiring `doctrine/collections ^1.5` in `dev`.  So, for people using this lib, that have other dependencies that need `doctrine/collections ^2`, they could use it.  Currently, this lib is blocking and problematic.

phansys:
The dev dependencies generating the mentioned conflict are `doctrine/mongodb-odm` and `doctrine/orm`. Are you thinking about a use case for this library where none of these packages are available?
From [README.md](https://github.com/doctrine-extensions/DoctrineExtensions/tree/9327c814f3aba3f3f565f080f2296efb92e4e72d#readme):
>This package contains extensions for Doctrine ORM and MongoDB ODM that offer new functionality or tools to use Doctrine more efficiently.

oojacoboo:
Without mongo - absolutely.  Without `doctrine/orm`, probably not.  But, this lib doesn't restrict the version that can be required anyway.  Irregardless, I've tested this fork/PR on our app which includes `doctrine/orm ~2.13` and `doctrine/collections` 2.1.0 gets "installed".

phansys:
I see. Maybe we should split our test suite in order to have separate jobs for both model managers.

PR #2540, opened 2022-11-29
Make LazyCollection extend AbstractLazyCollection

franmomu:
In `doctrine/collections` 2, `ReadableCollection` has two new methods: https://github.com/doctrine/collections/pull/252

We need them to be able to add compatibility with `doctrine/collections` 2

**Update**: It extends from `AbstractLazyCollection` as suggested in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2540#issuecomment-1359333873


stof:
It might be worth changing this LazyCollection to extend the `AbstractLazyCollection` of doctrine/collections (available in 1.2+), which will simplify this in the future (any new method would automatically be implemented in a way that trigger initialization before delegating it, needing an override only if you need a special behavior)

mbabker:
Just did a comparison, the `LazyCollection` here is pretty much a clone of the `AbstractLazyCollection` with a slightly different initialize method.

```php
class LazyCollection extends AbstractLazyCollection
{
    /**
     * @var callable
     */
    private $callback;

    /**
     * @param callable $callback
     */
    public function __construct($callback)
    {
        $this->callback = $callback;
    }

    protected function doInitialize(): void
    {
        $this->collection = call_user_func($this->callback);
    }
}
```

That's all this class needs to be.

phansys (approved):

PR #2542, opened 2022-11-30
[Sortable] Fix return value check of Comparable interface (#2541)

### LINKED ISSUE #2541: SortableListener uses wrong compareTo check if Comparable is used
In https://github.com/doctrine-extensions/DoctrineExtensions/blob/v3.10.0/src/Sortable/SortableListener.php#L266 

the sortable listener uses `compareTo` to find out if the object of the group is matching.

However, the corresponding interface https://github.com/doctrine/common/blob/3.4.3/src/Comparable.php

says it should return `0` upon a match and `1` / `-1` according to sorting as that interface could also be used for sorting.

In combination with https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Sortable/SortableListener.php#L275 the sortable listener expects `$matches` to be a boolean value (or at least 1 and not 0).

see also: 
https://github.com/doctrine-extensions/DoctrineExtensions/blob/ea1e0253f5c148754a525865c7277f38396033e9/src/Sortable/SortableListener.php#L460

There was a previous PR https://github.com/doctrine-extensions/DoctrineExtensions/pull/2185 but that was never merged - probably as the one comment was never fixed.

AndreasA:
Fixes: #2541

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Sortable: Fix return value check of Comparable interface (#2541)
+
+### Deprecated
+- Sortable: Accepting a return type other than "integer" from `Comparable::compareTo()` is deprecated in `SortableListener::postFlush()`. This will not be accepted in version 4.0.
+
 ## [3.10.0] - 2022-11-14
 ### Changed
 - Bump "doctrine/event-manager" dependency from ^1.0 to ^1.2.
--- a/src/Sortable/SortableListener.php
+++ b/src/Sortable/SortableListener.php
@@ -264,6 +264,17 @@ class SortableListener extends MappedEventSubscriber
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
                                     $matches = $gr->compareTo($value);
+                                    // @todo: Remove "is_int" check and only support integer as the interface expects.
+                                    if (is_int($matches)) {
+                                        $matches = 0 === $matches;
+                                    } else {
+                                        @trigger_error(sprintf(
+                                            'Support for "%s" as return type from "%s::compareTo()" is deprecated since'
+                                            .' gedmo/doctrine-extensions 3.x and will be removed in version 4.0. Return "integer" instead.',
+                                            gettype($matches),
+                                            Comparable::class
+                                        ), E_USER_DEPRECATED);
+                                    }
                                 } else {
                                     $matches = $gr == $value;
                                 }
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -441,6 +441,11 @@ parameters:
 			path: src/Sortable/Mapping/Driver/Yaml.php
 
 		-
+			message: "#^Else branch is unreachable because previous condition is always true\\.$#"
+			count: 1
+			path: src/Sortable/SortableListener.php
+
+		-
 			message: "#^Call to an undefined method Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadata\\:\\:getFieldMapping\\(\\)\\.$#"
 			count: 1
 			path: src/Sortable/Mapping/Driver/Yaml.php


src/Sortable/SortableListener.php
```diff
@@ -264,6 +265,10 @@ public function postFlush(EventArgs $args)
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
                                     $matches = $gr->compareTo($value);
+                                    // According to the interface "0" is returned, if the objects match.
+                                    // However, previously a bool was expected by the listener.
+                                    // Therefore, in order to avoid a breaking change, "true" and "0" are accepted.
+                                    $matches = 0 === $matches || true === $matches;
```
AndreasA, line 271: it is probably not 100% non breaking because in theory someone could have returned `1` to achieve the same result but that is contrary to the previous implementation and the PHPdoc of the interface.

phansys:
Thank you @AndreasA!

Even in the old doctrine/common 2.2 version, the [`Comparable::compareTo()`](https://github.com/doctrine/common/blob/d14fd323c9d8346334b8bc077287613a7113d6ef/lib/Doctrine/Common/Comparable.php#L34-L45) is declaring `int` as the only possible return type.
IMO, we have 2 choices:
1. Accept this BC break in a minor version with a clear message in our upgrade notes;
2. Add a `@todo` comment and trigger a `E_USER_DEPRECATED` error telling about the deprecation.

AndreasA:
@phansys I now trigger a deprecation and made a mention about the potential breaking change in the CHANGELOG.md (not sure that is the correct place). As in theory someone could have previously returned an integer value (to be in sync with the expected return type), but expected it to be evaluated truthy.

though that is an extreme edge case as it ignores the expected behavior of the Comparable and the behavior of the code.

The phpstan error should be fixed.

src/Sortable/SortableListener.php
```diff
@@ -264,6 +264,16 @@ public function postFlush(EventArgs $args)
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
                                     $matches = $gr->compareTo($value);
+                                    // @todo: Remove "boolean" check and only support integer as the interface expects.
+                                    if (is_int($matches)) {
+                                        $matches = 0 === $matches;
+                                    } else {
+                                        @trigger_error(sprintf(
+                                            '%1$s should return an integer. A %2$s was returned instead.',
```
franmomu, line 272:
```suggestion
                                            '%1$s::compareTo() method should return an integer. A %2$s was returned instead.',
```
pointing out the method would be nice

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Sortable: Fix return value check of Comparable interface (#2541)
+  - BC: The listener now expects the Comparable interface to return integer values according to PHPDoc. Truthy integer values will now produce invalid behavior and non integer values are deprecated.
```
franmomu, line 24: I would move this to a new  `### BC break` section I think

franmomu (commented):
I think I'm fine breaking BC (I guess it would affect few people)

AndreasA:
@franmomu Done.

franmomu (approved):

src/Sortable/SortableListener.php
```diff
@@ -264,6 +264,16 @@ public function postFlush(EventArgs $args)
...
                                 if ($gr instanceof Comparable) {
                                     $matches = $gr->compareTo($value);
+                                    // @todo: Remove "boolean" check and only support integer as the interface expects.
+                                    if (is_int($matches)) {
+                                        $matches = 0 === $matches;
+                                    } else {
+                                        @trigger_error(sprintf(
+                                            '%1$s::compareTo() method should return an integer. A %2$s was returned instead.',
+                                            Comparable::class,
+                                            gettype($matches)
```
phansys, line 274:
```suggestion
                                            'Support for "%s" as return type from "%s::compareTo()" is deprecated since'
                                            .' gedmo/doctrine-extensions 3.x and will be removed in version 4.0. Return "integer" instead.',
                                            gettype($matches),
                                            Comparable::class
```

phansys (changes requested):

CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- Sortable: Fix return value check of Comparable interface (#2541)
+
+### BC break
```
phansys, line 25: As we are not removing the support for other types, there is no BC break here.

src/Sortable/SortableListener.php
AndreasA: @phansys Done.

src/Sortable/SortableListener.php
```diff
@@ -264,6 +264,17 @@ public function postFlush(EventArgs $args)
                                 // Otherwise we fallback to normal object comparison
                                 if ($gr instanceof Comparable) {
                                     $matches = $gr->compareTo($value);
+                                    // @todo: Remove "boolean" check and only support integer as the interface expects.
```
phansys, line 267: IIUC, there is no boolean check here.

AndreasA: good catch 😄

AndreasA: changed it to "is_int".

CHANGELOG.md
AndreasA: @phansys strictly speaking if someone would return `1` and expect it to be evaluated as `truthy` it would fail now as it would result in `false` which strictly speaking is a change from before. 

However, it is an edge case as someone used the correct return type but invalid behavior which makes no sense.

phansys (approved):

phansys:
Thank you @AndreasA! :tada:

AndreasA:
@phansys Any idea when this fix will be released? I am asking because I currently use this in a project and I would like to switch to using the correct return value. 

Thanks for reviewing and merging it so fast.

phansys:
@franmomu, do you have any PR that we should look at before the next release?

franmomu:
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2447 this has been waiting for long (it needs to be rebased) and review again, I'll try later today to have a look.

PR #2543, opened 2022-12-04
Add generics

Warxcell:


src/Tree/RepositoryInterface.php
```diff
@@ -25,39 +27,39 @@ interface RepositoryInterface extends RepositoryUtilsInterface
...
      *
-     * @param object|null          $node        If null, all tree nodes will be taken
+     * @param T|null          $node        If null, all tree nodes will be taken
      * @param bool                 $direct      True to take only direct children
      * @param string|string[]|null $sortByField Field name or array of fields names to sort by
      * @param string|string[]      $direction   Sort order ('ASC'|'DESC'). If $sortByField is an array, this may also be an array with matching number of elements
      * @param bool                 $includeNode Include the root node in results?
      *
-     * @return array|null List of children or null on failure
+     * @return T[]|null List of children or null on failure
```
7ochem, line 55: Maybe this could be `iterable<array-key, T>` to allow any iterable to be returned?

stof: this would widen the return type, which is a BC break

franmomu:
closing here since https://github.com/doctrine-extensions/DoctrineExtensions/pull/2556 was merged, thanks @Warxcell

PR #2546, opened 2022-12-12
Add `Nested::ALLOWED_NODE_POSITIONS` constant in order to expose the available node positions

phansys:


src/Tree/Strategy/ORM/Nested.php
```diff
@@ -63,22 +71,26 @@ class Nested implements Strategy
...
     private $treeEdges = [];
 
     /**
      * Stores a list of node position strategies
      * for each node by object id
      *
-     * @var array
+     * @var array<int, string>
+     *
+     * @phpstan-var array<int, value-of<self::ALLOWED_NODE_POSITIONS>>|array{}
```
franmomu, line 84: do we need the `|array{}` part?

src/Tree/Strategy/ORM/Nested.php
```diff
@@ -63,22 +71,26 @@ class Nested implements Strategy
...
      */
     private $nodePositions = [];
 
     /**
      * Stores a list of delayed nodes for correct order of updates
      *
-     * @var array
+     * @var array<int, array<int, array<string, Node|object|string>>>
+     *
+     * @phpstan-var array<int, array<int, array{node: Node|object, position: value-of<self::ALLOWED_NODE_POSITIONS>}>>|array{}
```
franmomu, line 93: same as above about the `|array{}` part.

src/Tree/Strategy/ORM/Nested.php
franmomu: hmm I think in that example is necessary because it is described with an array shape, but in this case it is a "general" array, it would be like [this example](https://phpstan.org/r/b852a8b0-2a50-4d01-ba5c-b93528acb17b) I think.

phansys: You're right. Thank you!

franmomu:
Thanks @phansys!

PR #2544, opened 2022-12-05
Allow PHP 8.2 in CI and Docker environments

phansys:
**TODO**:
- [x] Wait for #2547.

~~Some findings from the installed version of PHP-CS-Fixer in the pipeline of this PR were also fixed (settings for `phpdoc_order` and `phpdoc_separation` rules were explicitly configured in order to avoid more changes).~~

PR #2547, opened 2022-12-12
Tree: Add `setSibling()` and `getSibling()` methods in the `Node` interface

phansys:
In PHP 8.2, [the use of dynamic properties is deprecated](https://www.php.net/manual/en/migration82.deprecated.php#migration82.deprecated.core.dynamic-properties).

See [this action](https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/3622870152/jobs/6157207902#step:7:32) from #2544.

### DIFF
--- a/src/Tree/Node.php
+++ b/src/Tree/Node.php
@@ -14,6 +14,9 @@ namespace Gedmo\Tree;
  * Entities which in some cases needs to be identified as
  * Tree Node
  *
+ * @method void  setSibling(self $node)
+ * @method ?self getSibling()
+ *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  */
 interface Node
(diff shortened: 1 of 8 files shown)


src/Tree/Node.php
```diff
@@ -42,4 +45,10 @@ interface Node
      * @gedmo:TreeLevel
      * level of node.
      */
+
+    // @todo: In the next major release, remove this line and uncomment the method in the next line.
+    // public function setSibling(self $node);
```
franmomu, line 50:
```suggestion
    // public function setSibling(self $node): void;
```

franmomu (approved):

src/Tree/Node.php
phansys: Done. Thank you!

franmomu:
thanks @phansys!

PR #2548, opened 2022-12-12
Minor typing improvements

phansys:
* ~~`Doctrine\Common\Annotations\Reader` was declared were possible, as the interface is currently available in all the supported versions of [`doctrine/annotations`](https://github.com/doctrine/annotations/blob/03cb2123a67d4be806554fe670d0adc298199808/lib/Doctrine/Common/Annotations/Reader.php#L12)~~.
   This change should wait for #2258;
* Private and read only properties `$validTypes` were replaced by `VALID_TYPES` private constants.

### DIFF
--- a/src/References/Mapping/Driver/Annotation.php
+++ b/src/References/Mapping/Driver/Annotation.php
@@ -9,11 +9,13 @@
 
 namespace Gedmo\References\Mapping\Driver;
 
+use Doctrine\Common\Annotations\Reader;
 use Doctrine\Persistence\Mapping\Driver\MappingDriver;
 use Gedmo\Mapping\Annotation\ReferenceMany;
 use Gedmo\Mapping\Annotation\ReferenceManyEmbed;
 use Gedmo\Mapping\Annotation\ReferenceOne;
 use Gedmo\Mapping\Driver\AnnotationDriverInterface;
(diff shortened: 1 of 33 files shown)


src/Tool/WrapperInterface.php
```diff
@@ -63,7 +65,7 @@ public function hasValidIdentifier();
     /**
      * Get the object metadata.
      *
-     * @return ClassMetadata
+     * @return ClassMetadata&(OrmClassMetadata|OdmClassMetadata)
```
franmomu, line 68: Can we add a `@template TClassMetadata of ClassMetadata` to the interface? so `EntityWrapper` and `MongoDocumentWrapper` could specify their own `ClassMetadata`.

Not sure if it would solve other issues or maybe it would create more 😬

franmomu (approved):

src/References/Mapping/Driver/Annotation.php
```diff
@@ -61,7 +62,7 @@ class Annotation implements AnnotationDriverInterface
     /**
      * Annotation reader instance
      *
-     * @var object
+     * @var Reader
```
phansys, line 65: I need to confirm this.

phansys: It was updated to `Reader|AttributeReader|object`.

PR #2549, opened 2022-12-13
Force load proxy

dmitryuk:
If where are some relocations while update and object is proxy it is impossible to get value on line 260 - it always return `null`

Finally it cause the error
```
//vendor/gedmo/doctrine-extensions/src/Mapping/MappedEventSubscriber.php:264
Cannot call recomputeSingleEntityChangeSet before computeChangeSet on an entity.
```

```
$pos = $meta->getReflectionProperty($config['position'])->getValue($object);
```
image

After entry is force-loaded I've got what expected
image

### DIFF
--- a/src/Sortable/SortableListener.php
+++ b/src/Sortable/SortableListener.php
@@ -15,6 +15,7 @@ use Doctrine\Common\Util\ClassUtils;
 use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
+use Doctrine\Persistence\Proxy;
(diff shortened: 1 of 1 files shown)


src/Sortable/SortableListener.php
```diff
@@ -236,6 +237,10 @@ public function postFlush(EventArgs $args)
                             continue;
                         }
 
+                        if ($object instanceof Proxy && !$object->__isInitialized()) {
+                            $object->__load();
```
phansys, line 241: I wonder what impact this could produce in terms of performance :thinking:

phansys (changes requested):
Thank you for your first contribution @dmitryuk!

In order to be merged, this PR requires a test case covering the described issue.
Additionally, an entry in `CHANGELOG.md` is needed in order to explain the users what is fixed.

PR #2551, opened 2022-12-20
[documentation] Updates docker-compose related commands in README

JoolsMcFly:
Hi,

`docker compose` needs to be changed to `docker-compose` and since you are not using the default `docker-compose.yaml` file then we need to use `-f ./compose.yaml` for commands to work.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,9 +96,9 @@ XML mapping xsd schemas are also versioned and can be used by version suffix:
 
 To set up and run the tests, follow these steps:
 
-- Install [Docker](https://www.docker.com/) and ensure you have `docker compose`
-- From the project root, run `docker compose up -d` to start containers in daemon mode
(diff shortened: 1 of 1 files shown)


phansys:
Thank you @JoolsMcFly.
Please, note that the current implementation is using the [Docker Compose plugin](https://docs.docker.com/compose/install/linux/), which provides the `docker compose` command.
About the compose file, `compose.yaml` is currently the first file that the command looks for. See [Compose file](https://docs.docker.com/compose/compose-file/#compose-file).

JoolsMcFly:
Right, so ... useless PR?

phansys:
IMO, not useless at all, but I think we can safely close this one based on our current implementation.
Thank you very much!

JoolsMcFly:
Closing as `docker compose` is legit! :)

PR #2556, opened 2022-12-23
[Tree] Add `@template` and `@template-extends` annotations to the Tree repositories

7ochem:
I've added `@template T of object` and `@template-extends EntityRepository<T>` annotations to the Tree repositories.

In my code I can do:
```php
/**
 * @template-extends \Doctrine\ORM\EntityRepository<Foo>
 */
class FooRepository extends \Doctrine\ORM\EntityRepository
{
```

This is because `EntityRepository` has a `@template` annotation:
https://github.com/doctrine/orm/blob/f8bf84d1aa8b67f13f9a5c1c538938eb77e66bd6/lib/Doctrine/ORM/EntityRepository.php#L36

Now static analysis and type hints can evaluate that `$fooRepository->find(123)` will return a `Foo` object.

But I can't do the same in my code for any repository extending a Tree repository:
```php
/**
 * @template-extends \Gedmo\Tree\Entity\Repository\NestedTreeRepository<Foo>
 */
class FooRepository extends \Gedmo\Tree\Entity\Repository\NestedTreeRepository
{
```

These changes enable that. If you think this is a good idea, then I could also add `@phpstan-return ?list<T>` on `children()` for example to indicate that it will return an array of `T`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- Tree: Added `@template` and `@template-extends` annotations to the Tree repositories
+
 ### Changed
 - Dropped support for PHP < 7.4
 - Dropped support for Symfony < 5.4
--- a/example/app/Entity/Repository/CategoryRepository.php
+++ b/example/app/Entity/Repository/CategoryRepository.php
@@ -11,8 +11,12 @@ declare(strict_types=1);
 
 namespace App\Entity\Repository;
 
+use App\Entity\Category;
 use Gedmo\Tree\Entity\Repository\NestedTreeRepository;
 
+/**
+ * @template-extends NestedTreeRepository<Category>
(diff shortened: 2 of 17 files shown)


franmomu:
Thanks for the PR @7ochem!

> These changes enable that. If you think this is a good idea, then I could also add `@phpstan-return ?list<T>` on `children()` for example to indicate that it will return an array of `T`.

Sure, if we specify the template I would use it when possible (like in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2543) and what about use `TNode` instead of `T`?

ro0NL (approved):
please merge :)

ro0NL (approved):

ro0NL (approved):

ro0NL (approved):

franmomu (approved):

franmomu:
thanks @7ochem!

PR #2557, opened 2022-12-25
Update composer.json

sinus22:


phansys:
Duplicates #2537.
Depends on #2540.

PR #2558, opened 2022-12-27
Deprecate DoctrineExtensions::registerAnnotations()

mbabker:
The `Gedmo\DoctrineExtensions::registerAnnotations()` method is used to manually load this package's annotation classes.  Since `doctrine/annotations` 1.10, the package has supported autoloading in full and deprecated these manual loading mechanisms, with them removed in the Annotations 2.0 release.

This PR, therefore, deprecates `Gedmo\DoctrineExtensions::registerAnnotations()` in order to remove it in 4.0 as it no longer serves a purpose.  It also no-op's the method to help unblock Annotations 2.0 support, and removes the uses of this newly deprecated method and the deprecated Annotations loader methods in the example app and test bootstrap.

### DIFF
--- a/src/Mapping/Annotation/All.php
+++ b/src/Mapping/Annotation/All.php
@@ -7,6 +7,11 @@
  * file that was distributed with this source code.
  */
 
+@trigger_error(sprintf(
+    'Requiring the file at "%s" is deprecated since gedmo/doctrine-extensions 3.11, this file will be removed in version 4.0.',
+    __FILE__
+), E_USER_DEPRECATED);
+
 // Contains all annotations for extensions
 // NOTE: should be included with require_once
 foreach (glob(__DIR__.'/*.php') as $filename) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ a release.
 - Sortable: Accepting a return type other than "integer" from `Comparable::compareTo()` is deprecated in `SortableListener::postFlush()`. This will not be accepted in version 4.0.
 - Deprecate the annotation reader being allowed to be any object.
   In 4.0, a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` instance will be required.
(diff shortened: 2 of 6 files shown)


franmomu (approved):
We should probably deprecate the [All.php](https://github.com/doctrine-extensions/DoctrineExtensions/blob/fae671d499c16e7fb69a79adf6c99f9e97d060e8/src/Mapping/Annotation/All.php) file as well.

mbabker:
> We should probably deprecate the [All.php](https://github.com/doctrine-extensions/DoctrineExtensions/blob/fae671d499c16e7fb69a79adf6c99f9e97d060e8/src/Mapping/Annotation/All.php) file as well.

Done

src/Mapping/Annotation/All.php
```diff
@@ -7,6 +7,11 @@
  * file that was distributed with this source code.
  */
 
+@trigger_error(sprintf(
+    'Requiring the file at "%s()" is deprecated since gedmo/doctrine-extensions 3.11, this file will be removed in version 4.0.',
+    __METHOD__
```
phansys, line 12: How can we guarantee this scripts is always loaded inside a method?

mbabker: I'm not quite following here.

Including this file is not needed since Annotations 1.10 as it will fall back to using the autoloader, with support for manual requires in the AnnotationRegistry deprecated and removed in 2.0.  This package requires Annotations 1.13, so there isn't a need for supporting the older behavior.

To avoid unfixable deprecation notices, this file is no longer included anywhere in this package's code, with the notice being there if someone is loading it in their own code.

phansys: I mean, we haven't a way to be sure this script was loaded from a method. In that case, the resulting message could be confusing. Check [this example](https://3v4l.org/Tl7FG).

mbabker: OH!  🤦 

That was supposed to be `__FILE__` but clearly I not only messed up the copy/paste, but couldn't even read the code right on your comment 😅 

It's fixed.

phansys (approved):

phansys:
Thank you @mbabker!

Issue #2554, opened 2022-12-22
Make Annotations an optional dependency


mbabker:
With the adoption of PHP 8 increasing and the ecosystem moving toward supporting attributes over annotations (with the ORM going so far as to have deprecated annotation support in 2.x and [removing it in 3.x](https://github.com/doctrine/orm/blob/3.0.x/UPGRADE.md#bc-break-removed-annotationdriver), maybe now is a good time to look at making the `doctrine/annotations` package an optional dependency.

Just on a quick scan, the biggest reasons that the Annotations package looks to be a hard dependency include:

- Not providing XML mappings for any of this package's objects (loggable and translatable for both drivers, and tree for the ORM)
- The helper methods in the `Gedmo\DoctrineExtensions` class only support registering an annotation reader (the objects being registered do have both annotations and attributes on them and current versions of both the ORM and ODM have attribute drivers)
- The `Gedmo\Mapping\ExtensionMetadataFactory` constructor has a hard dependency to either an annotation or attribute reader (but there are `instanceof` checks in the code paths where the reader would be used and it's not necessary if you're using the XML or YAML drivers)
    - As a result of this, `Gedmo\Mapping\MappedEventSubscriber::getExtensionMetadataFactory()` would fail hard because its internals expect either an annotation or attribute reader to have been injected or for it to be able to create a default annotation reader

mbabker:
Hardly stale, just not recently active 😉

PR #2559, opened 2022-12-27
Allow `Gedmo\DoctrineExtensions` to use an attribute driver

mbabker:
Ref: #2554

Changes the `Gedmo\DoctrineExtensions` class to conditionally use an annotation or attribute driver based on the runtime PHP version, with the attribute driver automatically being used on PHP 8.0 or later.

And because Codecov was yelling about the drop in coverage, the `Gedmo\DoctrineExtensions` class now has tests.

### DIFF
--- a/src/DoctrineExtensions.php
+++ b/src/DoctrineExtensions.php
@@ -15,6 +15,7 @@ use Doctrine\Common\Annotations\Reader;
 use Doctrine\ODM\MongoDB\Mapping\Driver as DriverMongodbODM;
 use Doctrine\ORM\Mapping\Driver as DriverORM;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
+use Gedmo\Exception\RuntimeException;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 /**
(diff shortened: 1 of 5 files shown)


stof:
Do we need the explicit opt-in for attributes in the helper method ? Given that the mapping provided by attributes and by annotations in the extension classes is totally equivalent, I see no BC impact of registering an attribute driver instead for supported cases (i.e. PHP 8+ with a version of the ORM/ODM supporting attributes).
The driver registered by those helpers will have no impact on the kind of driver used by the entities of the project (as the registered driver is not handling them).

mbabker:
TBH, I don't actually know.  I've never used this library in a non-Symfony context, so what I had was just a bit of guessing and hoping it worked.  CI passes with it changed to be PHP version conditional, so maybe this can work just fine.

src/DoctrineExtensions.php
```diff
@@ -36,15 +37,19 @@ final class DoctrineExtensions
...
-        $annotationDriver = new DriverORM\AnnotationDriver($reader, [
+        $paths = [
             __DIR__.'/Translatable/Entity',
             __DIR__.'/Loggable/Entity',
             __DIR__.'/Tree/Entity',
-        ]);
-        $driverChain->addDriver($annotationDriver, 'Gedmo');
+        ];
+
+        if (PHP_VERSION_ID >= 80000) {
```
stof, line 46: I suggest using `\PHP_VERSION_ID` or `use const PHP_VERSION_ID` (but I'm not sure the second version is supported in all PHP versions supported by this library) instead of relying on the fallback to the global namespace. For a qualified constant coming from the core, PHP will resolve it at compile time, which will then allow OPCache to resolve the whole `if` at compile time.

mbabker: Done.

stof (approved):

phansys (approved):
Thank you @mbabker!

PR #2563, opened 2023-01-02
Allow doctrine/annotations 2.0

mbabker:
The ORM's 2.14 release supports it and MongoDB ODM's 2.5 will add support.  The PHP-CS-Fixer tool will block installing the 2.0 version in a dev environment, and if https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/6730 is merged as is as of the time I opened this PR, there won't be a release of that tool supporting both Annotations versions at the same time.

### DIFF
--- a/tests/Gedmo/Mapping/SluggableMappingTest.php
+++ b/tests/Gedmo/Mapping/SluggableMappingTest.php
@@ -12,7 +12,6 @@ declare(strict_types=1);
 namespace Gedmo\Tests\Mapping;
 
 use Doctrine\Common\Annotations\AnnotationReader;
-use Doctrine\Common\Annotations\AnnotationRegistry;
 use Doctrine\Common\EventManager;
 use Doctrine\ORM\Mapping\Driver\YamlDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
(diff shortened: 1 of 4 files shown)


franmomu:
What about removing `PHP-CS-Fixer` in [`continuous-integration.yml`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/2bf3ad55001bff04f283fdbb7b057305cbb16e27/.github/workflows/continuous-integration.yml)? So we can run the tests with `doctrine/annotations` 2 installed.

mbabker:
> What about removing `PHP-CS-Fixer` in [`continuous-integration.yml`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/2bf3ad55001bff04f283fdbb7b057305cbb16e27/.github/workflows/continuous-integration.yml)? So we can run the tests with `doctrine/annotations` 2 installed.

Done

phansys:
Could you please share the CI job where "doctrine/annotations" is resolved to 2.x?

mbabker:
The ODM 2.5 dev job is installing it at 2.0.

phansys:
Thank you!

https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/3831637320/jobs/6521000161#step:8:95 :tada:

tests/Gedmo/Mapping/SluggableMappingTest.php
```diff
@@ -49,10 +48,6 @@ protected function setUp(): void
             'Gedmo\Tests\Mapping\Fixture\Yaml'
         );
         $reader = new AnnotationReader();
-        AnnotationRegistry::registerAutoloadNamespace(
-            'Gedmo\\Mapping\\Annotation',
-            dirname(VENDOR_PATH).'/src'
```
phansys, line 54: I think this is the only use for this constant. IMO, we can also safely remove its definition.

mbabker: Done

phansys (approved):

franmomu (approved):

franmomu:
thanks @mbabker!

PR #2564, opened 2023-01-02
Drop support for Symfony 5.3 and older

mbabker:
Symfony 4.4 is now only [getting security fixes](https://symfony.com/releases/4.4) and 5.4 is the next LTS.  This will drop support for Symfony 4.4 and 5.3 in the handful of components that are used.


phansys:
I'd like to publish a release with the pending fixes before merging this PR.
`/cc` @franmomu.

phansys (approved):

franmomu:
thanks @mbabker!

PR #2566, opened 2023-01-04
Don't wrap the reader if it is already an attribute reader

mbabker:
`Gedmo\Mapping\ExtensionMetadataFactory` will handle pushing a reader into the mapping driver when it's working with an attribute or annotation reader.  When an attribute driver is detected, it tries to provide a `Gedmo\Mapping\Driver\AttributeAnnotationReader` with the injected reader.  However, the second argument of the `Gedmo\Mapping\Driver\AttributeAnnotationReader` constructor requires a `Doctrine\Common\Annotations\Reader`.  So if the extension metadata factory is instantiated with a `Gedmo\Mapping\Driver\AttributeReader` passed through its constructor, things fail with a type error:

```sh
Gedmo\Mapping\Driver\AttributeAnnotationReader::__construct(): Argument #2 ($annotationReader) must be of type Doctrine\Common\Annotations\Reader, Gedmo\Mapping\Driver\AttributeReader given, called in /app/vendor/gedmo/doctrine-extensions/src/Mapping/ExtensionMetadataFactory.php on line 208
```

This adds a check to determine if the reader needs to be wrapped; if a `Gedmo\Mapping\Driver\AttributeReader` was provided, then we don't need to wrap the reader.

### DIFF
--- a/tests/Gedmo/Loggable/AttributeLoggableEntityTest.php
+++ b/tests/Gedmo/Loggable/AttributeLoggableEntityTest.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Loggable;
+
+use Doctrine\Common\EventManager;
+use Gedmo\Mapping\Driver\AttributeReader;
+use Gedmo\Tests\Loggable\LoggableEntityTest;
+
+/**
+ * These are tests for loggable behavior with an attribute reader
+ *
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ */
+final class AttributeLoggableEntityTest extends LoggableEntityTest
+{
+    public static function setUpBeforeClass(): void
+    {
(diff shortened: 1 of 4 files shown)


franmomu (approved):
I wonder why PHPStan didn't complain about this, maybe because of the current level.

tests/Gedmo/Loggable/AttributeLoggableEntityTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Gedmo\Loggable;
+
+use Doctrine\Common\EventManager;
+use Gedmo\Mapping\Driver\AttributeReader;
+use Gedmo\Tests\Loggable\LoggableEntityTest;
+
+/**
+ * These are tests for loggable behavior with an attribute reader
+ *
```
phansys, line 20:
```suggestion
 *
 * @requires PHP >= 8.0
 *
```

tests/Gedmo/Loggable/AttributeLoggableEntityTest.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+final class AttributeLoggableEntityTest extends LoggableEntityTest
+{
+    public static function setUpBeforeClass(): void
+    {
+        if (PHP_VERSION_ID < 80000) {
+            static::markTestSkipped('Test requires PHP 8');
+        }
+    }
+
```
phansys, line 31:
```suggestion
```

phansys (approved):

phansys:
Thank you @mbabker!

PR #2567, opened 2023-01-04
[Loggable] Add `LogEntryInterface`

phansys:
* Add `LogEntryInterface` interface in order to be implemented by log entry models;
* Deprecate `LoggableListener::ACTION_CREATE`, `LoggableListener::ACTION_UPDATE` and `LoggableListener::ACTION_REMOVE` constants in favor of `LogEntryInterface::ACTION_CREATE`, `LogEntryInterface::ACTION_UPDATE` and `LogEntryInterface::ACTION_REMOVE`.

These changes are proposed based on the work I made in #2548.

### DIFF
--- a/src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
+++ b/src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
@@ -11,12 +11,17 @@ namespace Gedmo\Loggable\Entity\MappedSuperclass;
 
 use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Loggable\LogEntryInterface;
 
 /**
+ * @phpstan-template T of object
+ *
+ * @phpstan-implements LogEntryInterface<T>
+ *
  * @ORM\MappedSuperclass
  */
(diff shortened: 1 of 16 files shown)


src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
```diff
@@ -96,153 +105,80 @@ public function getId()
...
-     */
     public function getAction()
     {
         return $this->action;
     }
 
-    /**
-     * Set action
-     *
-     * @param string $action
```
franmomu, line 112: should we keep the `@param` info? in some places is kept, apparently PHPStan complains if it is not from level 6: https://phpstan.org/r/e1a70fb8-4b4d-48c4-bb16-7fd65e2e5d5a

phansys: Thank you for pointing.
I reverted the removal. The few methods that kept the annotations were the only ones that were reported by PHPStan (I don't know why).

franmomu (approved):

franmomu:
thanks @phansys!

PR #2574, opened 2023-01-26
Remove `@final` annotation from repository classes

### LINKED ISSUE #2573: [Sortable] SortableRepository is final
With the latest release the repository for sortables is marked final.

The docs state:
> You should register SortableRepository (**or a subclass**) …

How is the SortableRepository supposed to be used now after classes are final?

It this an oversight and we can still extend from SortableRepository and ignore the final annotation?
Should we handle our own repo's constructor like a decorator?

mbabker:
Fixes #2573

Doctrine repository classes by design need to support inheritance as you can only have a single repository class per managed object, and using a decorator pattern for the repository classes in this package gets to be somewhat complicated as long as they extend from the root object repository classes in the MongoDB ODM and ORM.  So, these classes should not be treated as final.

### DIFF
--- a/src/Loggable/Document/Repository/LogEntryRepository.php
+++ b/src/Loggable/Document/Repository/LogEntryRepository.php
@@ -20,8 +20,6 @@ use Gedmo\Tool\Wrapper\MongoDocumentWrapper;
  * to interact with log entries.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class LogEntryRepository extends DocumentRepository
 {
--- a/src/Loggable/Entity/Repository/LogEntryRepository.php
+++ b/src/Loggable/Entity/Repository/LogEntryRepository.php
@@ -21,8 +21,6 @@ use Gedmo\Tool\Wrapper\EntityWrapper;
  * to interact with log entries.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class LogEntryRepository extends EntityRepository
 {
--- a/src/Sortable/Entity/Repository/SortableRepository.php
+++ b/src/Sortable/Entity/Repository/SortableRepository.php
@@ -18,8 +18,6 @@ use Gedmo\Sortable\SortableListener;
  * Sortable Repository
  *
  * @author Lukas Botsch <lukas.botsch@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class SortableRepository extends EntityRepository
 {
--- a/src/Translatable/Document/Repository/TranslationRepository.php
+++ b/src/Translatable/Document/Repository/TranslationRepository.php
@@ -25,8 +25,6 @@ use Gedmo\Translatable\TranslatableListener;
  * to interact with translations.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class TranslationRepository extends DocumentRepository
 {
--- a/src/Translatable/Entity/Repository/TranslationRepository.php
+++ b/src/Translatable/Entity/Repository/TranslationRepository.php
@@ -24,8 +24,6 @@ use Gedmo\Translatable\TranslatableListener;
  * to interact with translations.
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class TranslationRepository extends EntityRepository
 {
--- a/src/Tree/Document/MongoDB/Repository/MaterializedPathRepository.php
+++ b/src/Tree/Document/MongoDB/Repository/MaterializedPathRepository.php
@@ -22,8 +22,6 @@ use MongoDB\BSON\Regex;
  *
  * @author Gustavo Falco <comfortablynumb84@gmail.com>
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class MaterializedPathRepository extends AbstractTreeRepository
 {
--- a/src/Tree/Entity/Repository/ClosureTreeRepository.php
+++ b/src/Tree/Entity/Repository/ClosureTreeRepository.php
@@ -23,8 +23,6 @@ use Gedmo\Tree\Strategy;
  *
  * @author Gustavo Adrian <comfortablynumb84@gmail.com>
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
- *
- * @final since gedmo/doctrine-extensions 3.11
  */
 class ClosureTreeRepository extends AbstractTreeRepository
 {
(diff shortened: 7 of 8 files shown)


franmomu (approved):
makes sense 👍

phansys:
IMO, a better approach would be to make abstract the classes that we allow to be extended, and provide basic implementation for them with final classes. Otherwise, composition should be used instead.
WDYT?

mbabker:
For the Doctrine repositories specifically, that still goes against their design.

You can use the root `Doctrine\ODM\MongoDB\Repository\DocumentRepository` or `Doctrine\ORM\EntityRepository` classes as your entity's repository class, at least for the ORM that's the default behavior if you don't explicitly configure a repository class.  IMO making something that subclasses them abstract or final adds extra friction to using those classes.  Final repository classes might be OK in an application, but in a library, it doesn't really work well.

If the intention is to provide some generic functionality that can optionally be used with managed objects, and have that be a closed-off implementation (i.e. final), then you need some kind of helper class that's not a `Doctrine\Persistence\ObjectRepository` object.  As long as this functionality lives in a Doctrine repository class, the abstract or final modifiers just do not fit in well.

kevinpapst:
@phansys the commit https://github.com/doctrine-extensions/DoctrineExtensions/commit/35826403baf7ca06a094b808046793cd2119fc5b says `Closing the API by marking some classes as final and internal`. I guess `some` doesn't truly match here 😁  The commit was a pretty big BC break. I'd assume that these changes will cause issues (and I might be overexaggerating here) for almost anyone who is using this package.

Let me give you an example, and sorry for going slightly offtopic here @mbabker, but ONLY reverting that for repositories isn't really enough, as @develth mentioned in #2573

There is for example the `LoggableListener` which offers the protected method `prePersistLogEntry` as extension point in https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Loggable/LoggableListener.php#L224

This class is, in its current state, written for inheritance and not for composition. Can you share how you see the upgrade path for this specific case and/or all other repository and listener `finals`? I would vote for reverting the finals at least from the listeners as well UNTIL the class are rewritten in a way that makes them re-usable.

phansys:
Thank you very much for sharing your concerns.

@mbabker, I'd like to analyze what you are mentioning in order to check our options here.

@kevinpapst, about your thoughts:
>The commit was a pretty big BC break.

AFAIK, there is no BC break until these classes are declared as final. The `@final` tag is part of a deprecation path, which allows us to keep BC until the next major release. Are you experiencing a different behavior?

>I'd assume that these changes will cause issues (and I might be overexaggerating here) for almost anyone who is using this package.

I understand that closing the API may break some implementations that rely on the inheritance of these classes. That's why a deprecation path is provided instead of making a change like this in a MINOR release.
IMO, viewing this like an over-exaggerated move is acceptable from a user-land perspective, but from a maintainer point of view, any class that is extensible from the consumer's implementations is a blocker for confident fixes and improvements without being afraid of break BC.

>There is for example the LoggableListener which offers the protected method prePersistLogEntry as extension point in https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Loggable/LoggableListener.php#L224
>
>This class is, in its current state, written for inheritance and not for composition. Can you share how you see the upgrade path for this specific case and/or all other repository and listener finals? I would vote for reverting the finals at least from the listeners as well UNTIL the class are rewritten in a way that makes them re-usable.

I didn't think about this yet, but I guess we could use the same approach I mentioned before for repositories:
1. Create an abstract `AbstractLoggableListener` class (maybe with the contents of the current `LoggableListener` class);
2. Update `LoggableListener` with the default concrete implementation, declaring it as final.

Please, let me know if you think there is some blocker, oversight or misconception in what I am proposing.

develth:
hi @phansys 

> AFAIK, there is no BC break until these classes are declared as final. The @final tag is part of a deprecation path, which allows us to keep BC until the next major release. Are you experiencing a different behavior?

Yes, with build tools and code quality checks the Pipeline fails because of @fail annotation

aleho:
@develth I've had to deal with several libs in the past that made a concrete implementation final with opinionated behavior that clearly would not apply to a broad range of projects. This always lead to the problem you described: having to copy an entire implementation, duplicating it's boiler-plate code just to change one or two lines.

In one project I now have dozens of these classes, all annotated with PHPDoc to remind be of why I copy/pasted the original implementation. Even worse, it's very hard to catch upstream changes to my copied code, so essentially every minor release is potentially breaking without me even noticing or not getting important fixes (I spend days reading through all the changes).

In my experience it's always better to break BC by bumping the major version number while allowing non-final extension points for your lib. Just adding an `@final` or `final class` without providing well thought out extension points (via e.g. DI) for the opinionated parts of your code – which I admit is pretty hard to get right – is not the right way to go forward.

TL;DR: I understand why, but please don't `final` implementations without extension points.

develth:
> In one project I now have dozens of these classes, all annotated with PHPDoc to remind be of why I copy/pasted the original implementation. Even worse, it's very hard to catch upstream changes to my copied code, so essentially every minor release is potentially breaking without me even noticing or not getting important fixes (I spend days reading through all the changes).

I am also unsure about the legal setup of this. IMHO we would need to fork the project to make our changes based on the original code public available?

stof (approved):

kevinpapst:
> AFAIK, there is no BC break until these classes are declared as final. The @final tag is part of a deprecation path, which allows us to keep BC until the next major release. Are you experiencing a different behavior?

Yes. Many broken CI builds. Which is good on one hand, so I know there is something coming up. But there is no "deprecation path". Making something final without defining a way forward feels a bit rushed. 

Now I see only three options:
Copy & paste the code from the listener into my own project (this is a terrible idea) or pin the composer dependency (which is not better, as it doesn't help your long term goal) or add something to PHPStan baseline (which is also not better then before, hidden problems usually don't go away by themselves).

Maybe the feedback shows that adding `final` to over 120 classes probably was only the beginning of this journey. I'd vote for a partial revert and a patch release, until a real way forward is worked out.

kevinpapst (approved):

mbabker:
> But there is no "deprecation path". Making something final without defining a way forward feels a bit rushed.

Well, it's hit or miss.  The `@final` annotation communicates an intent to add the engine keyword (or communicates the intent that the class is treated as final by the project but the engine keyword won't be added so extend at your own risk), so it does define a way forward in the sense that inheritance won't be supported.  But, clearly, some valid use cases weren't considered when adding the annotation (which is fine, PRs like this one help fix things).

The pain point with changes like adding the `@final` annotation is we have a lot more tooling these days that will more eagerly throw warnings at us when those changes land (i.e. Symfony's ErrorHandler component has some checks for `@final` annotated classes, and there are warnings in PHPStan and Psalm).  If we use that as the benchmark for B/C breaks, then we're all stuck between a rock and a hard place; if a new PHPStan warning counts as a break, then any `@phpstan-*` type annotation that's added outside a major release could be seen as a B/C break.  Finding the right balance in how to move things forward is hard, you either don't document the planned change (which essentially counts as a deprecation notice) or you ship a major version with a bunch of hard breaks that nobody had time to prep for to avoid new warnings in a non-major release.

This PR and a patch for the `LoggableListener` (which until https://github.com/doctrine-extensions/DoctrineExtensions/pull/2574#issuecomment-1407247610 I didn't realize had the `prePersistLogEntry` hook and I don't have a project using Doctrine without the loggable extension from here) covers the lowest hanging fruit, anything else is really going to need its own discussion on how to best deal with those.

phansys (approved):
I have no time to propose the next step for the replacements on these classes now, so I'll merge this as is.
I'll try to provide a sample approach the next weeks, as I'm currently on vacation.

phansys:
Thank yoi @mbabker!

PR #2575, opened 2023-01-27
fix: possibility to match "MappingDriverChain" for doctrine 2.14.1

develth:
with doctrine 2.14.1 the driverName is `MappingDriverChain`, that results in the `MetadataFactory` with exception `... extension driver was not found`


develth:
nvmd, doesn´t fixed the issue, check for mapped was there already

PR #2579, opened 2023-02-02
Remove @final annotation from listener class

kevinpapst:
The `LoggableListener` class has a protected method, which is a clear inheritance extension point:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/a2364710d2f367f74d9f041c3356f2929b8c6690/src/Loggable/LoggableListener.php#L215-L226

Until there is no clear solution offered, to achieve the same goals without inheritance (or copy&pasting code), the simplest solution is to revert the `final`.

See discussion in #2574


mbabker (approved):

franmomu (approved):
I think this is fine until we find a way to configure the extension with your own listeners easily

phansys (approved):

phansys:
Thank you @kevinpapst!

kevinpapst:
Thanks for merging this temporary fix @phansys 👍

PR #2584, opened 2023-02-20
Replace note annotation

### LINKED ISSUE #2581: The @note annotation from MappedEventSubscriber causes AnnotationException
[This annotation](https://github.com/doctrine-extensions/DoctrineExtensions/blob/2b7d1abed85c6b8adb5e137a9d849a39c8c70330/src/Mapping/MappedEventSubscriber.php#L180) is causing the following issue:
> In AnnotationException.php line 40:
>                                                                                                                                                                                                         
>   [Doctrine\Common\Annotations\AnnotationException]                                                                                                                                                      
>   [Semantical Error] The annotation "@note" in method Gedmo\Mapping\MappedEventSubscriber::setAnnotationReader() was never imported. Did you maybe forget to add a "use" statement for this annotation?

Would it be possible to fix this? There are other examples of the same problem (for example FriendsOfSymfony/FOSRestBundle#535).

I would create a PR, but I'm not sure how you want to replace this comment. I personally use `NOTE` (which is highlighted by most editors) instead of `@note`.

mcorteel-harel:
Fixes #2581.

### DIFF
--- a/src/Mapping/MappedEventSubscriber.php
+++ b/src/Mapping/MappedEventSubscriber.php
@@ -177,7 +177,7 @@ abstract class MappedEventSubscriber implements EventSubscriber
      *
      * @return void
      *
-     * @note Providing any object is deprecated, as of 4.0 a `Doctrine\Common\Annotations\Reader` or `Gedmo\Mapping\Driver\AttributeReader` will be required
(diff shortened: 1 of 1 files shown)


phansys (approved):

franmomu (approved):

franmomu:
thanks @mcorteel-harel!

PR #2589, opened 2023-02-23
[Sluggable] Document the callables used for transliteration and urlization

mbabker:
As advertised, documents the callable signatures for the sluggable extension's transliterator and urlizer callables.

### DIFF
--- a/src/Sluggable/SluggableListener.php
+++ b/src/Sluggable/SluggableListener.php
@@ -75,6 +75,8 @@ class SluggableListener extends MappedEventSubscriber
      * Transliteration callback for slugs
      *
      * @var callable
+     *
+     * @phpstan-var callable(string $text, string $separator, object $object): string
      */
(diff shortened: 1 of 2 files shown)


phansys (approved):

franmomu (approved):

franmomu:
Thanks @mbabker! the phpstan errors are unrelated

franmomu:
Now that I take a second look at this, I think the `$object` parameter in the transliterator callable should not exist.

The idea of transliterator is: `Generates a slug of the text after transliterating the UTF-8 string to ASCII.`

The `Urlizer::transliterate` method [only accepts a `$text` and a `$separator`](https://github.com/Behat/Transliterator/blob/baac5873bac3749887d28ab68e2f74db3a4408af/src/Behat/Transliterator/Transliterator.php#L421-L438).

PR #2592, opened 2023-03-01
[SoftDeleteable] Fix soft-deleting an object which proxy was not yet initialized

spiroski:
I ran into issues where if an object proxy is not yet initialized ( due to being loaded as a relation, but not yet fetched from the database ) and that object is deleted, the soft-delete listener fails and InvalidArgumentException is thrown with a message: "Document is removed.".

This is because the proxy will be initialized when setValue is called in the listener, the object is initialized which causes it's state to be changed from removed to managed. This then causes problem when the object is persisted ( the document is not removed from the objects scheduled for deletion ). 

This PR fixes this edge-case by checking making sure the object is initialized in the preRemove event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Fixed
+- SoftDeleteable: Fix issue with soft-deleting objects which proxy was not yet initialized
 
 ## [3.11.1] - 2023-02-20
 ### Fixed
--- a/src/SoftDeleteable/SoftDeleteableListener.php
+++ b/src/SoftDeleteable/SoftDeleteableListener.php
@@ -11,6 +11,7 @@ namespace Gedmo\SoftDeleteable;
 
 use Doctrine\Common\EventArgs;
 use Doctrine\ODM\MongoDB\UnitOfWork as MongoDBUnitOfWork;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
 use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
 use Gedmo\Mapping\MappedEventSubscriber;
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,12 +12,15 @@ a release.
 ```markdown
 ## [3.6.1] - 2022-07-26
 ### Fixed
-- Sortable: Fix issue with add+delete position synchronization (#1932)
+- SoftDeleteable: Fix issue with soft-deleting objects which proxy was not yet initialized
```
phansys, line 15: Please, use the `[Unreleased]` section to describe the changes related to this PR.

spiroski: Sorry I mistakenly added the line to the wrong section. I've corrected it and updated the PR.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Fixed
+- SoftDeleteable: Fix issue with soft-deleting objects which proxy was not yet initialized
+
```
phansys, line 23:
```suggestion
```

spiroski: I removed the extra line.

phansys:
Thanks for this @spiroski, and congratulations for your first PR! :rocket: 

Could you please rebase your commits and add a test covering the fixed behavior?

phansys:
I was trying to reproduce the reported bug locally, but I failed to get the described situation. This is the test I used in https://github.com/doctrine-extensions/DoctrineExtensions/blob/36f2ee16e92df1848337e0a18a6b4a51238e4b4a/tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php#L42

```php
    public function testProxyObjectDeletion(): void
    {
        $newArticle = new Article();
        $title = 'Great Article!';
        $newArticle->setTitle($title);

        $this->em->persist($newArticle);

        $this->em->flush();
        $this->em->clear();

        $articleId = $newArticle->getId();

        static::assertNotNull($articleId);

        $article = $this->em->getReference(Article::class, $articleId);

        static::assertInstanceOf(Article::class, $article);
        static::assertInstanceOf(Proxy::class, $article);
        static::assertFalse($article->__isInitialized());

        $this->em->remove($article);

        $this->em->flush();
        $this->em->clear();

        $article = $this->em->find(Article::class, $articleId);

        static::assertNull($article);

        $this->em->getFilters()->disable(self::SOFT_DELETEABLE_FILTER_NAME);

        $article = $this->em->find(Article::class, $articleId);

        static::assertInstanceOf(Article::class, $article, 'Article should be fetched when filter is disabled.');
    }
```

I'm closing this one because the lack of response and the impossibility to reproduce the reported problem.
Please, feel free to reopen if you can provide a test for this.

Thank you!

PR #2604, opened 2023-03-13
Prioritize the values from the named arguments in the annotation classes

phansys:
Named arguments have precedence over the values passed in the `$data` array in annotation classes at `Gedmo\Mapping\Annotation\`  namespace.

### DIFF
--- a/src/Mapping/Annotation/ForwardCompatibilityTrait.php
+++ b/src/Mapping/Annotation/ForwardCompatibilityTrait.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Mapping\Annotation;
+
+/**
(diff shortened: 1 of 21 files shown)


franmomu (approved):
did you have any issue with this?

phansys:
>did you have any issue with this?

No, I didn't. I'm proposing this change because I think we should prioritize the way we are recommending now over the way we are considering as deprecated.

phansys:
@franmomu, please let me know if you have any additional concern on this. Even if you think the priority must keep focused on the deprecated options.

franmomu:
> @franmomu, please let me know if you have any additional concern on this. 

nope, I see this fine, I was just curious 😬

src/Mapping/Annotation/ForwardCompatibilityTrait.php
```diff
@@ -0,0 +1,50 @@
...
+        $reflectionMethod = new \ReflectionMethod($this, '__construct');
+
+        foreach ($reflectionMethod->getParameters() as $param) {
+            if ($param->getName() !== $attributeName) {
+                continue;
+            }
+
+            $legacyValue = $data[$attributeName] ?? null;
+
+            if (null === $legacyValue || $param->getDefaultValue() !== $value) {
```
phansys, line 41: We should check if the argument was provided, regardless its value is the same as the default one.

phansys: The behavior was updated in order to address this comment.

PR #2620, opened 2023-05-24
[Sluggable] Use TranslationWalker for Translatable objects when looking for similar slugs

### LINKED ISSUE #100: [Sluggable] uniqueness breaks when combined with translations

### LINKED ISSUE #2530: [Sluggable] getSimilarSlugs with locale context
When using sluggable with translations the **getSimilarSlugs** method must use the entity object locale: 

example in **Gedmo\Sluggable\Mapping\Event\Adapter\ORM**

```
...
$q = $qb->getQuery();

$q->setHint(
  Query::HINT_CUSTOM_OUTPUT_WALKER,
  'Gedmo\\Translatable\\Query\\TreeWalker\\TranslationWalker'
);

$q->setHint(
  TranslatableListener::HINT_TRANSLATABLE_LOCALE,
  $locale // got with object parameter
);

$q->setHydrationMode(Query::HYDRATE_ARRAY);
        
return $q->execute();
...
```

ambroisemaupate:
Force translation walker to look for slug translations to avoid duplicated slugs.

---

Fixes #100, fixes #2530


ambroisemaupate:
Changelog and Unit test have been added.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ a release.
 
 ### Fixed
 - Tree: Creation of dynamic `Node::$sibling` property, which is deprecated as of PHP >= 8.2
+- Sluggable: Use `TranslationWalker` hint only for Translatable objects when looking for similar slugs (`getSimilarSlugs` method) (#100, #2530)
```
phansys, line 33: IMO, this description conveys the idea that previous to this change, any object (even when they were not implementing `Translatable`) were being translated.
IIUC, this is not the case. Could you please confirm?

ambroisemaupate: Yes this is not the case. I meant that now we are looking for similar slugs using `TranslationWalker` but only when entities implements `Translatable` interface.

phansys (commented):
Even if this may be considered as a fix, I guess the change will break the behavior for users relying on the current behavior.

Is this behavior documented? If not, I think this PR must also provide the required documentation.

How we can build a strong and indisputable argument to support this change as a **fix** and not as a **new feature**?

I'd like to hear more opinions here.
`/cc` @mbabker, @franmomu.

tharsanvn:
Hello guys :) 

Any news about this PR ?

mbabker:
> Even if this may be considered as a fix, I guess the change will break the behavior for users relying on the current behavior.
> 
> Is this behavior documented? If not, I think this PR must also provide the required documentation.
> 
> How we can build a strong and indisputable argument to support this change as a **fix** and not as a **new feature**?

I'd put it behind a config from the metadata instead of making a hard behavioral change.  For the folks who want globally unique slugs, even across languages, they'd turn on this behavior with the config in the metadata.

If a platform is OK with duplicated slugs across languages, then it should also have something in place to deal with that potential duplication (i.e. if your route definition is `/{locale}/{category}/{post}`, then `/en/tips/translation` and `/fr/tips/translation` should already provide fully unique paths that allow querying the right info without requiring globally unique slugs for the category or post.).

ambroisemaupate:
How would you implement this as configurable? @mbakker
Adding `uniqueOverTranlations` option in `Gedmo\Slug` attribute ?

ambroisemaupate:
@phansys @mbabker 

I've added a new option `uniqueOverTranslations` (default: false) to trigger `TranslationWalker::class`  ORM hint in `getSimilarSlugs` method.

src/Mapping/Annotation/Slug.php
```diff
@@ -38,6 +38,7 @@ final class Slug implements GedmoAnnotation
     public bool $updatable = true;
     public string $style = 'default'; // or "camel"
     public bool $unique = true;
+    public bool $uniqueOverTranslations = false;
```
mbabker, line 41: The new argument needs to be added at the end of the constructor, throwing it in the middle of the list is a B/C break.

tests/Gedmo/Sluggable/Fixture/Issue100/Article.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
+    #[Gedmo\Translatable]
+    private $title;
+
+    /**
+     * @var string|null
+     *
+     * @Gedmo\Translatable
+     * @Gedmo\Slug(separator="-", updatable=true, fields={"title"}, unique=true)
```
mbabker, line 52:
```suggestion
     * @Gedmo\Slug(separator="-", updatable=true, fields={"title"}, unique=true, uniqueOverTranslations=true)
```

Otherwise the PHP 7.4 build using annotations will have a different behavior from the PHP 8 builds with attributes

CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ a release.
 
 ### Fixed
 - Tree: Creation of dynamic `Node::$sibling` property, which is deprecated as of PHP >= 8.2
+- Sluggable: Use `TranslationWalker` hint when looking for similar slugs (`getSimilarSlugs` method) for entities which implement `Translatable` interface (#100, #2530)
```
mbabker, line 76: The changelog entry can be moved up to the unreleased section near the top, this definitely isn't part of the already tagged 3.12 release 😅

mbabker (approved):

franmomu (approved):

franmomu:
thanks @ambroisemaupate for keeping working on this! and @mbabker

PR #2621, opened 2023-06-05
chore: fix deprecation message for node12 for github action "stale"

Chris53897:
It looks like this is changes from single to multiple values allowed
https://github.com/actions/stale#exempt-issue-labels

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

I am not sure how to test it.


phansys:
See actions/stale#19.

phansys:
Thanks @Chris53897!

Issue #2624, opened 2023-06-07
docker-compose up does not run local on Mac M1


Chris53897:
I try to run the tests local, and could not get the dockerfile running.

### Environment

Mac M1

## Steps to reproduce

`docker-compose up`

## Expected results

Docker build should be correct.

## Actual results

=> #0 3.021 E: Version '1:2.30.2-1' for 'git' was not found

➜  DoctrineExtensions-1 git:(feature/run-ci-on-every-branch) ✗ docker-compose up
[+] Building 4.5s (10/10) FINISHED                                                                                                                                                                      
 => [php internal] load build definition from Dockerfile                                                                                                                                           0.0s
 => => transferring dockerfile: 733B                                                                                                                                                               0.0s
 => [php internal] load .dockerignore                                                                                                                                                              0.0s
 => => transferring context: 2B                                                                                                                                                                    0.0s
 => [php] resolve image config for docker.io/docker/dockerfile:1                                                                                                                                   0.6s
 => CACHED [php] docker-image://docker.io/docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14                                                              0.0s
 => [php internal] load metadata for docker.io/library/php:8.2-cli                                                                                                                                 0.7s
 => [php internal] load metadata for docker.io/library/composer:2                                                                                                                                  0.6s
 => [php composer 1/1] FROM docker.io/library/composer:2@sha256:40fea1182aa0fe24f389e20c6604e881a775e1c6b5457c2d74cdda8f5206808d                                                                   0.0s
 => [php php 1/3] FROM docker.io/library/php:8.2-cli@sha256:b0eca9a9cb893d096dc0fc4a80a44697cabe6e1ed965cbccf5fd6046b4eda341                                                                       0.0s
 => CACHED [php php 2/3] COPY --from=composer /usr/bin/composer /usr/bin/composer                                                                                                                  0.0s
 => ERROR [php php 3/3] RUN apt-get update     && apt-get install -y --no-install-recommends     git=1:2.30.2-1     libzip-dev=1.7.3-1     unzip=6.0-26+deb11u1     zip=3.0-12     zlib1g-dev=1:1  3.0s
------
[...]

PR #2625, opened 2023-06-07
chore: run tests on all branches for push

Chris53897:
I can not test local. see https://github.com/doctrine-extensions/DoctrineExtensions/issues/2624

I now have to send PRs more or less blind and wait for a workflow approval to see the CI Results ;(
Or i have to use the branch-name `main` to send in PRs. One at a time

Can we have a solution to make it easier to contribute?
I suggest to run checks on all code pushes.

Of course i am open for other solutions

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -2,8 +2,6 @@ name: "Coding Standards"
 
 on:
   push:
-    branches:
-      - main
   pull_request:
 
 jobs:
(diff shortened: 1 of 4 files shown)


phansys (commented):
I understand the motivation, but this change will not only affect your fork, but the upstream repository as well.
Every time I need the CI pipeline to run on my branch, I add a commit that disables the filter on my branch.

Chris53897:
Is this mabe a solution? All branches with the prefix 'feature/' will be run by the ci.
And only on .github/workflows/continuous-integration.yml
The other 3 files could be reverted.

```
on:
  push:
    branches:
      - main
      - 'feature/**'
    pull_request:

```

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-filters

PR #2626, opened 2023-06-07
feat: allow friendsofphp/php-cs-fixer >= 3.10

Chris53897:
bump "rector/rector" to 0.17

is the upper constraint for friendsofphp/php-cs-fixer still in need?

On local development, this blocks doctrine/annotations v2.
The CI removes it before running the tests.
https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/.github/workflows/continuous-integration.yml#L65

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -98,6 +98,7 @@
         }
     },
     "scripts": {
-        "fix-cs": "php-cs-fixer fix --config=.php-cs-fixer.dist.php"
+        "fix-cs": "php-cs-fixer fix",
+        "fix-cs-test": "php-cs-fixer fix --ansi --verbose --diff --dry-run"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.3",
         "doctrine/orm": "^2.14.0",
-        "friendsofphp/php-cs-fixer": "^3.4.0",
+        "friendsofphp/php-cs-fixer": "^3.14.0",
         "nesbot/carbon": "^2.55",
         "phpstan/phpstan": "^1.10.2",
         "phpstan/phpstan-doctrine": "^1.0",
@@ -98,7 +98,6 @@
         }
     },
     "scripts": {
-        "fix-cs": "php-cs-fixer fix",
-        "fix-cs-test": "php-cs-fixer fix --ansi --verbose --diff --dry-run"
+        "fix-cs": "php-cs-fixer fix --config=.php-cs-fixer.dist.php"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.3",
         "doctrine/orm": "^2.14.0",
-        "friendsofphp/php-cs-fixer": "^3.4.0 <3.10",
+        "friendsofphp/php-cs-fixer": "^3.4.0",
         "nesbot/carbon": "^2.55",
         "phpstan/phpstan": "^1.10.2",
         "phpstan/phpstan-doctrine": "^1.0",
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -52,6 +52,7 @@ return (new PhpCsFixer\Config())
         'no_homoglyph_names' => true,
         'no_null_property_initialization' => true,
         'no_superfluous_elseif' => true,
+        'no_superfluous_phpdoc_tags' => false,
         'no_unset_on_property' => true,
         'no_useless_else' => true,
         'nullable_type_declaration_for_default_null_value' => ['use_nullable_type_declaration' => true],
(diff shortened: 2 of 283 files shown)


phansys (commented):
I'm fine allowing higher versions of `friendsofphp/php-cs-fixer` as long the considerations mentioned at #2507 regarding the `phpdoc_order` rule are covered.

Chris53897:
Ok i see a lot of default rules changed. I had the same problem at another repo.

Can you please decide how you want this configs to be set.
`phpdoc_separation` 
`global_namespace_import` @var \DateTime|null vs @var DateTime|null

I have added `no_superfluous_phpdoc_tags => false,` because it would otherwise remove useful phpdocs (in my point of view).
Additional i have added a composer-script to check it locally (dry-run same as in ci) to make it easier to check it local.

I run it locally on php 8.2. Looks like the default is correct followed
'phpdoc_order' => [
                'order' => [
                    'param',
                    'return',
                    'throws',
                ],
            ],

franmomu:
> Can you please decide how you want this configs to be set. `phpdoc_separation` `global_namespace_import` @var \DateTime|null vs @var DateTime|null

Now I see this, I think we can use for `phpdoc_separation`:

```
'phpdoc_separation' => ['groups' => [
            ['Gedmo\\*'],
            ['ODM\\*'],
            ['ORM\\*'],
        ]],
```

`global_namespace_import` I would leave it like `@var \DateTime|null` which is what we have now.

> I have added `no_superfluous_phpdoc_tags => false,` because it would otherwise remove useful phpdocs (in my point of view). Additional i have added a composer-script to check it locally (dry-run same as in ci) to make it easier to check it local.
> 
> I run it locally on php 8.2. Looks like the default is correct followed
> 
> ```
> 'phpdoc_order' => [
>                 'order' => [
>                     'param',
>                     'return',
>                     'throws',
>                 ],
>             ],
> ```

About the order, we have now ['param', 'throws', 'return'] which was the default order, I think @phansys meant to keep it

phansys (approved):

composer.json
```diff
@@ -98,6 +98,7 @@
         }
     },
     "scripts": {
-        "fix-cs": "php-cs-fixer fix --config=.php-cs-fixer.dist.php"
+        "fix-cs": "php-cs-fixer fix",
+        "fix-cs-test": "php-cs-fixer fix --ansi --verbose --diff --dry-run"
```
phansys, line 102:
```suggestion
        "lint-cs": "php-cs-fixer fix --ansi --verbose --diff --dry-run"
```
BTW, personally, I'd try to avoid adding more of these shortcuts as they don't bring any added value IMO.

Chris53897: The value for me is, that i do not have to remember all the options to run it. I can just run it with a single click. Otherwise this could be added to the readme.

Chris53897:
Anything todo for me here?

composer.json
phansys: IMO, this argument is valid for your private projects. In OS projects, the dependencies must focus on their own responsibilities, relying on other projects where they need to attend different requirements.
What I mean is, the CLI options for php-cs-fixer command are exhaustively documented in [their own docs]( https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/usage.rst#usage).
If we add a new shortcut here for each combination of possible options in the required 3rd party tools, we will end with a big list.

phansys:
> Anything todo for me here?

No @Chris53897, thank you.
I'll be waiting for the @franmomu's review.

composer.json
franmomu: > If we add a new shortcut here for each combination of possible options in the required 3rd party tools, we will end with a big list.

Yeah, that's a valid argument, I personally don't use these shortcuts.

I see fine to add something in the Readme file for development though.

composer.json
```diff
@@ -57,13 +57,13 @@
         "doctrine/doctrine-bundle": "^2.3",
         "doctrine/mongodb-odm": "^2.3",
         "doctrine/orm": "^2.10.2",
-        "friendsofphp/php-cs-fixer": "^3.4.0 <3.10",
+        "friendsofphp/php-cs-fixer": "^3.4.0",
```
phansys, line 60: The `groups` option for the `phpdoc_separation` rule was introduced in [version 3.14.0](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/releases/tag/v3.14.0).

See PHP-CS-Fixer/PHP-CS-Fixer#6668.

phansys (changes requested):

composer.json
Chris53897: And https://packagist.org/packages/friendsofphp/php-cs-fixer#v3.14.0 is `php: ^7.4 || ^8.0`
In that case, https://github.com/doctrine-extensions/DoctrineExtensions/pull/2636#issuecomment-1685240982 could be merged upfront?

phansys: Yes, we should bump the PHP version before.

phansys (approved):

franmomu:
Thanks @Chris53897! and @phansys for reviewing!

Issue #2635, opened 2023-06-19
RFC: Drop support of php < 7.4?


Chris53897:
I like to suggest to drop support of php < 7.4
The [Usage](https://packagist.org/packages/gedmo/doctrine-extensions/php-stats#3) is under 3%

Im in favour to drop support for 7.4 as well and only use attributes.
But i guess this is to fast for most users.

franmomu:
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2636 was merged

PR #2636, opened 2023-06-20
Feature/drop support for php 7.3

Chris53897:
https://github.com/doctrine-extensions/DoctrineExtensions/issues/2635

- drop support for php < 7.3
- add property types for final classes
- Remove 2 function related to PHPunit < 9.1
- use null coalescing operator 2x

Needs a check for BC breaks.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,8 +23,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.2"
-          - "7.3"
           - "7.4"
           - "8.0"
           - "8.1"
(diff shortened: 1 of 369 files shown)


stof (approved):

phansys (changes requested):
This change should be supported by Rector:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/070f350987f76f12a0f63b82fdfe92b48123e279/rector.php#L24

Chris53897:
Using rector as you suggested will introduce a lot of BC Breaks.

@TomasVotruba Is there a safe way to autoupdate with rector for libraries without BC Breaks?

I would suggest to wait until the problems with doctrine/orm 2.16 are fixed.
After that, merge this PR as it is, and improve after that in follow up PRs.

phansys (approved):
Maybe we could also implement the rule set [`@PHP74Migration`](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/ruleSets/PHP74Migration.rst#rule-set-php74migration) from PHP-CS-Fixer.

franmomu:
Thanks @Chris53897!

PR #2637, opened 2023-06-27
Get real table name on translation insert

asaintvanne:
Get table name with schema as it's done in Doctrine on translation part.


phansys:
Thanks @asaintvanne!
Is this a fix? Please, add a test covering this change and a note in the corresponding section of the [`CHANGELOG.md`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/0b1de148959a31f2df9b8a4295d190c0299e6fc1/CHANGELOG.md?plain=1#L20) file.

Issue #2638, opened 2023-06-27
Doctrine deprecations


DubbleClick:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Mapping/Event/Adapter/ORM.php#L98

This calls LifeCycleEventArgs->getEntityManager() which has been deprecated for getObjectManager() instead. Likewise, calls to getEntity() have been deprecated for getObject() as well.

phansys:
Thank you @DubbleClick.
Do you plan to create a PR for these deprecation?

franmomu:
Fixed in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2639

PR #2634, opened 2023-06-19
feat: fix deprecation of creating EntityManager

Chris53897:
Needs CI tests to see if --prefer-lowest is ok for this PR

should be available with 2.13.0
https://github.com/doctrine/orm/commit/518d7f2ef1a1ffcd7e33c1db7663686db2d78bb4
https://github.com/doctrine/orm/releases/tag/2.13.0

But the --prefer-lowest run fails (maybe more).
I guess this PR should wait after the minimum versions are bumped.

PR #2639, opened 2023-06-28
fix EventArgs::getEntityManger deprecation

DubbleClick:
tackles #2638 and #2640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,11 @@ a release.
 
 ## [Unreleased]
 
-## [3.13.0]
+### Deprecated
+- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager()` and `getObject()` on EventArgs that do not implement `getObjectManager()` and `getObject()` (such as old EventArgs implementing `getEntityManager()` and `getEntity()`) 
+- Calling `Gedmo\Uploadable\Event\UploadableBaseEventArgs::getEntityManager()` and `getEntity()`. Call `getObjectManager()` and `getObject()` instead. 
+
+## [3.13.0] - 2023-09-06
 ### Fixed
 - References: fixed condition in `XML` Driver that did not allow to retrieve from the entity definition the `mappedBy` and `inversedBy` fields.
 - Fix bug collecting metadata for inherited mapped classes
--- a/src/Mapping/Event/Adapter/ORM.php
+++ b/src/Mapping/Event/Adapter/ORM.php
@@ -95,12 +95,15 @@ class ORM implements AdapterInterface
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
         }
 
+        // todo: for the next major release, uncomment the next line:
+        // return $this->args->getObjectManager();
+        // and remove anything past this
         if (\method_exists($this->args, 'getObjectManager')) {
             return $this->args->getObjectManager();
         }
 
         @trigger_error(sprintf(
-            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
+            'Calling "%s()" on event args that do not implement "getObjectManager()" is deprecated since gedmo/doctrine-extensions 3.x'
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
@@ -116,12 +119,15 @@ class ORM implements AdapterInterface
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
         }
 
+        // todo: for the next major release, uncomment the next line:
+        // return $this->args->getObject();
+        // and remove anything past this
         if (\method_exists($this->args, 'getObject')) {
             return $this->args->getObject();
         }
 
         @trigger_error(sprintf(
-            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
+            'Calling "%s()" on event args that do not imeplement "getObject()" is deprecated since gedmo/doctrine-extensions 3.x'
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
--- a/src/Mapping/Event/Adapter/ORM.php
+++ b/src/Mapping/Event/Adapter/ORM.php
@@ -104,7 +104,7 @@ class ORM implements AdapterInterface
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
-            \TypeError::class
+            \Error::class
         ), E_USER_DEPRECATED);
 
         return $this->args->getEntityManager();
@@ -125,7 +125,7 @@ class ORM implements AdapterInterface
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
-            \TypeError::class
+            \Error::class
         ), E_USER_DEPRECATED);
 
         return $this->args->getEntity();
--- a/src/Mapping/Event/Adapter/ORM.php
+++ b/src/Mapping/Event/Adapter/ORM.php
@@ -99,6 +99,14 @@ class ORM implements AdapterInterface
             return $this->args->getObjectManager();
         }
 
+        @trigger_error(sprintf(
+            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
+            .' and will throw a "%s" error in version 4.0.',
+            __METHOD__,
+            get_class($this->args),
+            \TypeError::class
+        ), E_USER_DEPRECATED);
+
         return $this->args->getEntityManager();
     }
 
@@ -111,6 +119,14 @@ class ORM implements AdapterInterface
         if (\method_exists($this->args, 'getObject')) {
             return $this->args->getObject();
         }
+
+        @trigger_error(sprintf(
+            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
+            .' and will throw a "%s" error in version 4.0.',
+            __METHOD__,
+            get_class($this->args),
+            \TypeError::class
+        ), E_USER_DEPRECATED);
 
         return $this->args->getEntity();
     }
--- a/src/Mapping/Event/Adapter/ORM.php
+++ b/src/Mapping/Event/Adapter/ORM.php
@@ -103,7 +103,7 @@ class ORM implements AdapterInterface
         }
 
         @trigger_error(sprintf(
-            'Calling "%s()" on event args that do not implement "getObjectManager()" is deprecated since gedmo/doctrine-extensions 3.x'
+            'Calling "%s()" on event args of class "%s" that does not implement "getObjectManager()" is deprecated since gedmo/doctrine-extensions 3.x'
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
@@ -127,7 +127,7 @@ class ORM implements AdapterInterface
         }
 
         @trigger_error(sprintf(
-            'Calling "%s()" on event args that do not imeplement "getObject()" is deprecated since gedmo/doctrine-extensions 3.x'
+            'Calling "%s()" on event args of class "%s" that does not imeplement "getObject()" is deprecated since gedmo/doctrine-extensions 3.x'
             .' and will throw a "%s" error in version 4.0.',
             __METHOD__,
             get_class($this->args),
--- a/src/Mapping/Event/Adapter/ORM.php
+++ b/src/Mapping/Event/Adapter/ORM.php
@@ -95,6 +95,10 @@ class ORM implements AdapterInterface
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
         }
 
+        if (\method_exists($this->args, 'getObjectManager')) {
+            return $this->args->getObjectManager();
+        }
+
         return $this->args->getEntityManager();
     }
 
@@ -102,6 +106,10 @@ class ORM implements AdapterInterface
     {
         if (null === $this->args) {
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
+        }
+
+        if (\method_exists($this->args, 'getObject')) {
+            return $this->args->getObject();
         }
 
         return $this->args->getEntity();
--- a/src/Uploadable/Event/UploadableBaseEventArgs.php
+++ b/src/Uploadable/Event/UploadableBaseEventArgs.php
@@ -11,6 +11,7 @@ namespace Gedmo\Uploadable\Event;
 
 use Doctrine\Common\EventArgs;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\ObjectManager;
 use Gedmo\Uploadable\FileInfo\FileInfoInterface;
 use Gedmo\Uploadable\UploadableListener;
 
@@ -101,6 +102,21 @@ abstract class UploadableBaseEventArgs extends EventArgs
      */
     public function getEntityManager()
     {
+        @trigger_error(sprintf(
+            '"%s()" is deprecated since gedmo/doctrine-extensions 3.x and will be removed in version 4.0.',
+            __METHOD__
+        ), E_USER_DEPRECATED);
+
+        return $this->em;
+    }
+
+    /**
+     * Retrieve associated EntityManager
+     *
+     * @return ObjectManager
+     */
+    public function getObjectManager()
+    {
         return $this->em;
     }
 
@@ -110,6 +126,21 @@ abstract class UploadableBaseEventArgs extends EventArgs
      * @return object
      */
     public function getEntity()
+    {
+        @trigger_error(sprintf(
+            '"%s()" is deprecated since gedmo/doctrine-extensions 3.x and will be removed in version 4.0.',
+            __METHOD__
+        ), E_USER_DEPRECATED);
+
+        return $this->entity;
+    }
+
+    /**
+     * Retrieve associated Object
+     *
+     * @return object
+     */
+    public function getObject()
     {
         return $this->entity;
     }
--- a/tests/Gedmo/Mapping/MappingEventAdapterTest.php
+++ b/tests/Gedmo/Mapping/MappingEventAdapterTest.php
@@ -53,10 +53,10 @@ final class MappingEventAdapterTest extends TestCase
             ->disableOriginalConstructor()
             ->getMock();
         $eventArgsMock->expects(static::once())
-            ->method('getEntityManager');
+            ->method('getObjectManager');
 
         $eventArgsMock->expects(static::once())
-            ->method('getEntity')
+            ->method('getObject')
             ->willReturn(new \stdClass());
 
         $eventAdapter = new EventAdapterORM();


doc/symfony6.md
```diff
@@ -0,0 +1,445 @@
...
+        calls:
+            - [ setAnnotationReader, [ "@annotation_reader" ] ]
+
+    # Symfony 6.3 change          
+    gedmo.listener.timestampable:
+        class: Gedmo\Timestampable\TimestampableListener
+        tags:
+            - { name: doctrine.event_listener, event: 'prePersist', connection: default }
+            - { name: doctrine.event_listener, event: 'onFlush', connection: default }
+            - { name: doctrine.event_listener, event: 'loadClassMetadata', connection: default }
```
DubbleClick, line 170: This isn't optimal, the correct way to go about this would be to stop inheriting from MappedEventSubscriber in https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/AbstractTrackingListener.php and changing the code entirely, but that's unfortunately a little past my familiarity with the project.

src/Uploadable/Event/UploadableBaseEventArgs.php
```diff
@@ -104,6 +105,16 @@ public function getEntityManager()
         return $this->em;
```
DubbleClick, line 105: I decided against firing a deprecation here as an EntityManagerInterface is indeed injected

DubbleClick:
@franmomu please trigger the workflows again, getEntity() should no longer be called by us.

DubbleClick:
It is however noteworthy that there is use of plenty other deprecated functionality throughout the project (EntityManager::create) for example. Not all of it is trivial to work around and that unfortunately exceeds the time I can currently invest.

doc/symfony6.md
```diff
@@ -0,0 +1,445 @@
+# Install Gedmo Doctrine extensions in Symfony 6
```
phansys, line 1: How is the integration with Symfony 6 related to the deprecation fix for `EventArgs::getEntityManger()`?

DubbleClick: It's not related to EntityManager, it's related to the deprecation of Event Subscribers. This comes from the Symfony Bridge: https://github.com/symfony/symfony/issues/49586

franmomu:
I think we are doing two things here, apart from avoiding calls to `getEntityManager()` from doctrine, we are also deprecating some of our methods, I would leave deprecating these methods for another PR (which would require a minor release)

DubbleClick:
I'll be happy to remove the deprecation of the methods and open another PR for that. Should I?

franmomu:
I was thinking about adding some fixes like this one and https://github.com/doctrine-extensions/DoctrineExtensions/pull/2651 at least and make a patch release and then make a minor one adding deprecations + removing support for php < 7.4 and probably symfony < 5.4 cc @phansys

DubbleClick:
and #2634 for the holy trinity of pull requests to get rid of all deprecations for now

franmomu:
@DubbleClick if you could please split this PR into two, one that fixes triggering deprecations (could be this one) and another one deprecating methods. 

The `symfony6.md` file won't be necessary, we'll deal with that in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2667 (thanks for creating the file).

DubbleClick:
yep will do it tomorrow

DubbleClick:
@franmomu @phansys sorry, took a little longer

I removed all the deprecations and only kept in the added methods (and a safety check for $platform)

DubbleClick:
rebased it all to only show up in one commit and also created one for EntityManager::create deprecations at https://github.com/doctrine-extensions/DoctrineExtensions/pull/2669

DubbleClick:
I'm not entirely sure how to get rid of all the warnings. Either there will be a warning that ObjectManager is returned rather than EntityManagerInterface, or there will be warnings that ObjectManager doesn't have methods like createQueryBuilder.

franmomu (commented):
Thanks for keep working on this, sorry about reverting changes, but it's better if in the same commit there are only changes related to the current fix.

Sorry that I wasn't clear 😞, what I meant in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2639#issuecomment-1677277868 was to only add the necessary code to not trigger deprecations.

I think adding `ORM::setObjectManager`, `UploadableBaseEventArgs::getObjectManager()` and `UploadableBaseEventArgs::getObject()` makes sense if we deprecate the old ones, so I would leave that to the other PR.

src/Translatable/Query/TreeWalker/TranslationWalker.php
```diff
@@ -358,7 +356,7 @@ private function prepareTranslatedComponents(): void
                 // Treat translation as original field type
                 $fieldMapping = $meta->getFieldMapping($field);
                 if ((($this->platform instanceof MySQLPlatform) &&
-                    in_array($fieldMapping['type'], ['decimal'], true)) ||
+                    'decimal' === $fieldMapping['type']) ||
```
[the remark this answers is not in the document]
DubbleClick: Time to run
https://cs.symfony.com/doc/rules/return_notation/return_assignment.html
and
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#singleinarraytocomparerector

then, but I'll leave that to you :)

src/SoftDeleteable/Filter/SoftDeleteableFilter.php
```diff
@@ -72,7 +72,7 @@ public function addFilterConstraint(ClassMetadata $targetEntity, $targetTableAli
         $column = $quoteStrategy->getColumnName($config['fieldName'], $targetEntity, $platform);
 
         $addCondSql = $targetTableAlias.'.'.$column.' IS NULL';
-        if (isset($config['timeAware']) && $config['timeAware']) {
+        if (isset($config['timeAware']) && $config['timeAware'] && $platform) {
```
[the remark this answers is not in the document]
DubbleClick: I've reverted it, but it certainly can be:
image

DubbleClick:
> Sorry that I wasn't clear 😞, what I meant in [#2639 (comment)](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2639#issuecomment-1677277868) was to only add the necessary code to not trigger deprecations.

No worries, I've removed everything not strictly necessary in order to not trigger deprecations now.

> I think adding `ORM::setObjectManager`, `UploadableBaseEventArgs::getObjectManager()` and `UploadableBaseEventArgs::getObject()` makes sense if we deprecate the old ones, so I would leave that to the other PR.

It'll be inevitable at some point, but it doesn't have to happen right now.

tests/Gedmo/Mapping/MappingEventAdapterTest.php
```diff
@@ -12,7 +12,7 @@
 namespace Gedmo\Tests\Mapping;
 
 use Doctrine\ORM\EntityManager;
-use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
```
franmomu, line 15: There is this code 😞 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/351caed07e47552ab0df63cafb4fdcccc009eb59/src/Mapping/MappedEventSubscriber.php#L252-L277

that do some funny things to determine which event adapter has to be instantiated and it's based on the event class namespace that should include `ORM` or `ODM`.

It would be awesome to be able to get rid of this code, but since it doesn't seem easy, maybe we should keep using the deprecated `Doctrine\Persistence\Event\LifecycleEventArgs;`, once https://github.com/doctrine-extensions/DoctrineExtensions/pull/2669 is merged we can use one of the new dedicated events.

src/SoftDeleteable/Filter/SoftDeleteableFilter.php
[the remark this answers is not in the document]
franmomu: That's weird, the implementation is:

https://github.com/doctrine/dbal/blob/63646ffd71d1676d2f747f871be31b7e921c7864/src/Connection.php#L306-L321

and also in the information says it returns a `AbstractPlatform`, don't know why it says `AbstractPlatform|null` on top 🤷‍♂️

tests/Gedmo/Mapping/MappingEventAdapterTest.php
DubbleClick: GitHub doesn't have the correct emoji to express my feelings about that code. Sometimes I feel like code reflection goes just a little too far in some languages, hahaha.

Anyway, I've reverted back to Doctrine\ORM\Event\LifecycleEventArgs.

franmomu (approved):
I see this fine after adding the changelog, I'll approve since I'm not sure if this week I'll be available

src/Mapping/Event/Adapter/ORM.php
```diff
@@ -95,6 +95,10 @@ public function getObjectManager()
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
         }
 
+        if (\method_exists($this->args, 'getObjectManager')) {
```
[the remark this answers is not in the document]
phansys: I'm asking about the implementation provided by this PR, in order to determine if a `@todo` comment with the instructions for the next major release is required.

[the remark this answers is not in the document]
DubbleClick: once #2669 is merged this check becomes obsolete

doc/symfony6.md
stof: The simpler solution here might be to make the documentation link to the bundle at https://packagist.org/packages/stof/doctrine-extensions-bundle, which handles registering listeners instead of subscribers, with CI checks warning us if the events used by a listener are updated.

src/Mapping/Event/Adapter/ORM.php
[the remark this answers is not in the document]
stof: Given that all versions before 2.14.x are EOL (even 2.14.x is), I would suggest simplifying the implementation by bumping the lower bound of the constraint.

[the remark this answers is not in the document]
DubbleClick: +1, will change this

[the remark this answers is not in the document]
DubbleClick: @phansys did a version bump to 2.14 and removed the \method_exists check in another commit. If you wish to keep the version at 2.10 for now, just merge the first commit.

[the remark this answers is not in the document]
franmomu: Sorry for taking so long, I think the `\method_exists` will still be necessary, we are working with `EventArgs`: https://github.com/doctrine/event-manager/blob/12f08bd277ce61d7683253264c1d58b797841604/src/EventArgs.php which doesn't have `getObject()` nor `getObjectManager()`.

Custom implementations like `UploadableBaseEventArgs` doesn't have these methods (we should add these methods and deprecate `getEntity` and `getEntityManager` as discussed before).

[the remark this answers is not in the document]
DubbleClick: Do you want me to undo the last commit, or add these necessary changes back in?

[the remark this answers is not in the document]
franmomu: IMO the last commit should be reverted, I've created https://github.com/doctrine-extensions/DoctrineExtensions/pull/2697 for the conflict `<2.14`.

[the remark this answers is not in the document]
DubbleClick: Will do tomorrow.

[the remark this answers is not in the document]
phansys: I think we need a `@todo` comment explaining what to do with these checks when dropping the support for the versions that don't are declaring these methods.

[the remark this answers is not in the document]
stof: to me, the way to handle custom implementations provided inside the doctrine-extensions package is to add the methods in those classes (the old methods can then be deprecated).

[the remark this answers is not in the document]
DubbleClick: That's what I did at first, haha.

src/Mapping/Event/Adapter/ORM.php
```diff
@@ -95,6 +95,11 @@ public function getObjectManager()
             throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
         }
 
+        // @todo create getObjectManager in UploadableBaseEventArgs.php and remove this check
```
phansys, line 98: Could you please use the same approach than #2547, creating and commenting the method, and adding the `@method` annotation at `UploadableBaseEventArgs`?
https://github.com/doctrine-extensions/DoctrineExtensions/blob/c8a995317bb0a3de88f64f912c8d2ca7b76095e4/src/Tree/Node.php#L12-L54

DubbleClick: @phansys sure I could do that, but at that point, what's the point in even keeping the check in place?
Current master already restricts doctrine/orm to 2.14+, when the method is implemented in UploadableBaseEventArgs, it will always exist.

DubbleClick: I think the better way to go about it is to create the method, change the pipeline to ignore getObjectManager calls on EventArgs (it currently ignores getEventManager calls) and get rid of the check. It doesn't actually hurt backward compatibility in any way, or am I missing something?

phansys: It sounds logic at first sight.
Let me check out the PR locally to have a better understanding of the scenario here, as I'm not following this topic.

phansys: From what I see, the case seems to be even worse than I thought, as the property used here (`ORM::$args`) is using a type (`Doctrine\Common\EventArgs`) that does not guarantee the `getEntityManager()` or the `getEntityManager()` methods:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/c8a995317bb0a3de88f64f912c8d2ca7b76095e4/src/Mapping/Event/Adapter/ORM.php#L52

AFAIK, the only usable class declaring these methods is [`Doctrine\Persistence\Event\LifecycleEventArgs`](
https://github.com/doctrine/persistence/blob/8a710c21dfe6e88df21c371012be1bf6f2337525/src/Persistence/Event/LifecycleEventArgs.php#L16), as the [implementation for the ORM](https://github.com/doctrine/orm/blob/9f555ea8fb4f7ae9a362285dc71e85776a1c5cc3/lib/Doctrine/ORM/Event/LifecycleEventArgs.php#L19) is deprecated.

That said, I know we surely couldn't use `LifecycleEventArgs` as type declaration here, as there are more implementations of `EventArgs` passed here (`Doctrine\ORM\Event\LoadClassMetadataEventArgs` by instance).

With that on mind, I guess we could use some type check and a deprecation path on these getters to prevent calling `$this->args->getObjectManager()` and `$this->args->getObject()` on objects with other type than `Doctrine\Persistence\Event\LifecycleEventArgs`.

I mean something like this:
```php
public function getObject(): object
{
    if (null === $this->args) {
        throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
    }

    if (!$this->args instanceof \Doctrine\Persistence\Event\LifecycleEventArgs) {
        @trigger_error(sprintf(
            'Calling "%s()" on an event that is not of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
            .' and will throw a "%s" error in version 4.0.',
            __METHOD__,
            LifecycleEventArgs::class,
            \TypeError::class
        ), E_USER_DEPRECATED);
    }

    // @todo Remove this check when `\Doctrine\Persistence\Event\LifecycleEventArgs` is used.
    if (\method_exists($this->args, 'getObject')) {
        return $this->args->getObject();
    }

    return $this->args->getEntity();
}

Please, confirm if my understanding about the current situation is right and let me know if I'm missing something.
Thank you.

Comment (DubbleClick) [reply src/Mapping/Event/Adapter/ORM.php:98]: > From what I see, the case seems to be even worse than I thought, as the property used here (ORM::$args) is using a type (Doctrine\Common\EventArgs) that does not guarantee the getEntityManager() or the getEntityManager() methods

Correct, that's why these warnings are ignored for class in the php actions here.

> That said, I know we surely couldn't use LifecycleEventArgs as type declaration here, as there are more implementations of EventArgs passed here (Doctrine\ORM\Event\LoadClassMetadataEventArgs by instance).

I cannot confirm that is the case, but if it is, the current implementation would crash anyway because those EventArgs do not have a `getEntity` method.

From my understanding (and I could definitely be wrong there) all EventArgs actually passed from Doctrine to EventListeners are of type Persistence/LifecycleEventArgs. The only way I could see this breaking bc is if people implemented their own EventArgs implementations with getEntityManager and getEntity methods and these were somehow passed to our ORM Adapter.

Comment (DubbleClick) [reply src/Mapping/Event/Adapter/ORM.php:98]: I'll just add some deprecation notices if the $args don't have a getObjectManager/getObject method and see if these are ever triggered to confirm. Give me a few days to test it.

Comment (phansys) [reply src/Mapping/Event/Adapter/ORM.php:98]: Thank you very much for the time and effort put on this!

In the test I used, I can confirm other types than `Doctrine\Persistence\Event\LifecycleEventArgs` are passed to this class, like `Doctrine\ORM\Event\LoadClassMetadataEventArgs`.

Comment (DubbleClick) [reply src/Mapping/Event/Adapter/ORM.php:98]: Time to figure out why that doesn't cause a crash then. `getEntity` must just never be called on those.

Comment (DubbleClick) [reply src/Mapping/Event/Adapter/ORM.php:98]: Edit: actually, being on my pc, there's a very simple explanation for why that would still be okay:
https://github.com/doctrine/orm/blob/2.14.x/lib/Doctrine/ORM/Event/LoadClassMetadataEventArgs.php

File: tests/Gedmo/Mapping/MappingEventAdapterTest.php [hunk 9]
```diff
@@ -53,10 +53,10 @@ public function testAdapterBehavior(): void
             ->disableOriginalConstructor()
             ->getMock();
         $eventArgsMock->expects(static::once())
```
phansys, line 55: Maybe we could even replace these mocks to use a class that is no deprecated, like `Doctrine\ORM\Event\PrePersistEventArgs` or `Doctrine\ORM\Event\PreUpdateEventArgs`.

tests/Gedmo/Mapping/MappingEventAdapterTest.php
phansys: Regarding `getEventAdapter()`, I stumbled with that yesterday in my tests :disappointed_relieved: 

>GitHub doesn't have the correct emoji to express my feelings about that code. Sometimes I feel like code reflection goes just a little too far in some languages, hahaha.

:dizzy_face: :face_with_spiral_eyes: :rofl: 

>It would be awesome to be able to get rid of this code, but since it doesn't seem easy, maybe we should keep using the deprecated Doctrine\Persistence\Event\LifecycleEventArgs;, once https://github.com/doctrine-extensions/DoctrineExtensions/pull/2669 is merged we can use one of the new dedicated events.

Just for the records, `Doctrine\ORM\Event\LifecycleEventArgs` is deprecated, but `Doctrine\Persistence\Event\LifecycleEventArgs` is not.

DubbleClick:
I didn't run into any deprecation triggers in the last two days with the code of this PR, although I'm only using Loggable, Timestampable and Blameable.

franmomu (approved):
Thanks @DubbleClick! sorry for the back and forth, please add a changelog entry

DubbleClick:
Hi, sorry but I can't access a pc until next week. Will do then.

src/Mapping/Event/Adapter/ORM.php
```diff
@@ -95,6 +95,18 @@ public function getObjectManager()
...
+        if (\method_exists($this->args, 'getObjectManager')) {
+            return $this->args->getObjectManager();
+        }
+
+        @trigger_error(sprintf(
+            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
+            .' and will throw a "%s" error in version 4.0.',
+            __METHOD__,
+            get_class($this->args),
+            \TypeError::class
```
franmomu, line 107: Should this be `\Error::class`?

DubbleClick: I think a TypeError is fine if we look at it as the $args class (type) not implementing the getObjectManager method. If we look at it like "the type $args::class must implement getObjectManager for version 4.0" then it should be \Error.

DubbleClick:
@franmomu I've added the changelog note and changed the \TypeError to \Error

CHANGELOG.md
```diff
@@ -19,7 +19,12 @@ a release.
 
 ## [Unreleased]
 
-## [3.13.0]
+## [3.14.0]
+### Deprecated
+- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager` and `getObject` on EventArgs that do not implement `getObjectManager` and `getObject` (such as old EventArgs implementing `getEntityManager` and `getEntity`)
```
```suggestion
- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager()` and `getObject()` on EventArgs that do not implement `getObjectManager` and `getObject()` (such as old EventArgs implementing `getEntityManager()` and `getEntity()`) 
```

CHANGELOG.md
```diff
@@ -19,7 +19,12 @@ a release.
 
 ## [Unreleased]
 
-## [3.13.0]
+## [3.14.0]
+### Deprecated
+- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager` and `getObject` on EventArgs that do not implement `getObjectManager` and `getObject` (such as old EventArgs implementing `getEntityManager` and `getEntity`) 
+- Calling `Gedmo\Uploadable\Event\UploadableBaseEventArgs::getEntityManager` and `getEntity`. Call `getObjectManager` and `getObject` instead.
```
phansys, line 25:
```suggestion
- Calling `Gedmo\Uploadable\Event\UploadableBaseEventArgs::getEntityManager()` and `getEntity()`. Call `getObjectManager()` and `getObject()` instead. 
```

CHANGELOG.md
```diff
@@ -19,7 +19,12 @@ a release.
 
 ## [Unreleased]
 
-## [3.13.0]
+## [3.14.0]
```
phansys, line 22: This will be added when a new release is created.
```suggestion
```

src/Mapping/Event/Adapter/ORM.php
phansys: If there won't be an `instanceof` check here, I guess we can't even be sure if`\Error` will be thrown. If that is the case, I guess we should say something like "it will be not supported" instead.

raziel057:
Hi, just for information I switched to the https://github.com/DubbleClick/DoctrineExtensions/commits/main branch on my project and works as expected. I still get deprecation related to the use of ``Doctrine\Persistence\Event\LifecycleEventArgs``. That should be fixed by merge of the following PR: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2649

DubbleClick:
@phansys "If there won't be an instanceof check here, I guess we can't even be sure if\Error will be thrown. If that is the case, I guess we should say something like "it will be not supported" instead."

No instanceof check is required, calling a method that doesn't exist throws an \Error.

`Fatal error: Uncaught Error: Call to undefined method OldEventArgs::getObjectManager()`

@raziel057 thanks for the confirmation, but be warned that the fork will likely be deleted the day this is merged :p

src/Mapping/Event/Adapter/ORM.php
DubbleClick: No instanceof check is required, calling a method that doesn't exist throws an \Error.

Fatal error: Uncaught Error: Call to undefined method OldEventArgs::getObjectManager()

raziel057:
@DubbleClick No problem thanks we will switch back as soon as this is merged.

phansys:
>@phansys "If there won't be an instanceof check here, I guess we can't even be sure if\Error will be thrown. If that is the case, I guess we should say something like "it will be not supported" instead."
>No instanceof check is required, calling a method that doesn't exist throws an \Error.

What is strange (and wrong) at first glance to me, is that we will call a method that is not guaranteed by the object we are trying to call it from. In other words, we are allowing a method call without confirming whether that is possible.
I'd like to understand what you think the method's body will be like in the next major release, because I guess we should avoid calling a non-existent method, preventing the call with our own mechanism in the worst case if we can't trust the contract provided by the class or interface.

DubbleClick:
> What is strange (and wrong) at first glance to me, is that we will call a method that is not guaranteed by the object we are trying to call it from. In other words, we are allowing a method call without confirming whether that is possible. I'd like to understand what you think the method's body will be like in the next major release, because I guess we should avoid calling a non-existent method, preventing the call with our own mechanism in the worst case if we can't trust the contract provided by the class or interface.

```php
if (\method_exists($this->args, 'getObject')) {
    return $this->args->getObject();
}

@trigger_error(sprintf(
    'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
    .' and will throw a "%s" error in version 4.0.',
    __METHOD__,
    get_class($this->args),
    \Error::class
), E_USER_DEPRECATED);
```

becomes 

```php
return $this->args->getObject();
```

If the method doesn't exist, PHP throws an \Error.

phansys:
>If the method doesn't exist, PHP throws an \Error.
Just like Doctrine will result in the same error when you attempt to call getEntity() on their EventArgs next release.

Thanks for the clarification.
I don't feel comfortable admitting an internal PHP error (`\Error`) when we already know this problem is part of a design flaw on our side.

DubbleClick:
> > If the method doesn't exist, PHP throws an \Error.
> > Just like Doctrine will result in the same error when you attempt to call getEntity() on their EventArgs next release.
> 
> Thanks for the clarification. I don't feel comfortable admitting an internal PHP error (`\Error`) when we already know this problem is part of a design flaw on our side.

I don't think it's a design flaw - doctrine changed its functionality to deprecate `getEntity()` and replace it by `getObject()`, so we react to it and call `getObject()`. For the current version we keep `getEntity()` as a fallback for people using their own EventArgs implementations, but deprecate it. In the next major version, following Doctrine, we can expect people to have refactored their own EventArgs.

src/Mapping/Event/Adapter/ORM.php
franmomu: Can we maybe add instanceof checks with `LifecycleEventArgs`, `LoadClassMetadataEventArgs`, `ManagerEventArgs` and `OnClearEventArgs`? Maybe they should all extend from `ManagerEventArgs`  in `doctrine/persistence` 🤔 ? For the `getObjectManager()` call I mean, or any other solution?

It would be nice to finally move this forward.

DubbleClick: I don't see any need to do that. The \method_exists call will make sure that it's fully backwards compatible for now.

If anything, that would be a discussion for later when we remove the  \method_exists call and directly call `getObjectManager()`. But even then, I don't see a point in checking the type of the args - all we care about is that they abide by Doctrines interfaces and implement `getObjectManager()` instead of `getEntityManager()` (that we currently also call without checking).

franmomu:
> > If the method doesn't exist, PHP throws an \Error.
> > Just like Doctrine will result in the same error when you attempt to call getEntity() on their EventArgs next release.
> 
> Thanks for the clarification. I don't feel comfortable admitting an internal PHP error (`\Error`) when we already know this problem is part of a design flaw on our side.

@phansys so what do you think we should do here, it would be nice to finally finish and merge this PR.

phansys:
>@phansys so what do you think we should do here, it would be nice to finally finish and merge this PR.

First of all, sorry for the delay on the feedback.

Currently, the PR has some deprecation messages that indicate what will be disallowed in the next major release, but IMO, there is no clear path about how the restriction will be implemented.

```php
    public function getObject(): object
    {
        if (null === $this->args) {
            throw new \LogicException(sprintf('Event args must be set before calling "%s()".', __METHOD__));
        }

        if (\method_exists($this->args, 'getObject')) {
            return $this->args->getObject();
        }

        @trigger_error(sprintf(
            'Calling "%s()" on an event of type "%s" is deprecated since gedmo/doctrine-extensions 3.x'
            .' and will throw a "%s" error in version 4.0.',
            __METHOD__,
            get_class($this->args),
            \Error::class
        ), E_USER_DEPRECATED);

        return $this->args->getEntity();
    }
```

By instance, I don't know if the call to `method_exists()` will be removed because we should be trusting in the type of `$this->args` based on a previous type-safe check, or if we are proposing to let the PHP engine to reach a condition where we call a non-existent method `getObject()` which can be not defining depending on the implementation of `$this->args`.

I think a comment with the code that must replace these deprecations should be enough.
I'm thinking of a kind of instruction like this:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/aaa8c15645df41c6247b039cade29948162e27f6/src/Mapping/Driver/Chain.php#L90-L92

DubbleClick:
> By instance, I don't know if the call to method_exists() will be removed because we should be trusting in the type of $this->args based on a previous type-safe check, or if we are proposing to let the PHP engine to reach a condition where we call a non-existent method getObject() which can be not defining depending on the implementation of $this->args.

Isn't that necessarily the same behaviour as it is now? There's even less of a guarantee that the EventArgs implement `getEntityManager` than there is that they implement `getObjectManager`.

I'm happy to write a todo block for the next major release, though.

phansys:
>Isn't that necessarily the same behaviour as it is now? There's even less of a guarantee that the EventArgs implement getEntityManager than there is that they implement getObjectManager.

If this is the case, I'd choose a path that leaves under our own control the responsibility to throw an error explaining in a clear way why the configured `$this->args` is not valid for the context where it is used.

DubbleClick:
I added instructions and cleared up the error message a little better.

Honestly, I am not even sure how far out of their way people would have to go to have DoctrineExtensions catch their own EventArgs implementations. All the Doctrine events we catch and our own implement `getObjectManager`.

phansys (approved):

franmomu:
Thanks @DubbleClick ❤️

brzuchal:
When to expect a release of this?

DubbleClick:
in all likelihood before orm 3.0, so I wouldn't worry about it :)

franmomu:
I hope to release this soon, but I would like to include in the release also these PR:
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2715
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2716
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2720
https://github.com/doctrine-extensions/DoctrineExtensions/pull/2556

Issue #2641, opened 2023-06-29
[Blameable] Allow UUID fields to be used for Blameable fields


fwolfsjaeger:
## Feature Request

I'm using UUID as primary key fields and I would like to store the users' UUID in the createdBy/updatedBy fields using the Blameable annotation/attribute.

PR #2642, opened 2023-06-29
Allow UUIDs for Blameable fields

fwolfsjaeger:
**Feature request:** #2641

I'm using UUID as primary key fields and I would like to store the users' UUID in the createdBy/updatedBy fields using the Blameable annotation/attribute.

To allow UUIDs, I've extended the `validTypes` array in the annotation/XML/YAML driver with 'uuid'.

**Moved to separate PR:** #2645
In order to convert the given value to whatever type is required, in my case UuidV6 from the Symfony UID component, I've amended the functionality of `src/AbstractTrackingListener.php`. I'm checking for a setter method to use that one instead of setting the property value directly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,7 @@
         "rector/rector": "^0.19",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
         "symfony/phpunit-bridge": "^6.0 || ^7.0",
-        "symfony/uid": "^7.0",
+        "symfony/uid": "^5.4 || ^6.0 || ^7.0",
         "symfony/yaml": "^5.4 || ^6.0 || ^7.0"
     },
     "conflict": {
--- a/composer.json
+++ b/composer.json
@@ -67,6 +67,7 @@
         "rector/rector": "^0.19",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
         "symfony/phpunit-bridge": "^6.0 || ^7.0",
+        "symfony/uid": "^7.0",
         "symfony/yaml": "^5.4 || ^6.0 || ^7.0"
     },
     "conflict": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ a release.
 - Dropped support for PHP < 7.4
 - Dropped support for Symfony < 5.4
 - Dropped support for doctrine/dbal < 3.2
+- Blameable: Added UUID in allowed types list for Blameable fields in Annotation
 
 ### Deprecated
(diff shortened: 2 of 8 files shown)


fwolfsjaeger:
I think it would make sense to move the `src/AbstractTrackingListener.php` changes to another PR and make it configurable. Any objections?

fwolfsjaeger:
I've added the CHANGELOG.md entry, but without requiring ramsey/uuid-doctrine this can't be tested. I can add it as a DEV requirement if you want?

fwolfsjaeger:
Do you require/want any additional changes?

fwolfsjaeger:
It seems that `ramsey/uuid-doctrine` requires PHP 7.4 or higher.

fwolfsjaeger:
I've updated the branch and moved the CHANGELOG.md entry to [Unreleased].

mbabker:
> It seems that `ramsey/uuid-doctrine` requires PHP 7.4 or higher.

Instead of using that package, the tests I set up looking at #2701 uses Symfony's UID and Doctrine Bridge packages (the latter of which is already pulled in thanks to the dev dependency chain).

The PHP version constraint isn't an issue anymore, but, I'd say we'd want to support both `uuid` and `ulid` types in all extensions (both `ramsey/uuid-doctrine` and Symfony use `uuid` as the name for that type, so implicitly supporting one should explicitly support both), and using the Symfony packages would let us test support for both types.

fwolfsjaeger:
When I created this PR, Doctrine was mostly using `ramsey/uuid-doctrine` for UUIDs. As of now all my projects are using `symfony/uid` as well.

I've added `ulid` as allowed type and replaced the DEV package now.

The packages are interchangeable, you just have to replace the generator in the annotation/attribute as well.

composer.json
```diff
@@ -67,6 +67,7 @@
         "rector/rector": "^0.19",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
         "symfony/phpunit-bridge": "^6.0 || ^7.0",
+        "symfony/uid": "^7.0",
```
mbabker, line 70:
```suggestion
        "symfony/uid": "^5.4 || ^6.0 || ^7.0",
```

Need at least 5.4 for PHP 7.4 compat

mbabker (approved):

fwolfsjaeger:
@mbabker Would you mind taking a look at #2645  as well?

franmomu (approved):

franmomu:
Thanks @fwolfsjaeger! and @mbabker

Issue #2644, opened 2023-07-03
Add functionality to use setter-method instead of setting property directly


fwolfsjaeger:
## Feature Request

I have to modify the value that is set for `Blameable`, `IpTraceable`, `Timestampable`, `SoftDeletable`. The easiest solution I can think of is using the setter-method instead of setting the property value directly.

PR #2645, opened 2023-07-03
Add config options to allow fields to be updated using a setter method.

fwolfsjaeger:
**Feature request:** #2644

This change is applied in Blameable, IpTraceable, SoftDeleteable & Timestampable annotations. The new attribute "setterMethod" can be passed to specify the setter method to be used for a field. If no setter method is specified, the property value would be set directly as before.

### DIFF
--- a/src/AbstractTrackingListener.php
+++ b/src/AbstractTrackingListener.php
@@ -20,6 +20,7 @@ use Doctrine\Persistence\Event\ManagerEventArgs;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\NotifyPropertyChanged;
 use Doctrine\Persistence\ObjectManager;
+use Gedmo\Exception\InvalidMappingException;
 use Gedmo\Exception\UnexpectedValueException;
 use Gedmo\Mapping\Event\AdapterInterface;
 use Gedmo\Mapping\MappedEventSubscriber;
@@ -89,7 +90,7 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
                     $new = array_key_exists($field, $changeSet) ? $changeSet[$field][1] : false;
                     if (null === $new) { // let manual values
                         $needChanges = true;
-                        $this->updateField($object, $ea, $meta, $field);
+                        $this->updateField($object, $ea, $meta, $field, $config);
                     }
                 }
             }
@@ -101,7 +102,7 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
                         && null === $changeSet[$field][1];
                     if (!isset($changeSet[$field]) || $isInsertAndNull) { // let manual values
                         $needChanges = true;
-                        $this->updateField($object, $ea, $meta, $field);
+                        $this->updateField($object, $ea, $meta, $field, $config);
                     }
                 }
             }
@@ -152,7 +153,7 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
 
                             if (null === $configuredValues || ($singleField && in_array($value, $configuredValues, true))) {
                                 $needChanges = true;
-                                $this->updateField($object, $ea, $meta, $options['field']);
+                                $this->updateField($object, $ea, $meta, $options['field'], $config);
                             }
                         }
                     }
@@ -184,14 +185,14 @@ abstract class AbstractTrackingListener extends MappedEventSubscriber
             if (isset($config['update'])) {
                 foreach ($config['update'] as $field) {
                     if (null === $meta->getReflectionProperty($field)->getValue($object)) { // let manual values
-                        $this->updateField($object, $ea, $meta, $field);
+                        $this->updateField($object, $ea, $meta, $field, $config);
                     }
                 }
             }
             if (isset($config['create'])) {
                 foreach ($config['create'] as $field) {
                     if (null === $meta->getReflectionProperty($field)->getValue($object)) { // let manual values
-                        $this->updateField($object, $ea, $meta, $field);
+                        $this->updateField($object, $ea, $meta, $field, $config);
                     }
                 }
             }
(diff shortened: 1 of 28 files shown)


fwolfsjaeger:
I would like to add tests for those changes as well. Do you want a separate/duplicated test for every affected annotation/listener?

RomanApunts:
I'm interested with this functionality.

I hope soon will merged to master.

src/AbstractTrackingListener.php
```diff
@@ -208,7 +209,7 @@ abstract protected function getFieldValue($meta, $field, $eventAdapter);
      *
      * @return void
      */
-    protected function updateField($object, $eventAdapter, $meta, $field)
+    protected function updateField($object, $eventAdapter, $meta, $field, array $config = [])
```
phansys, line 212: This change breaks BC.

src/Mapping/Driver/AbstractAnnotationDriver.php
```diff
@@ -133,4 +133,20 @@ protected function getRelatedClassName($metadata, $name)
 
         return class_exists($className) ? $className : '';
     }
+
+    /**
+     * Set the setter method for the given field.
+     */
+    protected function setSetterMethod(string $field, ?string $method, array &$config): void
+    {
+        if (empty($method)) {
```
phansys, line 142:
```suggestion
        if ('' === $method) {
```

phansys (changes requested):

src/AbstractTrackingListener.php
fwolfsjaeger: Right, you're supporting PHP 7.2 ... I'll remove the property type.

src/Mapping/Driver/AbstractAnnotationDriver.php
fwolfsjaeger: `\Gedmo\Mapping\Annotation\Blameable::$setterMethod` is nullable and null by default. I can ensure it's always a string if you want.

src/AbstractTrackingListener.php
fwolfsjaeger: I just checked, type-hinting arrays is available in the supported PHP version. The parameter does have a default value, so unless someone passes anything other than an array this won't change a thing. Are am I missing something?

yakobe:
I'm also interested in this functionality. At the moment I override and extend the BlameableListener to add this functionality but it was made `final` so that wont work for long 😄. Any chance we could have this functionality?

fwolfsjaeger:
Alright, all done:
- I've rebased the branch,
- added the missing logic in the SoftDeleteable listener,
- fixed getting the property name in the annotation drivers
- and added Unit tests for all 4 annotations.

Let me know if you need anything else ;)

fwolfsjaeger:
Any chance that this PR is getting merged?

Even though there is no BC breaking change right now: I could refactor the changes to auto-detect a setter method by simply checking for setFieldName(). That however could break BC as there's no way to stick to the original way of setting the property directly.

src/AbstractTrackingListener.php
```diff
@@ -225,7 +226,18 @@ protected function updateField($object, $eventAdapter, $meta, $field)
...
-        $property->setValue($object, $newValue);
+        if (!empty($config['setterMethod'][$field])) {
+            $reflectionClass = $meta->getReflectionClass();
+            $setterName = $config['setterMethod'][$field];
+
+            if (!$reflectionClass->hasMethod($setterName)) {
+                throw new InvalidMappingException("Setter method - [{$setterName}] does not exist in class - {$meta->getName()}");
+            }
+
+            $reflectionClass->getMethod($setterName)->invoke($object, $newValue);
```
mbabker, line 237: Is there a reason for using reflection to check and call the method here instead of `$object->{$setterName}($newValue)`?  Same for the `SoftDeleteableListener`.

fwolfsjaeger: Calling $object->{$setterName}($newValue) without checking if the method exists would throw a warning if it doesn't. I do prefer an Exception with proper information.

The only reason for using reflection to actually call the method is because it's already loaded for the check.

mbabker: The check can be done with `method_exists()` as well.

I think the only gotcha here is that this could unintentionally call into a non-public setter, whereas `method_exists()` and the direct call restricts it to public scope (unless you check that through reflection as well).

fwolfsjaeger: Yes, of course. I'll replace the reflection checks.

fwolfsjaeger: I was also considering to auto-detect if there is a setter method for the configured field. That would be a bc-breaking change though.

mbabker: Yeah, at this point I think opt-in with config is better than a breaking behavioral change.

mbabker (approved):
Needs rebasing but in general looks good to me.

fwolfsjaeger:
I'll rebase the latest changes right away. Many thanks for taking the time to review the changes!

franmomu:
Thanks for keeping working into this @fwolfsjaeger, in general I see the code fine, but I would like to [avoid `empty` calls](https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/) and we were also trying to not working with array references, I don't know how hard would be to make that change.

fwolfsjaeger:
Sorry, but I'm no longer using DoctrineExtensions. I've created this and another PR about a year ago. That and the incompatibility with DBAL 4 forced me to just add my own listeners for the 3 attributes I was using.

I'll look into the recent deprecations and requested changes, but it might take some time.

yakobe:
This would still be very useful

senaria:
Any news about this? Would be really cool to get this working.

fwolfsjaeger:
Feel free to take over. Because of the long waiting time I am no longer using DoctrineExtensions, sorry.

PR #2649, opened 2023-07-13
use dedicated event eventArgs

### LINKED ISSUE #2648: SoftDeleteable dedicated events eventargs
Hello,

in doctrine 3, the class **Doctrine\ORM\Event\LifecycleEventArgs** will be removed, and so dedicated classes of events should be used instead (_PrePersistEventArgs, PostPersistEventArgs, PostUpdateEventArgs, ..._),

which means the function **createLifecycleEventArgsInstance** in the **Gedmo\Mapping\Event\Adapter\ORM** will be invalid to use which create an instance of the **class Doctrine\ORM\Event\LifecycleEventArgs**

i think there should be dedicated events eventargs for the _PRE_SOFT_DELETE_ and _POST_SOFT_DELETE_ events for both ORM and ODM

i implemented those points in my [fork](https://github.com/yassinefikri/DoctrineExtensions/commit/0fad774cf71df23f742c2da7d982b1d9fb9ae92a)
and would like to open a pull request, and of course, open to suggestions

currenty on doctrine 2, the ORM dedicated eventargs i added extends the **Doctrine\ORM\Event\LifecycleEventArgs**, in doctrine 3 they should extend the **Doctrine\Persistence\Event\LifecycleEventArgs** instead

yassinefikri:
Resolves #2648


src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -120,4 +125,40 @@ protected function getNamespace()
     {
         return __NAMESPACE__;
     }
+
+    /**
+     * create the appropriate event args for pre_soft_delete event
+     *
+     * @return OdmPreSoftDeleteEventArgs|OrmPreSoftDeleteEventArgs
+     */
+    private function createAppropriatePreSoftDeleteEventArgs(object $object, ObjectManager $om)
```
phansys, line 134:
```suggestion
    private function createPreSoftDeleteEventArgs(object $object, ObjectManager $om): \Doctrine\Persistence\Event\LifecycleEventArgs
```

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -120,4 +125,40 @@ protected function getNamespace()
...
+
+        throw new \RuntimeException('Not supported object manager');
+    }
+
+    /**
+     * create the appropriate event args for post_soft_delete event
+     *
+     * @return OdmPostSoftDeleteEventArgs|OrmPostSoftDeleteEventArgs
+     */
+    private function createAppropriatePostSoftDeleteEventArgs(object $object, ObjectManager $om)
```
phansys, line 152:
```suggestion
    private function createPostSoftDeleteEventArgs(object $object, ObjectManager $om): \Doctrine\Persistence\Event\LifecycleEventArgs
```

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -120,4 +125,40 @@ protected function getNamespace()
...
+    {
+        if ($om instanceof EntityManagerInterface) {
+            return new OrmPreSoftDeleteEventArgs($object, $om);
+        }
+
+        if ($om instanceof DocumentManager) {
+            return new OdmPreSoftDeleteEventArgs($object, $om);
+        }
+
+        throw new \RuntimeException('Not supported object manager');
```
phansys, line 144:
```suggestion
        throw new \TypeError(\sprintf('Not supported object manager "%s".', \get_class($om)));
```

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -120,4 +125,40 @@ protected function getNamespace()
...
+    {
+        if ($om instanceof EntityManagerInterface) {
+            return new OrmPostSoftDeleteEventArgs($object, $om);
+        }
+
+        if ($om instanceof DocumentManager) {
+            return new OdmPostSoftDeleteEventArgs($object, $om);
+        }
+
+        throw new \RuntimeException('Not supported object manager');
```
phansys, line 162:
```suggestion
        throw new \TypeError(\sprintf('Not supported object manager "%s".', \get_class($om)));
```

phansys (changes requested):

franmomu:
After this change, looks like `EventInterface::createLifecycleEventArgsInstance` is not used anymore (if so, we should deprecate it):

https://github.com/search?q=repo%3Adoctrine-extensions%2FDoctrineExtensions%20createLifecycleEventArgsInstance&type=code

The problem I see is that we would still triggering deprecations when creating the ORM event, but I don't know if that's avoidable (given that we should still extend from the ORM `LifecycleEventArgs` to keep BC).

franmomu:
I've rebased since there were some conflicts with the latest changes.

src/Mapping/Event/AdapterInterface.php
```diff
@@ -21,7 +21,9 @@
  *
  * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
  *
- * @method LifecycleEventArgs createLifecycleEventArgsInstance(object $object, ObjectManager $manager)
+ * @method LifecycleEventArgs createLifecycleEventArgsInstance(object $object, ObjectManager $manager) @deprecated Use createPreSoftDeleteEventArgs() or createPostSoftDeleteEventArgs() instead
+ * @method LifecycleEventArgs createPreSoftDeleteEventArgs(object $object, ObjectManager $manager)
+ * @method LifecycleEventArgs createPostSoftDeleteEventArgs(object $object, ObjectManager $manager)
```
franmomu, line 26: I would move these methods to [`SoftDeleteableAdapter`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/058e95a34847b32d57b5583a006911a5465dac3f/src/SoftDeleteable/Mapping/Event/SoftDeleteableAdapter.php) since they are only related to that extension.

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -83,7 +83,7 @@ public function onFlush(EventArgs $args)
 
                 $evm->dispatchEvent(
                     self::PRE_SOFT_DELETE,
-                    $ea->createLifecycleEventArgsInstance($object, $om)
+                    $ea->createPreSoftDeleteEventArgs($object, $om)
```
franmomu, line 86: Here we should check if `createPreSoftDeleteEventArgs` exists

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -100,7 +100,7 @@ public function onFlush(EventArgs $args)
 
                 $evm->dispatchEvent(
                     self::POST_SOFT_DELETE,
-                    $ea->createLifecycleEventArgsInstance($object, $om)
+                    $ea->createPostSoftDeleteEventArgs($object, $om)
```
franmomu, line 103: Here we should check if `createPostSoftDeleteEventArgs` exists

src/Mapping/Event/Adapter/ORM.php
```diff
@@ -182,13 +185,36 @@ public function clearObjectChangeSet($uow, $object)
...
+     * Creates a ORM specific PostSoftDeleteEventArgs.
+     *
+     * @param EntityManagerInterface $manager
+     *
+     * @return PostSoftDeleteEventArgs
+     */
+    public function createPostSoftDeleteEventArgs(object $object, ObjectManager $manager): LifecycleEventArgs
+    {
+        return new PostSoftDeleteEventArgs($object, $manager);
     }
```
franmomu, line 219: And move these to `src/SoftDeleteable/Mapping/Event/Adapter/ORM.php`

src/Mapping/Event/Adapter/ODM.php
```diff
@@ -152,13 +155,36 @@ public function clearObjectChangeSet($uow, $object)
...
+     * Creates a ODM specific PostSoftDeleteEventArgs.
+     *
+     * @param DocumentManager $manager
+     *
+     * @return PostSoftDeleteEventArgs
+     */
+    public function createPostSoftDeleteEventArgs(object $object, ObjectManager $manager): LifecycleEventArgs
+    {
+        return new PostSoftDeleteEventArgs($object, $manager);
     }
```
franmomu, line 189: And move these to `src/SoftDeleteable/Mapping/Event/Adapter/ODM.php`

franmomu (commented):
I've left some comments, please add some tests.

yassinefikri:
i've updated tests, let me know if i'm missing something

franmomu:
I wanted to check if we can somehow avoid the `LifecycleEventArgs` deprecation, I was thinking that we can do a couple of things.

First of all, check with `EventManager::hasListeners()` method if there are listeners before creating the events, this way we prevent triggering the deprecation for those who don't have listeners.

And to finally remove the deprecation... I was thinking about creating `PreSoftDeleteEventArgs` and `PostSoftDeleteEventArgs` for ORM extending from the `LifecycleEventArgs` class from persistence and adding the `getEntity()` and `getEntityManager()` methods (trigger deprecations inside these methods).

That would create a BC break for someone with a listener typehinted with `LifecycleEventArgs` from the ORM, in that case we can maybe with Reflection check the listener if it's typehinted and in that case trigger a deprecation telling to change it to the dedicated event and for creating the event we would use `AdapterInterface::createLifecycleEventArgsInstance()`.

If the listener is not typehinted with `LifecycleEventArgs` from the ORM, then we create the new dedicated event. If the listener is calling `getEntity()` or `getEntityManager()` would work, the only problem is if it has instance check with `LifecycleEventArgs` from the ORM, but I guess that's unlikely to happen.

These are some thoughts that I haven't tried and I don't know if that would work.

franmomu:
@mbabker what about this? this is similar to https://github.com/doctrine-extensions/DoctrineExtensions/pull/2725 and maintain a BC layer. When not all the listeners have type-hinted the new event class, it dispatches the old event. 

Given that this is Softdeleteable behavior I think it is important to try not break BC.

> > And to finally remove the deprecation... I was thinking about creating `PreSoftDeleteEventArgs` and `PostSoftDeleteEventArgs` for ORM extending from the `LifecycleEventArgs` class from persistence and adding the `getEntity()` and `getEntityManager()` methods (trigger deprecations inside these methods).
> 
> i don't get what deprecations should be triggered inside the methods, can you provide an example ?

Sorry, I took another approach (adding a new commit), simplifying events and trying to maintain BC.

mbabker:
> @mbabker what about this?

Looks fine to me.

franmomu:
thanks @yassinefikri!

PR #2651, opened 2023-07-20
[Mapping] Fix inherited classes metadata collection

### LINKED ISSUE #2600: [Mapping] Metadata for inherited mapped classes are not well collected
## Subject

The `ExtensionMetadataFactory` is not collecting as expected configuration from metadata in particular conditions.

Here are the conditions I managed to define to get this unexpected behavior:

- Entity using ORM Inherited entity / Mapped super class with any extensions.
- Cache change between entity manipulations.

This [issue](https://github.com/doctrine-extensions/DoctrineExtensions/issues/2463) is similar but has been closed considering that specifying the metadata cache pool solves the problem, which is true but does not solves the root cause of the problem.

@dpfaffenbauer pointed out this [issue](https://github.com/doctrine/orm/issues/8866) from `doctrine/orm` reporting that not all classes metadata are loaded in production and derives on the `ClassMetadataFactory` interface method `hasMetadataFor` usage. The method name is misleading its usage as the interface comments states:

> Checks whether the factory has the metadata for a class loaded already.

This method does not return false if the entity has no metadata, it returns false if the cache does not contain any metadata for the given class which in our case can occur for mapped super classes / inherited entities. 

Therefor the method `isTransient` from the same interface can be used in combination to check if mapped super class metadata should be loaded and cached which is exactly what is missing in the `ExtensionMetadataFactory` to fully collect metadata for mapped super classes.

## Minimal repository with the bug

https://github.com/gaelreyrol/DoctrineExtensions/tree/fix-inherited-classes-metadata-collection

## Steps to reproduce

Run the test `MappingEventSubscriberTest::testGetMetadataFactoryCacheFromDoctrineForSuperClassExtension()`.

https://github.com/gaelreyrol/DoctrineExtensions/blob/fix-inherited-classes-metadata-collection/tests/Gedmo/Mapping/MappingEventSubscriberTest.php#L71

## Expected results

With this [fix](https://github.com/gaelreyrol/DoctrineExtensions/compare/main...gaelreyrol:DoctrineExtensions:fix-inherited-classes-metadata-collection#diff-743309499de09a3a2dccf90b49db367b9ec8a0602fb2567167b156826d3f8107R117) applied in `ExtensionMetadataFactory::getExtensionMetadata()` the test should pass.

If this explanation is self sufficient, I can submit a PR with the branch I created containing the fix and the related test.

Regards

magikid:
All the work here was done by @gaelreyrol, I'm just opening the PR.

Fixes #2600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ a release.
 - Loggable: Remove unfixable deprecation when extending `LoggableListener`
 - Remove unfixable deprecations when extending repository classes
 - Fix error caused by the attempt of "doctrine/annotations" parsing a `@note` annotation
+- Fix bug collecting metadata for inherited mapped classes
 
 ## [3.11.0] - 2023-01-26
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ a release.
 ### Changed
 - Add conflict with "doctrine/orm" >=2.16 (temporary change until code is fixed)
 
+### Fixed
+- Fix bug collecting metadata for inherited mapped classes
+
 ## [3.12.0] - 2023-07-08
 ### Added
 - Tree: `setSibling()` and `getSibling()` methods in the `Node` interface through the BC `@method` annotation
@@ -47,7 +50,6 @@ a release.
 - Loggable: Remove unfixable deprecation when extending `LoggableListener`
 - Remove unfixable deprecations when extending repository classes
 - Fix error caused by the attempt of "doctrine/annotations" parsing a `@note` annotation
-- Fix bug collecting metadata for inherited mapped classes
 
 ## [3.11.0] - 2023-01-26
 ### Added
--- a/src/Mapping/ExtensionMetadataFactory.php
+++ b/src/Mapping/ExtensionMetadataFactory.php
@@ -114,7 +114,7 @@ class ExtensionMetadataFactory
         if (null !== $meta->reflClass) {
             foreach (array_reverse(class_parents($meta->getName())) as $parentClass) {
                 // read only inherited mapped classes
-                if ($cmf->hasMetadataFor($parentClass)) {
+                if ($cmf->hasMetadataFor($parentClass) || !$cmf->isTransient($parentClass)) {
                     assert(class_exists($parentClass));
 
                     $class = $this->objectManager->getClassMetadata($parentClass);
--- a/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
+++ b/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
@@ -105,7 +105,7 @@ final class MappingEventSubscriberTest extends ORMMappingTestCase
 
         $config = $subscriber->getExtensionMetadataFactory($this->em)->getExtensionMetadata($classMetadata);
 
-        self::assertSame([
+        static::assertSame([
             'content' => [
                 'type' => 'md5',
                 'secret' => null,
--- a/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
+++ b/tests/Gedmo/Mapping/MappingEventSubscriberTest.php
@@ -19,7 +19,11 @@ use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
 use Gedmo\Mapping\ExtensionMetadataFactory;
 use Gedmo\Sluggable\SluggableListener;
 use Gedmo\Tests\Mapping\Fixture\Sluggable;
+use Gedmo\Tests\Mapping\Fixture\SuperClassExtension;
+use Gedmo\Tests\Mapping\Mock\Extension\Encoder\EncoderListener;
+use Gedmo\Tests\SoftDeleteable\Fixture\Entity\MappedSuperclass;
 use Psr\Cache\CacheItemPoolInterface;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 final class MappingEventSubscriberTest extends ORMMappingTestCase
 {
@@ -44,7 +48,7 @@ final class MappingEventSubscriberTest extends ORMMappingTestCase
         $this->em = EntityManager::create($conn, $config, new EventManager());
     }
 
-    public function testGetMetadataFactoryCacheFromDoctrine(): void
+    public function testGetMetadataFactoryCacheFromDoctrineForSluggable(): void
     {
         $metadataFactory = $this->em->getMetadataFactory();
         $getCache = \Closure::bind(static function (AbstractClassMetadataFactory $metadataFactory): ?CacheItemPoolInterface {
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ a release.
 - Loggable: Remove unfixable deprecation when extending `LoggableListener`
 - Remove unfixable deprecations when extending repository classes
 - Fix error caused by the attempt of "doctrine/annotations" parsing a `@note` annotation
+- Fix bug collecting metadata for inherited mapped classes
```
X-Coder264, line 48: @magikid  This should go under `Unreleased` , currently it's under `[3.11.1] - 2023-02-20` which is wrong.

src/Mapping/ExtensionMetadataFactory.php
```diff
@@ -114,7 +114,7 @@ public function getExtensionMetadata($meta)
         if (null !== $meta->reflClass) {
             foreach (array_reverse(class_parents($meta->getName())) as $parentClass) {
                 // read only inherited mapped classes
-                if ($cmf->hasMetadataFor($parentClass)) {
+                if ($cmf->hasMetadataFor($parentClass) || !$cmf->isTransient($parentClass)) {
```
franmomu, line 117: I'm wondering if the `$cmf->hasMetadataFor($parentClass) || ` part is necessary since it's confusing what it does and not reliable.

X-Coder264: IMO it can stay as is because the `hasMetadataFor` call is doing just an `isset` check under the hood while the `isTransient` call is more expensive since it goes to the mapping driver which then has to check whether the class with the specified name should have its metadata loaded  or not (for example in the case of the atrribute driver it has to use reflection to check whether the class is mapped as an entity or as a mapped super class or neither).

franmomu: yeah, I guess it would save some calls to the driver, looking at the implementation of `getParentClasses`, they only use `isTransient` 🤔 :
https://github.com/doctrine/persistence/blob/413eb71a22c31c309b5a1cea9701d723cc0d4ae2/src/Persistence/Mapping/AbstractClassMetadataFactory.php#L278-L292

CHANGELOG.md
magikid: Done, sorry, I was confused because there weren't any of the subheadings in the `Unreleased` section.

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
```diff
@@ -64,10 +68,56 @@ public function testGetMetadataFactoryCacheFromDoctrine(): void
...
+        $cacheKey = ExtensionMetadataFactory::getCacheId(SuperClassExtension::class, 'Gedmo\Tests\Mapping\Mock\Extension\Encoder');
+
+        static::assertFalse($cache->hasItem($cacheKey));
+
+        $subscriber = new EncoderListener();
+        $classMetadata = $this->em->getClassMetadata(SuperClassExtension::class);
+
+        $config = $subscriber->getExtensionMetadataFactory($this->em)->getExtensionMetadata($classMetadata);
+
+        self::assertEquals([
```
franmomu, line 90:
```suggestion
        self::assertSame([
```
To make the build green

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
```diff
@@ -64,10 +68,56 @@ public function testGetMetadataFactoryCacheFromDoctrine(): void
...
+        $config->setMetadataDriverImpl(new AnnotationDriver(new AnnotationReader()));
+        $config->setMetadataCache(new ArrayAdapter());
+        $this->em = EntityManager::create([
+            'driver' => 'pdo_sqlite',
+            'memory' => true,
+        ], $config, new EventManager());
+
+        $config = $subscriber->getExtensionMetadataFactory($this->em)->getExtensionMetadata($classMetadata);
+
+        self::assertEquals([
```
franmomu, line 108:
```suggestion
        self::assertSame([
```

BacLuc:
I tested it with the bug that appeared in https://github.com/ecamp/ecamp3
https://github.com/ecamp/ecamp3/issues/3662
And it worked, thank you for the fix.

solverat:
Any updates on this? This PR also successfully fixes a lot of issues on our site.

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
```diff
@@ -64,10 +68,56 @@ public function testGetMetadataFactoryCacheFromDoctrine(): void
...
+        $config->setMetadataDriverImpl(new AnnotationDriver(new AnnotationReader()));
+        $config->setMetadataCache(new ArrayAdapter());
+        $this->em = EntityManager::create([
+            'driver' => 'pdo_sqlite',
+            'memory' => true,
+        ], $config, new EventManager());
+
+        $config = $subscriber->getExtensionMetadataFactory($this->em)->getExtensionMetadata($classMetadata);
+
+        self::assertSame([
```
franmomu, line 108:
```suggestion
        static::assertSame([
```

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
```diff
@@ -64,10 +68,56 @@ public function testGetMetadataFactoryCacheFromDoctrine(): void
...
+        $cacheKey = ExtensionMetadataFactory::getCacheId(SuperClassExtension::class, 'Gedmo\Tests\Mapping\Mock\Extension\Encoder');
+
+        static::assertFalse($cache->hasItem($cacheKey));
+
+        $subscriber = new EncoderListener();
+        $classMetadata = $this->em->getClassMetadata(SuperClassExtension::class);
+
+        $config = $subscriber->getExtensionMetadataFactory($this->em)->getExtensionMetadata($classMetadata);
+
+        self::assertSame([
```
franmomu, line 90:
```suggestion
        static::assertSame([
```

franmomu:
@magikid can you please rebase? I've tried but I couldn't because of permissions 😞

franmomu (approved):

tristanbes:
Just figuring out this bug after hours trying to understand why my timestamps were no longer applied. Thought it was because of doctrine 2.16 (https://github.com/doctrine/orm/issues/10869) but in fact it was this issue. 

Using a timestampable trait on an Abstract entity + prod mode = boom SQLSTATE Exception

Thanks for your PR @magikid

franmomu:
Thanks, if you can fix the coding standard issues shown in https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/5869991947/job/15916135479?pr=2651 that would be all I think

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
franmomu, line 108:
```suggestion
        static::assertSame([
```

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
franmomu, line 90:
```suggestion
        static::assertSame([
```

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
franmomu, line 90:
```suggestion
        static::assertSame([
```

tests/Gedmo/Mapping/MappingEventSubscriberTest.php
franmomu, line 108:
```suggestion
        static::assertSame([
```

franmomu:
Thanks @gaelreyrol and @magikid!

yakobe:
@phansys Would it be possible to get a release with this included? Thanks everyone for helping to solve this 😸 !

phansys:
Sure! [3.13.0](https://github.com/doctrine-extensions/DoctrineExtensions/releases/tag/v3.13.0).

PR #2657, opened 2023-07-27
Add options in listeners: forceUseAttributeReader and separateXmlMapping

### LINKED ISSUE #2613: XML Validation Error with Doctrine 2.14.2
#### Package
Timestampable

```
$ composer show --latest gedmo/doctrine-extensions
name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.11.1
latest   : v3.11.1
```

#### Doctrine packages
```
$ composer show --latest 'doctrine/*'

Direct dependencies required in composer.json:
doctrine/annotations                1.14.3 2.0.1  Docblock Annotations Parser
doctrine/doctrine-bundle            2.9.0  2.9.1  Symfony DoctrineBundle
doctrine/doctrine-migrations-bundle 3.2.2  3.2.2  Symfony DoctrineMigrationsBundle
doctrine/migrations                 3.6.0  3.6.0  PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deplo...
doctrine/orm                        2.14.2 2.14.2 Object-Relational-Mapper for PHP

Transitive dependencies not required in composer.json:
doctrine/cache                      2.2.0  2.2.0  PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis, memcache, apc, mongodb and others.
doctrine/collections                2.1.2  2.1.2  PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
doctrine/common                     3.4.3  3.4.3  PHP Doctrine Common project is a library that provides additional functionality that other Doctrine projects depend on such as better reflection support, proxies a...
doctrine/dbal                       3.6.1  3.6.2  Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.
doctrine/deprecations               v1.0.0 v1.0.0 A small layer on top of trigger_error(E_USER_DEPRECATED) or PSR-3 logging with options to disable all deprecations or selectively for packages.
doctrine/event-manager              2.0.0  2.0.0  The Doctrine Event Manager is a simple PHP event system that was built to be used with the various Doctrine projects.
doctrine/inflector                  2.0.6  2.0.6  PHP Doctrine Inflector is a small library that can perform string manipulations with regard to upper/lowercase and singular/plural forms of words.
doctrine/instantiator               1.5.0  2.0.0  A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                      2.1.0  3.0.0  PHP Doctrine Lexer parser library that can be used in Top-Down, Recursive Descent Parsers.
doctrine/persistence                3.1.4  3.1.4  The Doctrine Persistence project is a set of shared interfaces and functionality that the different Doctrine object mappers share.
doctrine/sql-formatter              1.1.3  1.1.3  a PHP SQL highlighting library

```

#### PHP version

```
$ php -v
8.1.16
```

## Subject

Using XML Schema with a timestampable field:
```
    <field name="createdAt" type="datetime" nullable="true">
      <gedmo:timestampable on="create"/>
    </field>
```
the following validation error is displayed (eg. on cache:clear):
`ibxml error: Element '{http://gediminasm.org/schemas/orm/doctrine-extensions-mapping}timestampable': No matching global element declaration available, but demanded by the strict wildcard. in [file] at line 16`

doctrine/orm 2.14.1 is working fine.

eisberg:
Closes #2613  and #2318

 
### 1. Mapping Driver: Configure usage separately with Doctrine `.orm.xml` -> `.gedmo.xml` files

Thanks to this option, we can transfer options related to setting up Gedmo properties from a common configuration file with the doctrine to a separate EnittyName file.gedmo.xml !

```php
$blameableListener = new BlameableListener();
$blameableListener->setSeparateXmlMapping(true);
```

Example MyEntity.gedmo.xml:

```xml
<?xml version="1.0" encoding="utf-8"?>
<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping https://www.doctrine-project.org/schemas/orm/doctrine-mapping.xsd"
                  xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping"
>
    <entity name="MyEntity">
        <field name="lastEditedBy" >
                <gedmo:blameable on="update"/>
        </field>
    </entity>
</doctrine-mapping>

```

Default value: FALSE

### 2. Alternative option: force the use of AttributeReader and ignore the configuration of the Doctrine chain driver 

```php
$blameableListener = new BlameableListener();
$blameableListener->setForceUseAttributeReader(true);
...
# Entity:
    #[Gedmo\Blameable(on: 'create')]
    private $createdBy;
```

After that, the Gedmo options will be read from the attributes, despite the fact that the basic configuration of the mapping doctrine is read in XML files

Default value: FALSE


mbabker:
I have to be honest, the more I think about this PR, the more I'm not sure I agree with it.  While the XSD validation issue is a legitimate concern (and one the ORM maintainers took into account before releasing 3.0 by reverting the mandatory validation, and the MongoDB ODM maintainers closely following the relevant ORM issues), I think both ideas proposed in this PR just makes things more complicated from an end user and a maintainer perspective.

One of the biggest benefits to piggybacking off the Doctrine mapping API is all of the problems they've already solved are solved for us.  Separating out the mapping (either through separate XML configs or having models with XML definitions and requiring PHP attributes to add this package's functionality) seems like it would re-introduce a lot of those kinds of solved problems.  How well does this PR account for models with inheritance (which is a big thing in the Sylius ecosystem where model schemas are defined with XML, and is functionality provided through this package with a number of mapped superclasses)?

If this PR is to move forward, it will need a lot of tests covering this behavior as it is a huge behavioral change if you opt into either of these features.

eisberg:
There is nothing self-written in this solution, only (1) the doctrine XMLReader is launched for separate files OR (2) the standard doctrine AttributeDriver is launched instead of the one installed in the general configuration. Well, besides, when these options are disabled, the standard behavior is preserved. It's just a configuration manipulation of which reader to use.

We have been successfully using the patch from this PR for a year now, in a product that has about 1000 ORM models. Well, this is just a little confirmation that it works.
I will be glad if an alternative and better solution is proposed. Or, if someone takes this PR and writes a lot of tests for this behavior.
But for now, without this patch, we cannot use this module.

Convert all our ORMs to attributes, only because DoctrineExtension does not support XML, or rather, Doctrine forbade writing something third-party, dubious activity in its files.
Sorry, I’m judging from a consumer point of view, and it’s easier for us to abandon this library altogether.


PS: During the year of the known problem, no alternative was proposed

eisberg:
Add other PR: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2845
With only attributes reader.

PR #2660, opened 2023-08-02
chore: drop pinned versions for dockerfile

Chris53897:
Will fix https://github.com/doctrine-extensions/DoctrineExtensions/issues/2624
The other fixed versions are a problem today (they worked at the time at the creation of the issue). It looks that these versions are not available/reachable anymore on mac.

Are the fixed versions necessary?
If yes, i would suggest do duplicate the actual Docker-File `Dockerfile-fallback-mac`? as fallback for mac.

### DIFF
--- a/.docker/php/Dockerfile
+++ b/.docker/php/Dockerfile
@@ -10,11 +10,11 @@ COPY --from=composer /usr/bin/composer /usr/bin/composer
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-    git=1:2.30.2-1 \
-    libzip-dev=1.7.3-1 \
-    unzip=6.0-26+deb11u1 \
-    zip=3.0-12 \
-    zlib1g-dev=1:1.2.11.dfsg-2+deb11u2 \
+    git \
+    libzip-dev \
+    unzip \
+    zip \
+    zlib1g-dev \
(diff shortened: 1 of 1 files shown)


.docker/php/Dockerfile
```diff
@@ -10,11 +10,11 @@ COPY --from=composer /usr/bin/composer /usr/bin/composer
 
 RUN apt-get update \
```
arnaud-deabreu, line 11:
```suggestion
# hadolint ignore=DL3008
RUN apt-get update \
```

**suggestion:** adding this line will allow unpinned package version : https://github.com/hadolint/hadolint/wiki/DL3008

Chris53897: @arnaud-deabreu Thanks for this info. Do you know, if there is a mode to use the pinned versions, with a fallback to unpinned versions?

If not, i guess it is the best, that i revert my changes. And only add your suggested commit (with a comment) to this PR.

phansys: I think you could try with something like this:
```shell
apt-get satisfy -qq --yes --no-install-recommends \
    "git (>= 1:2.30.2), git (<< 1:3)"
```

Chris53897:
I will close this, as the linked PR solves this issue better.

PR #2661, opened 2023-08-02
fix: add conflict for doctrine/orm >= 2.16.0

Chris53897:
https://github.com/doctrine-extensions/DoctrineExtensions/issues/2659

This helps to get the ci green. 
But does not prevent the installation on userside of older versions with 3.12.0 and doctrine-orm >= 2.16.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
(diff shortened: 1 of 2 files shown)


franmomu (approved):
We should probably update several things to make it work, I think we can go with this for now and try to see if these bugs are in our code.

PR #2668, opened 2023-08-15
Do not rely on autoincremental ids

franmomu:
These tests fails using `doctrine/orm` 2.16 since they rely on autoincremental ids

### DIFF
--- a/tests/Gedmo/Translator/TranslatableTest.php
+++ b/tests/Gedmo/Translator/TranslatableTest.php
@@ -102,7 +102,7 @@ final class TranslatableTest extends BaseTestCaseORM
         $person->translate('ru')->setDescription('multilingual description');
 
         $parent = new Person();
-        $parent->setName('Jen');
+        $parent->setName('Jen parent');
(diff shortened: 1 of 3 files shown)


tests/Gedmo/Translator/TranslatableTest.php
```diff
@@ -102,7 +102,7 @@ public function testShouldTranslateRelation(): void
         $person->translate('ru')->setDescription('multilingual description');
 
         $parent = new Person();
-        $parent->setName('Jen');
+        $parent->setName('Jen parent');
```
franmomu, line 105: In this case when calling https://github.com/doctrine-extensions/DoctrineExtensions/blob/4fd6a62deb3f4406435e5a869f3b33e7560f5c11/tests/Gedmo/Translator/TranslatableTest.php#L117 it fetches the first one persisted, here we're interested in fetching the child.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2669, opened 2023-08-17
fix EntityManager::create deprecations

DubbleClick:


mbabker:
You'd need to bump the minimum ORM version to 2.13, as well.  In 2.12 and earlier, the constructor is protected, so those versions have to use the (now deprecated) `create()` method.

DubbleClick:
> You'd need to bump the minimum ORM version to 2.13, as well. In 2.12 and earlier, the constructor is protected, so those versions have to use the (now deprecated) `create()` method.

Good catch, fixed.

doc/annotations.md
```diff
@@ -149,15 +149,16 @@ $evm->addEventSubscriber($sortableListener);
 // mysql set names UTF-8 if required
 $evm->addEventSubscriber(new Doctrine\DBAL\Event\Listeners\MysqlSessionInit());
 // DBAL connection
-$connection = array(
+$conn = array(
```
phansys, line 152: I'd use `$driverConfig`, `$params`, `$settings` or similar in order to avoid the ambiguity between `$conn` and `$connection`.

DubbleClick: called it $driverParams because $driverConfig would be similar to the second parameter ($config) and the parameter is actually called $params in the method.

DubbleClick:
I assume phpunit hiccuped for no particular reason as I see no correlation to the changes in the PR. Would you trigger workflows again @phansys ?

DubbleClick:
Well, I have no idea what php 7.2's problem is.

franmomu (commented):
`DriverManager::getConnection()` receives an `EventManager` as third argument, I haven't checked why it only fails in using the lowest dependencies.

franmomu:
> `DriverManager::getConnection()` receives an `EventManager` as third argument, I haven't checked why it only fails using the lowest dependencies.

hmm probably it is not necessary to pass the eventmanager to the EntityManager

franmomu:
> `DriverManager::getConnection()` receives an `EventManager` as third argument, I haven't checked why it only fails using the lowest dependencies.

apparently the third argument of the `EntityManager` constructor [was added in 2.14](https://github.com/doctrine/orm/pull/9961), that's probably why it fails when installing the lowest dependencies.

I think we should move the `EventManager` argument to the `DriverManager::getConnection()` call and it should be fine (and do not pass it to `EntityManager::__construct()`).

DubbleClick:
Bumped the minimum requirement and also added the conflict. I'd have never realised the third argument was only added in 2.14, thanks for looking into it.

franmomu (approved):
@phansys should we defer this to next minor version? we are dropping support of `doctrine/orm` < 2.14 so I guess a changelog and a minor version is needed

phansys:
>@phansys should we defer this to next minor version? we are dropping support of doctrine/orm < 2.14 so I guess a changelog and a minor version is needed

Indeed.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Removed
+- Support for `doctrine/orm` < 2.14
```
phansys, line 22: IMO, if we are not dropping support for a major version, we don't need an entry here.
On the other hand, if these changes are intended to fix deprecations, I think we should add an entry in the `Fixed` section what deprecations are fixed.

composer.json
```diff
@@ -71,7 +71,7 @@
     "conflict": {
         "doctrine/dbal": "<2.13.1 || ^3.0 <3.2",
         "doctrine/mongodb-odm": "<2.3",
-        "doctrine/orm": "<2.10.2 || 2.16.0 || 2.16.1",
+        "doctrine/orm": "<2.14.0 || 2.16.0 || 2.16.1",
```
phansys, line 74:
```suggestion
        "doctrine/orm": "2.16.0 || 2.16.1",
```

phansys (changes requested):

CHANGELOG.md
franmomu: 👍 in this case these deprecations are only triggered in tests

composer.json
franmomu: Here we need at least `2.14` to use the `EntityManager` constructor.

phansys: My bad, I didn't see that the `"doctrine/orm": "^2.14.0"` constraint is in the `require-dev` section.
I'm sorry, I think you can use your original approach.

phansys (approved):

composer.json
DubbleClick: So what's the plan here, keep the 2.14 in the require, but <2.10.2 in the conflict? Anything I can do?

franmomu:
thanks @DubbleClick!

composer.json
franmomu: I should've kept the original `<2.14` conflict 😞, but let's solve it in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2639

PR #2673, opened 2023-08-19
Configure "global_namespace_import" CS rule

franmomu:
Ref https://github.com/doctrine-extensions/DoctrineExtensions/pull/2626#issuecomment-1596006700

We should choose what we want to import, [Symfony uses](https://cs.symfony.com/doc/rules/import/global_namespace_import.html#rule-sets):  

`['import_classes' => false, 'import_constants' => false, 'import_functions' => false],`

I've added that, but personally I import classes, so if we want to change this, no objection.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -42,6 +42,7 @@ return (new PhpCsFixer\Config())
         // @todo: Change the following rule to `true` in the next major release.
         'declare_strict_types' => false,
         'error_suppression' => true,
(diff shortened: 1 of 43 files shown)


phansys (approved):

phansys:
>I've added that, but personally I import classes, so if we want to change this, no objection.

The provided configuration is the one I'm more familiarized with :+1:

phansys:
Thanks @franmomu!

PR #2675, opened 2023-08-20
Use constants for arbitrary annotation values

phansys:
## To Do:
- [ ] Add the remaining annotation constants.

### DIFF
--- a/src/Mapping/Annotation/Slug.php
+++ b/src/Mapping/Annotation/Slug.php
@@ -27,12 +27,22 @@ final class Slug implements GedmoAnnotation
 {
     use ForwardCompatibilityTrait;
 
+    public const STYLE_LOWER = 'lower';
+    public const STYLE_UPPER = 'upper';
+    public const STYLE_CAMEL = 'camel';
+    public const STYLE_ORIGINAL = 'default';
+    public const DATE_FORMAT_DEFAULT = 'Y-m-d-H:i';
+
     /** @var string[] @Required */
     public $fields = [];
     /** @var bool */
     public $updatable = true;
-    /** @var string */
-    public $style = 'default'; // or "camel"
+    /**
+     * @var string
+     *
+     * @phpstan-var self::STYLE_*
+     */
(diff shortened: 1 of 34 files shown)


src/Mapping/Annotation/TrackingAwareAnnotationInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Mapping\Annotation;
+
+/**
+ * @internal
```
stof, line 15: If you expect projects to use those constants, I don't think it should be internal

src/Mapping/Annotation/TrackingAwareAnnotationInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+interface TrackingAwareAnnotationInterface extends Annotation
```
stof, line 17: instead of an interface extending Annotation, I would suggest putting those constants on an enum-like class (being final and forbidding instantiation thanks to a private constructor).
If the PHP requirements were allowing it, those events would be a good candidate for an actual enum, as it is a closed list.

src/Mapping/Annotation/Slug.php
```diff
@@ -27,12 +27,22 @@ final class Slug implements GedmoAnnotation
 {
     use ForwardCompatibilityTrait;
 
+    public const STYLE_LOWER = 'lower';
+    public const STYLE_UPPER = 'upper';
+    public const STYLE_CAMEL = 'camel';
+    public const STYLE_ORIGINAL = 'default';
```
stof, line 33: Those styles being a closed list also look like a good candidate for an enum class

Issue #236, opened 2012-01-17
doctrine:generate:entities fails


rodimus:
i have got latest update
calling doctrine:generate:entities and it fails with error:

Fatal error: Call to a member function getProperties() on a non-object in C:\wam
p\www\tonic2\vendor\gedmo-doctrine-extensions-2.2\lib\Gedmo\Timestampable\Mappin
g\Driver\Annotation.php on line 67

PR #2679, opened 2023-08-20
Remove workaround introduced in #236

phansys:
See:
* #236;
* 8c812c33df673e58f9313a360b0e03a694a3c64c.

As [`ClassMetadata::getReflectionClass()`](https://github.com/doctrine/persistence/blob/8a710c21dfe6e88df21c371012be1bf6f2337525/src/Persistence/Mapping/ClassMetadata.php#L34-L39) can't return falsy values, I guess we must remove this workaround and trust the contract defined there.
If some implementation is not respecting the contract, we should focus our efforts on fixing that implementation if it's worth it.

WDYT?


franmomu (approved):

franmomu:
thanks @phansys!

PR #2684, opened 2023-08-29
Do not rely on autoincremental ids in materialized path tests

franmomu:
First of all, I'm not sure if this is the right change, I'm not familiarized with materialized path, I've seen is that it stores the ids in the path, since we cannot rely on the autoincremental ids, this PR finds the root node of a particular node and compares that their `treeNodeValue` are the same.

This tries to fix the remaining failing test when using [`doctrine/orm` 2.16.2](https://github.com/doctrine-extensions/DoctrineExtensions/issues/2659#issuecomment-1690594815)

### DIFF
--- a/tests/Gedmo/Tree/MaterializedPathORMTest.php
+++ b/tests/Gedmo/Tree/MaterializedPathORMTest.php
@@ -87,10 +87,10 @@ final class MaterializedPathORMTest extends BaseTestCaseORM
         static::assertSame(3, $category3->getLevel());
         static::assertSame(1, $category4->getLevel());
 
-        static::assertSame('1-4', $category->getTreeRootValue());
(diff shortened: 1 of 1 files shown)


tests/Gedmo/Tree/MaterializedPathORMTest.php
```diff
@@ -165,4 +165,13 @@ protected function getUsedEntityFixtures(): array
             self::CATEGORY,
         ];
     }
+
+    private function generateTreeNodeValue(MPCategory $category): string
```
phansys, line 169: IMO, this method doesn't seem to generate a value, but instead performs a find or get operation.

phansys (approved):

phansys:
Thanks @franmomu!

PR #2687, opened 2023-09-03
Remove conflict with doctrine/orm >= 2.16.0

franmomu:
Reverts https://github.com/doctrine-extensions/DoctrineExtensions/pull/2661 (there was no release)

should we add a conflict with `2.16.0` and `2.16.1`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,12 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
-### Changed
-- Add conflict with "doctrine/orm" >=2.16 (temporary change until code is fixed)
 ### Fixed
 - References: fixed condition in `XML` Driver that did not allow to retrieve from the entity definition the `mappedBy` and `inversedBy` fields.
-
-### Fixed
 - Fix bug collecting metadata for inherited mapped classes
 
 ## [3.12.0] - 2023-07-08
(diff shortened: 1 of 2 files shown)


phansys:
>should we add a conflict with 2.16.0 and 2.16.1?

Sorry, I'm not following the history of the issue.
Is there a change in a 3rd party dependency that motivates this change or is it a improvement for the approach provided in #2661?

franmomu:
> > should we add a conflict with 2.16.0 and 2.16.1?
> 
> Sorry, I'm not following the history of the issue. Is there a change in a 3rd party dependency that motivates this change

yeah, for reference: https://github.com/doctrine-extensions/DoctrineExtensions/issues/2659

So in `doctrine/orm` 2.16.0 https://github.com/doctrine/orm/pull/10547 cause multiple failures in our tests (apart from relying on autoincremental id) and after https://github.com/doctrine/orm/pull/10915 (released in `2.16.2`) these tests are passing green again, so there is no need for the conflict anymore.

phansys:
Thanks for the clear explanation.
I agree we should prevent the conflicting versions then. I guess we can update the constraint in the `require` section to disallow the offending versions.

phansys (approved):

phansys:
Thank you @franmomu!

PR #2688, opened 2023-09-03
Allow "php" container to send `SIGINT` signal to the foreground processes

phansys:
See:
* https://docs.docker.com/compose/compose-file/compose-file-v2/#init;
* https://github.com/docker-library/php/issues/505.

This PR also adds the `healthcheck` definition for the `mysql` service. See https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck.


franmomu (approved):

franmomu:
thanks @phansys!

PR #2686, opened 2023-09-03
Fix some findings from PHPStan level 6

phansys:

PR #2689, opened 2023-09-03
Use different methods than `execute()` when retrieving results

phansys:
This should also contribute reducing the resource consumption where iterators are used instead of arrays.
Calls to `compact()` were also replaced in order to consume the related variables explicitly.


phansys:
I guess it would be better to merge #2686 first, just in case.

PR #2699, opened 2023-09-10
Fix generic type implementations

phansys:
Follows #2686.

Extra calls to `Iterator::rewind()` in `MaterializedPathODMMongoDBRepositoryTest` were also removed.

### DIFF
--- a/src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
+++ b/src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
@@ -21,6 +21,9 @@ use Gedmo\Tree\RepositoryUtils;
 use Gedmo\Tree\RepositoryUtilsInterface;
 use Gedmo\Tree\TreeListener;
 
+/**
+ * @phpstan-extends DocumentRepository<object>
+ */
 abstract class AbstractTreeRepository extends DocumentRepository implements RepositoryInterface
 {
     /**
--- a/tests/Gedmo/Tree/MaterializedPathODMMongoDBRepositoryTest.php
+++ b/tests/Gedmo/Tree/MaterializedPathODMMongoDBRepositoryTest.php
@@ -12,7 +12,6 @@ declare(strict_types=1);
 namespace Gedmo\Tests\Tree;
 
 use Doctrine\Common\EventManager;
-use Doctrine\ODM\MongoDB\Iterator\CachingIterator;
 use Doctrine\ODM\MongoDB\Iterator\Iterator;
 use Gedmo\Exception\InvalidArgumentException;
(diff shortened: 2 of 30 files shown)


tests/Gedmo/Tree/MaterializedPathODMMongoDBRepositoryTest.php
```diff
@@ -103,22 +99,20 @@ public function testGetChildren(): void
...
         $result->next();
         static::assertSame('Fruits', $result->current()->getTitle());
         $result->next();
         static::assertSame('Vegitables', $result->current()->getTitle());
 
         // Get direct children from the root, NOT including it
-        /** @var CachingIterator $result */
         $result = $this->repo->getChildren($root, true, 'title', 'asc', false);
-        static::assertInstanceOf(Iterator::class, $result);
+        static::assertInstanceOf(CachingIterator::class, $result);
```
franmomu, line 115: I think we should stick to `Doctrine\ODM\MongoDB\Iterator\Iterator` in these `static::assertInstanceOf()` calls since [`CachingIterator` is internal](https://github.com/doctrine/mongodb-odm/blob/69e64e7eb51f07bcfc09bc8e7ea8329f301b44ff/lib/Doctrine/ODM/MongoDB/Iterator/CachingIterator.php#L26)

src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
```diff
@@ -21,6 +21,9 @@
 use Gedmo\Tree\RepositoryUtilsInterface;
 use Gedmo\Tree\TreeListener;
 
+/**
+ * @phpstan-extends DocumentRepository<object>
+ */
```
franmomu, line 26: Based on https://github.com/doctrine-extensions/DoctrineExtensions/pull/2556 and https://github.com/doctrine-extensions/DoctrineExtensions/pull/2543 we should maybe make it generic

phansys: As this PR is focused on resolving current findings and not in adding more extension points, I guess we can let that responsibility on those PRs.
WDYT?

tests/Gedmo/Tree/MaterializedPathODMMongoDBRepositoryTest.php
phansys: Done!

src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
franmomu: 👍 just wondering if, as a user, static analysis issues would appear, I guess not because there are not in `tests`.

franmomu (approved):

PR #2703, opened 2023-09-27
[Translatable][Translator] unify alias usage of doctrine `ORM` mappin…

### LINKED ISSUE #2702: [Translatable][Translator] unify alias usage of doctrine ORM mapping in entities
## Feature Request

We came across an inconsistency in the annotations of the Translatable and Translator entities that lead to some php warnings in our custom class loader implementation (based on an old zend loader). Of course the current implementation works as intended but maybe it would make sense to change it for more consistency.

In Gedmo\Translatable\Entity\Translation and Gedmo\Translator\Entity\Translation the doctrine annotations are used without `ORM` aliases as in all the other entities (e.g. Gedmo\Loggable\Entity\LogEntry). Is there a specific reason for this differentiating?

m-idler:
…g in entities

fixes #2702

### DIFF
--- a/src/Translatable/Entity/Translation.php
+++ b/src/Translatable/Entity/Translation.php
@@ -9,40 +9,35 @@
 
 namespace Gedmo\Translatable\Entity;
 
-use Doctrine\ORM\Mapping\Entity;
-use Doctrine\ORM\Mapping\Index;
-use Doctrine\ORM\Mapping\Table;
-use Doctrine\ORM\Mapping\UniqueConstraint;
+use Doctrine\ORM\Mapping as ORM;
 use Gedmo\Translatable\Entity\MappedSuperclass\AbstractTranslation;
 use Gedmo\Translatable\Entity\Repository\TranslationRepository;
 
 /**
  * Gedmo\Translatable\Entity\Translation
  *
- * @Table(
+ * @ORM\Table(
  *     name="ext_translations",
  *     options={"row_format": "DYNAMIC"},
  *     indexes={
- *
- *         @Index(name="translations_lookup_idx", columns={
+ *         @ORM\Index(name="translations_lookup_idx", columns={
  *             "locale", "object_class", "foreign_key"
  *         }),
- *         @Index(name="general_translations_lookup_idx", columns={
+ *         @ORM\Index(name="general_translations_lookup_idx", columns={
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ a release.
 ---
 
 ## [Unreleased]
+- unify namespace usage of doctrine ORM mappings in `Translatable` and `Translator` entities (#2702)
```
phansys, line 21: Is this change fixing some bug, changing an existing behavior or adding a new feature?
If not, I think we should omit the changelog in this case.

phansys (commented):
Congratulations for your first contribution @m-idler! :heart:

CHANGELOG.md
m-idler: No, actually it isn't - I was just following the notes in the [Contributing](https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/CONTRIBUTING.md#changelog) file. So if you want to, i can remove it.

m-idler: I removed it ✅

src/Translatable/Entity/Translation.php
```diff
@@ -9,40 +9,37 @@
...
 
 /**
  * Gedmo\Translatable\Entity\Translation
  *
- * @Table(
+ * @ORM\Table(
  *     name="ext_translations",
  *     options={"row_format": "DYNAMIC"},
  *     indexes={
  *
```
franmomu, line 23:
```suggestion
```
According to https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/6693389572/job/18194664080?pr=2703

m-idler: Fixed it, somehow came in by the rebasing before

franmomu: Yep, out of curiosity I guess it was because this: 

https://github.com/doctrine-extensions/DoctrineExtensions/blob/058e95a34847b32d57b5583a006911a5465dac3f/.php-cs-fixer.dist.php#L62-L66

before this change they weren't in those groups

franmomu (approved):

franmomu:
thanks @m-idler!

PS: I've removed the `New Feature` label since apparently it is not, feel free to re-add it @phansys

Issue #2708, opened 2023-10-11
ORM 3.0 compat


mbabker:
The ORM [now has a 3.0 beta release](https://github.com/doctrine/orm/releases/tag/3.0.0-beta1) so it's probably time to start looking at that upgrade guide.

mbabker:
There are a few things that are coming up fairly regularly:

- `Call to undefined method Doctrine\ORM\UnitOfWork::clearEntityChangeSet()` - The `Gedmo\Mapping\Event\Adapter\ORM::clearObjectChangeSet()` method relies on being able to call a method on each manager's unit of work to remove a change set for an object.  The ORM removed this method in 3.0.
- `Class "Doctrine\ORM\Mapping\ClassMetadataInfo" not found` - The `Doctrine\ORM\Mapping\ClassMetadata` class is now used
- `Undefined constant Doctrine\ORM\Mapping\ClassMetadata::CHANGETRACKING_NOTIFY` - There are test fixtures configured with the "NOTIFY" change tracking policy, which was removed.  The ORM suggests "DEFERRED_EXPLICIT" instead, I didn't really dig in here to see if those tests purposefully validate behavior with that policy.
- `Indirect modification of overloaded element of Doctrine\ORM\Mapping\ManyToOneAssociationMapping has no effect` - This one's interesting because the stack trace starts at the PHPUnit Bridge's error handler

```sh
/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:132
/src/Tree/Strategy/ORM/Closure.php:462
/src/Tree/Strategy/ORM/Closure.php:286
/src/Tree/TreeListener.php:247
```

And this is on top of the removal of the annotation and YAML mapping drivers, which a lot of the tests use.

mbabker:
In that catch-all branch, I've got the `Indirect modification of overloaded element` error taken care of now.  So really it's just down to three huge focus areas:

- Test cases using fixtures that are annotation or YAML only (which may or may not uncover new issues as those get adapted to run with the other drivers)
- The removal of `Doctrine\ORM\UnitOfWork::clearEntityChangeSet()`
- The strict XML validation issue (it's been an issue with the MongoDB ODM since its 2.0 release, see #2055, and now ORM 3.0 has it)

mbabker:
With the tests run against the ORM 3.0 RC, we're down to:

[...]

PR #2709, opened 2023-10-12
ORM 3 compat fixes

mbabker:
This is some low-hanging fruit working as a very first pass for #2708.  Best viewed by individual commits.

Three things are tackled:

- The `Doctrine\ORM\EntityRepository::$_em` property was made private, the getter method is used instead
- The `Doctrine\ORM\Internal\Hydration\AbstractHydrator::$_em` property was renamed to `$em`, and it has no getter method; an internal trait was introduced and implemented in all ORM hydrators to support retrieving the entity manager from the parent class in a cross-version compatible manner
- The unit tests will now use the ORM's `AttributeDriver` instead of the deprecated `AnnotationDriver` when running on PHP 8 when configuring an entity manager

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -376,6 +376,16 @@ parameters:
 			path: src/Tool/Wrapper/MongoDocumentWrapper.php
 
 		-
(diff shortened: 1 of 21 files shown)


phansys (approved):

franmomu (approved):

franmomu:
thanks @mbabker!

PR #2710, opened 2023-10-17
Refactor ORM `QueryBuilder::setParameters()` calls to use `setParameter()` instead due to a stricter type in 3.0

mbabker:
For B/C, the ORM's `QueryBuilder::setParameters()` (plural) method has continued to accept a plain array throughout 2.x, but in 3.0 it is typehinted as an `ArrayCollection`.  So instead of having to pass a collection of Parameter objects, emulating the internals of `setParameter()` (singular), this just updates the code to call the `setParameter()` method one-by-one (which shouldn't hurt any since these are only 2-item arrays).

After #2639 it should be more feasible to see where some of the pain points lie, but for now, another piece of low-hanging fruit is knocked down.


franmomu (approved):

phansys (approved):

phansys:
Thank you @mbabker!

PR #2711, opened 2023-10-23
Use php-cs-fixer shim to test against doctrine/annotations 2

alexander-schranz:
The current used php-cs-fixer version blocks the ugprade to doctrine/annotations 2 and so it never really tested with it. Using the official php-cs-fixer shim does not block update.

Bildschirmfoto 2023-10-23 um 12 24 58

https://github.com/PHP-CS-Fixer/shim


alexander-schranz:
Looks likethis is only a local problem and is handled in the CI correctly: https://github.com/doctrine-extensions/DoctrineExtensions/blob/43073e0a1932671535f2f5f0b31bb7fa02594db0/.github/workflows/continuous-integration.yml#L67

PR #2715, opened 2023-10-29
Use the Persistence Proxy interface for checking ORM proxy objects

mbabker:
Ref: #2708

In ORM 2.14, `Doctrine\ORM\Proxy\Proxy` was deprecated and the upgrade guide suggests using `Doctrine\Persistence\Proxy` instead, with the ORM interface removed in 3.0.  This updates all checks to reflect this deprecation.

### DIFF
--- a/src/References/Mapping/Event/Adapter/ODM.php
+++ b/src/References/Mapping/Event/Adapter/ODM.php
@@ -11,7 +11,7 @@ namespace Gedmo\References\Mapping\Event\Adapter;
 
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ORM\EntityManagerInterface;
-use Doctrine\ORM\Proxy\Proxy as ORMProxy;
+use Doctrine\Persistence\Proxy as PersistenceProxy;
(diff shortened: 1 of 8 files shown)


franmomu (approved):

src/References/Mapping/Event/Adapter/ODM.php
```diff
@@ -32,7 +32,7 @@ public function getIdentifier($om, $object, $single = true)
         }
 
         if ($om instanceof EntityManagerInterface) {
-            if ($object instanceof ORMProxy) {
+            if ($object instanceof PersistenceProxy) {
```
phansys, line 35: I'm wondering if relaxing this check deserves an entry in the changelog, since previous to this change only instances of `Doctrine\ORM\Proxy\Proxy` were admitted in this condition.

mbabker: I'd say this is more of an internal change to match the upstream API changes and not a deliberate feature change in this package.  So not necessarily something that needs to be called out in a changelog entry.

franmomu:
thanks @mbabker!

PR #2716, opened 2023-10-29
Accept both `ClassMetadata` and `ClassMetadataInfo` objects from the ORM

mbabker:
Ref: #2708

The `Doctrine\ORM\Mapping\ClassMetadataInfo` class is deprecated in ORM 2.x and removed entirely in 3.0, with `Doctrine\ORM\Mapping\ClassMetadata` being the replacement.  `Doctrine\ORM\Mapping\ClassMetadata` exists in 2.x as a subclass of `Doctrine\ORM\Mapping\ClassMetadataInfo`, so no version conditional behaviors are really needed here, therefore, this PR just adds both ORM classes to some existing doc block types and assertions (I'm aware there are other places using the deprecated `ClassMetadataInfo` class but the changes here just unblocked being able to run a number of tests locally on ORM 3.0 for now).

### DIFF
--- a/src/Mapping/MappedEventSubscriber.php
+++ b/src/Mapping/MappedEventSubscriber.php
@@ -17,7 +17,8 @@ use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as DocumentClassMetadata;
 use Doctrine\ORM\EntityManagerInterface;
-use Doctrine\ORM\Mapping\ClassMetadataInfo as EntityClassMetadata;
+use Doctrine\ORM\Mapping\ClassMetadata as EntityClassMetadata;
+use Doctrine\ORM\Mapping\ClassMetadataInfo as LegacyEntityClassMetadata;
 use Doctrine\Persistence\Mapping\AbstractClassMetadataFactory;
 use Doctrine\Persistence\Mapping\ClassMetadata;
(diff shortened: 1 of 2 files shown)


franmomu (approved):
I guess we are keeping `ClassMetadataInfo` in case someone is extending it (unlikely, but 🤷‍♂️ ).

src/Mapping/MappedEventSubscriber.php
```diff
@@ -229,7 +230,7 @@ final public function setCacheItemPool(CacheItemPoolInterface $cacheItemPool): v
      */
     public function loadMetadataForObjectClass(ObjectManager $objectManager, $metadata)
     {
-        assert($metadata instanceof DocumentClassMetadata || $metadata instanceof EntityClassMetadata);
+        assert($metadata instanceof DocumentClassMetadata || $metadata instanceof EntityClassMetadata || $metadata instanceof LegacyEntityClassMetadata);
```
phansys, line 233: Maybe we could add the dockblock `@phpstan-assert ClassMetadata&(DocumentClassMetadata|EntityClassMetadata|LegacyEntityClassMetadata) $meta`.

mbabker: I tried it, that assertion didn't change anything on its own.  Removing the `assert()` runtime call added errors, or trying `@param ClassMetadata&(DocumentClassMetadata|EntityClassMetadata|LegacyEntityClassMetadata) $metadata` like is done elsewhere added more errors.

franmomu:
thanks @mbabker!

PR #2718, opened 2023-10-30
Use real event for test

franmomu:
Ref: https://github.com/doctrine-extensions/DoctrineExtensions/pull/2639#discussion_r1320742594

### DIFF
--- a/tests/Gedmo/Mapping/MappingEventAdapterTest.php
+++ b/tests/Gedmo/Mapping/MappingEventAdapterTest.php
@@ -11,8 +11,8 @@ declare(strict_types=1);
 
 namespace Gedmo\Tests\Mapping;
 
-use Doctrine\ORM\EntityManager;
-use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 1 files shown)


tests/Gedmo/Mapping/MappingEventAdapterTest.php
```diff
@@ -49,19 +44,14 @@ public function testCorrectAdapter(): void
...
-            ->willReturn(new \stdClass());
+        $args = new PrePersistEventArgs($entity, $emMock);
 
         $eventAdapter = new EventAdapterORM();
-        $eventAdapter->setEventArgs($eventArgsMock);
-        $eventAdapter->getObjectManager();
-        $eventAdapter->getObject();
+        $eventAdapter->setEventArgs($args);
+        static::assertSame($eventAdapter->getObjectManager(), $emMock);
+        static::assertInstanceOf(\stdClass::class, $eventAdapter->getObject());
```
franmomu, line 55: Since we are testing behavior, instead of expecting that the `EventArgs::getObjectManager()` and `EventArgs::getObject()` are called, we can check with the result of these calls.

phansys (approved):
Thank you @franmomu!

PR #2719, opened 2023-10-30
Removed explicit default options for PHPStan 6

franmomu:
These options are in https://github.com/phpstan/phpstan-src/blob/a849f06ff5195c95d0c03ef14dffc65986fe92a2/conf/config.level6.neon by default.

I don't know why it shows now less errors 🤷‍♂️


phansys (approved):

phansys:
Thanks @franmomu!

PR #2720, opened 2023-10-31
Drop support of `doctrine/dbal` 2

franmomu:
[SQLLogger was deprecated in `doctrine/dbal` 3.2](https://github.com/doctrine/dbal/pull/4967) in favor of middlewares.

We starting using a middleware for tests in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2397 which was a mock of `LoggerInterface`.

This PR creates a `QueryLogger` class for tests which stores the executed queries and is used to replace the use of the deprecated `DebugStack` (when using dbal > 3.2) and also we can remove the calls to the deprecated `InvocationMocker:: withConsecutive()` method.

Middlewares are configured when creating the connection, that's why we need to call `QueryLogger::reset()` every time before executing any query we may be interested in, so we can inspect only the queries we want.

PS: Maybe it's time to drop support of doctrine/dbal 2?


mbabker:
> PS: Maybe it's time to drop support of doctrine/dbal 2?

I'd say do it.

Over on the Symfony side of things, DoctrineBundle requires DBAL 3.6 now, and Symfony 7 drops DBAL 2 support, so it's effectively already gone if you're using recent versions of DoctrineBundle.

Other Doctrine libraries like Migrations have also dropped DBAL 2 support.

`doctrine/doctrine-orm-module` for Laminas still supports DBAL 2.13.  Looking over the dependents list on Packagist, a couple other "big" dependents include TYPO3 and `laravel-doctrine/orm`, both of which require DBAL 3.

franmomu:
I've added a commit removing support for dbal 2 in case we want to remove it

PR #2722, opened 2023-11-05
Remove custom vendor bin directory

franmomu:
I'm not if there was a reason why it was added in https://github.com/doctrine-extensions/DoctrineExtensions/commit/4e9bbd54ebb22bb004b2ec71b1b0ce3ed5b992d0, but this simplifies (using the common `vendor/bin` path) executing commands locally.

PS: We could also move `phpunit.xml.dist` from `tests` to the root directory.


phansys (approved):

phansys:
Thank you @franmomu!

PR #2724, opened 2023-11-19
chore: add testruns against php 8.3

Chris53897:
there are multiple possible solutions. I am not sure which one is the desired one.


franmomu (approved):

franmomu:
thanks @Chris53897!

PR #2725, opened 2023-11-20
Use custom lifecycle event classes for the soft-deleteable lifecycle events

mbabker:
Ref: #2708

In ORM 3.0, the `Doctrine\ORM\Event\LifecycleEventArgs` class is removed and the ORM's lifecycle event classes directly extend `Doctrine\Persistence\Event\LifecycleEventArgs`.  This makes the ORM's implementation of `Gedmo\Mapping\Event\AdapterInterface::createLifecycleEventArgsInstance()` unusable on 3.0 since it no longer has its own lifecycle event args subclass.

Currently, the only use of this method inside the package is for the soft-deleteable extension's events.

My solution to this ORM deprecation comes with a minor B/C break; instead of dispatching the events with manager-specific event classes, the events are now dispatched with custom subclasses of `Doctrine\Persistence\Event\LifecycleEventArgs` for both supported managers.  The B/C break here is that event listeners which typehint against `Doctrine\ODM\MongoDB\Event\LifecycleEventArgs` and `Doctrine\ORM\Event\LifecycleEventArgs` will be broken because the class inheritance chain is changed.  This would also deprecate the implementations of `Gedmo\Mapping\Event\AdapterInterface::createLifecycleEventArgsInstance()` (currently soft-required through `@method` annotations).

There is an alternative solution that only has a B/C break for ORM users; changing this method to always use the `Doctrine\Persistence\Event\LifecycleEventArgs` class while the ODM adapter still uses its manager-specific subclass.

But because of the ORM 3.0 class removal, there will have to be a B/C break at some point, and I figured this would be the best alternative going forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,14 @@ a release.
 - Dropped support for PHP < 7.4
 - Dropped support for Symfony < 5.4
 - Dropped support for doctrine/dbal < 3.2
+- ⚠️ [B/C Break] The `Gedmo\SoftDeleteable\SoftDeletableListener` dispatches events using the new `Gedmo\SoftDeleteable\Event\PreSoftDeleteEventArgs`
+  and `Gedmo\SoftDeleteable\Event\PostSoftDeleteEventArgs` classes, which directly extend from `Doctrine\Persistence\Event\LifecycleEventArgs`; previously,
+  each object manager's own lifecycle event args class was used.
 
 ### Deprecated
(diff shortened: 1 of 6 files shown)


franmomu:
There is also https://github.com/doctrine-extensions/DoctrineExtensions/pull/2649 that tries to avoid the BC break (I have to take a look again at it)

mbabker:
I think the one thing I don't like about #2649 is that it hides the ORM B/C break (and inherently leaves an unfixable deprecation in the package and downstream apps running on ORM 2) by using conditional inheritance, and it'll still silently change the class inheritance when one upgrades from ORM 2 to 3.  And, IMO, there really isn't a need for separate classes per manager for these events.  So this implementation is simpler but has more B/C implications, the other PR has less downstream B/C concern but adds technical debt.

mbabker:
Close in favor of #2649

Issue #2683, opened 2023-08-23
The dependency on doctrine/annotations should become optional


stof:
## Feature Request

For projects that have migrated from annotations to attributes (as recommended by the Doctrine team), the `doctrine/annotations` package is unnecessary.
As the Doctrine team is planning to mark the `doctrine/annotations` package as abandoned (see https://github.com/doctrine/annotations/issues/485), it would be great if this package could stop being a mandatory dependency of the extensions.

mbabker:
See #2554 as it has a more complete rundown of where the issues lie.

stof:
since this issue has been created, support for attributes has been implemented.
So for PHP 8+, attributes could be used to solve the issue:
- loggable and translatable provides mapping both using annotations and attributes now
- the helper methods could rely on those existing attributes to register an attribute driver instead of an annotations one
- ExtensionMetadataFactory could be creating a default attribute reader instead.

To me, the main blocker is actually the support for PHP 7.x. Maybe it is time to bump the min PHP version to 8.0 (or maybe 8.1 to be able to use enums in #2675 instead).

Note also that doctrine/orm 3.0 is dropping the annotation driver entirely, so you will have to do this migration to attributes before supporting ORM 3.0 (or making a conditional migration)

mbabker:
#2559 deals with the helpers.

As you pointed out, attributes exist now, and leaving the annotations in place won't hurt anything for the time being, so that one's adequately covered for now.

I think realistically the remaining hard dependencies are the non-nullable `$annotationReader` argument of the `ExtensionMetadataFactory` constructor (including the logic in `MappedEventSubscriber::getExtensionMetadataFactory()` which tries to create a default annotation reader to provide the factory if one doesn't exist) and inherently changing its `getDriver()` method to throw if it tries to use an annotation or attribute based driver without that dependency being passed in (as the way the method reads to me right now, if you're using a XML or YAML mapping driver, you should never reach the code paths where the reader would be required, so that's already a bit of an over-reaching constructor requirement).

tacman:
Any update on this?   I've removed doctrine/annotations from a project, so of course this fails.  

```yaml
services:
  gedmo.listener.timestampable:
    class: Gedmo\Timestampable\TimestampableListener
    tags:
      - { name: doctrine.event_subscriber, connection: default }
    calls:
      - [ setAnnotationReader, [ '@annotation_reader' ] ]
```

mbabker:
[...]

PR #2728, opened 2023-11-28
Make the annotation/attribute reader optional in the extension metadata factory

mbabker:
Continuing the effort toward completing #2683, this will make the annotation reader optional in `Gedmo\Mapping\ExtensionMetadataFactory`.

As this factory only needs the annotation reader when creating an annotation or attribute driver, this means it should not be a hard requirement when working with other driver mappings (i.e. XML or deprecated YAML support).  So, the factory's constructor now allows creating it without the reader and checks when it's creating drivers if the reader is needed.

The factory's `getDriver()` method also had a fallback to a default annotation driver if an extension didn't provide a driver in another format.  With that being hard coupled to `doctrine/annotations`, a similar approach to what is proposed in #2683 is used where the fallback will go to the annotation driver if `doctrine/annotations` is installed, and now it will try falling back to the attribute driver if `doctrine/annotations` is not installed and the app is running on PHP 8.

To make this work, the private state in the base `Gedmo\Mapping\MappedEventSubscriber` class now initializes the annotation reader properties to false, and either through setter injection or the internal `getDefaultAnnotationReader()` those will flip to either a reader instance or null.  The setter isn't documented to support the new internal false condition, and everything else is private, so we should be good on B/C here.

In theory, after this PR, `doctrine/annotations` could become a `require-dev` dependency as I think we've now got everything in the production code with a hard Annotations package dependency reviewed, but there's a long way to go before we could even consider a test run for this package with it fully uninstalled since so many of the tests absolutely require either annotation or YAML support.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -26,6 +26,7 @@ return static function (RectorConfig $rectorConfig): void {
 
     $rectorConfig->skip([
         TypedPropertyFromAssignsRector::class => [
+            __DIR__.'/src/Mapping/MappedEventSubscriber.php', // Rector is trying to set a type on the $annotationReader property which requires a union type, not supported on PHP 7.4
             __DIR__.'/tests/Gedmo/Wrapper/Fixture/Entity/CompositeRelation.php', // @todo: remove this when https://github.com/doctrine/orm/issues/8255 is solved
         ],
     ]);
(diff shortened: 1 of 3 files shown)


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2732, opened 2023-12-04
Refactor some mapping tests for improved coverage and forward compat

mbabker:
Ref: #2708

Right now there are a lot of tests that rely on the deprecated annotation and YAML drivers from the ORM to actually test anything.  This means in a lot of cases the XML mapping isn't explicitly tested in a mapping case (it is incidentally covered elsewhere), nor can the tests run without the deprecated Annotations package.  This is a first step at improving some of the test cases to be able to cover XML and explicitly test attribute mapping (and while all the extension-level attribute drivers are just an extension of their respective annotation drivers, there should be coverage when working purely in the context of attributes and not using the annotations package at all).

This PR refactors the loggable, sluggable, and soft-deleteable mapping tests to conditionally test mapping drivers based on whether the underlying dependencies are available so that the tests automatically skip themselves when needed while also supporting covering all mapping options.  I'm mostly building on and re-using existing test fixtures where they exist, and adding attributes where they don't exist on annotated classes, so some stuff might not exist at all right now (i.e. there isn't an XML mapping or fixture class for the `Category` models tested with other drivers).  Following this pattern for at least the mapping tests, this will help ensure explicit coverage on all drivers while only testing drivers that the environment supports.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -571,6 +571,11 @@ parameters:
 			path: src/Uploadable/UploadableListener.php
 
 		-
+			message: "#^Class Gedmo\\\\Tests\\\\Translatable\\\\Fixture\\\\CategoryTranslation not found\\.$#"
+			count: 1
+			path: tests/Gedmo/Mapping/Fixture/Category.php
+
+		-
 			message: "#^Call to an undefined method Doctrine\\\\Persistence\\\\Mapping\\\\ClassMetadata\\<T of object\\>\\:\\:mapField\\(\\)\\.$#"
 			count: 2
 			path: tests/Gedmo/Mapping/MetadataFactory/CustomDriverTest.php
--- a/tests/Gedmo/Mapping/Fixture/Loggable.php
+++ b/tests/Gedmo/Mapping/Fixture/Loggable.php
@@ -9,6 +9,7 @@
 
 namespace Gedmo\Tests\Mapping\Fixture;
 
+use Doctrine\DBAL\Types\Types;
 use Doctrine\ORM\Mapping as ORM;
 use Gedmo\Mapping\Annotation as Gedmo;
 
@@ -17,6 +18,8 @@ use Gedmo\Mapping\Annotation as Gedmo;
  *
  * @Gedmo\Loggable
  */
+#[ORM\Entity]
+#[Gedmo\Loggable]
 class Loggable
 {
     /**
@@ -26,6 +29,9 @@ class Loggable
      * @ORM\GeneratedValue
      * @ORM\Column(type="integer")
      */
+    #[ORM\Id]
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: Types::INTEGER)]
     private $id;
 
     /**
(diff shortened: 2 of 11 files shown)


phpstan-baseline.neon
```diff
@@ -570,6 +570,11 @@ parameters:
 			count: 3
 			path: src/Uploadable/UploadableListener.php
 
+		-
+			message: "#^Class Gedmo\\\\Tests\\\\Translatable\\\\Fixture\\\\CategoryTranslation not found\\.$#"
```
mbabker, line 574: Baselining this because the YAML mapping that the annotation/attribute mappings were generated from references this non-existing class, and PHPStan now (rightfully) groans about it.  The test fixtures should probably use an existing class at some point in the future, but for now, its absence doesn't break anything.

tests/Gedmo/Mapping/Fixture/Loggable.php
```diff
@@ -26,13 +29,17 @@ class Loggable
...
+    #[ORM\GeneratedValue]
+    #[ORM\Column(type: Types::INTEGER)]
     private $id;
 
     /**
      * @ORM\Column(name="title", type="string", length=64)
      *
      * @Gedmo\Versioned
      */
+    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
```
franmomu, line 42:
```suggestion
    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
    #[Gedmo\Versioned]
```

tests/Gedmo/Mapping/Fixture/LoggableComposite.php
```diff
@@ -33,13 +38,16 @@ class LoggableComposite
...
+    #[ORM\Id]
+    #[ORM\Column(type: Types::INTEGER)]
     private $two;
 
     /**
      * @ORM\Column(name="title", type="string", length=64)
      *
      * @Gedmo\Versioned
      */
+    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
```
franmomu, line 50:
```suggestion
    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
    #[Gedmo\Versioned]
```

tests/Gedmo/Mapping/Fixture/LoggableCompositeRelation.php
```diff
@@ -33,13 +38,16 @@ class LoggableCompositeRelation
...
+    #[ORM\Id]
+    #[ORM\Column(type: Types::INTEGER)]
     private $two;
 
     /**
      * @ORM\Column(name="title", type="string", length=64)
      *
      * @Gedmo\Versioned
      */
+    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
```
franmomu, line 50:
```suggestion
    #[ORM\Column(name: 'title', type: Types::STRING, length: 64)]
    #[Gedmo\Versioned]
```

tests/Gedmo/Mapping/Fixture/Loggable.php
mbabker: Fixed while rebasing.

franmomu (approved):

franmomu:
thanks @mbabker!

PR #2734, opened 2023-12-08
Another batch of mapping tests refactored

mbabker:
Ref: #2708

Same logic as #2732 for this batch.  Most of the changes in existing fixtures are meant to try and get them to have similar mapping definitions across mapping providers, which hopefully makes maintaining the tests simpler in the long run as one case is only needed for all 4 mapping configurations.  Unfortunately, some quirks and differences in the mapping drivers themselves make this tricky in some places (and consolidating these mapping tests is starting to show that, like the loggable embeddable configs); those should be revisited at some point down the road.

The deleted files are all mapping test classes that are now merged into a single test class.


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2735, opened 2023-12-08
Make the annotations package optional

### LINKED ISSUE #2683: The dependency on doctrine/annotations should become optional

mbabker:
Fix #2683

With all of the known dependencies to the Annotations package reviewed and adjusted, it should now be practical to make it an optional package.  And that's what this PR accomplishes.

It'd be hugely beneficial if folks would test this in downstream applications, especially if this package is the only reason `doctrine/annotations` gets installed in their application, to make sure the features they are using are not broken (I've done it in an app using the ORM plus the loggable and sortable extensions, but that's a very small amount of test coverage).  To apply this PR, update your `composer.json` as appropriate with this info:

```json
{
  "repositories": [
    {
      "type": "vcs",
      "url": "https://github.com/mbabker/DoctrineExtensions"
    }
  ],
  "require": {
    "gedmo/doctrine-extensions": "dev-optional-annotations as 3.15.0",
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ a release.
 - SoftDeleteable: `Gedmo\SoftDeleteable\Event\PreSoftDeleteEventArgs` and
   `Gedmo\SoftDeleteable\Event\PostSoftDeleteEventArgs` classes.
 
+### Changed
+- Make doctrine/annotations an optional dependency.
+
 ### Deprecated
 - Do not add type-hinted parameters `Gedmo\SoftDeleteable\Event\PreSoftDeleteEventArgs` and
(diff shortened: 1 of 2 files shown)


franmomu (approved):

franmomu:
Sorry for the late reply, let's merge it since it will be easier to test, thanks for the effort you are putting into this @mbabker!

PR #2738, opened 2023-12-14
Another round of mapping refactoring in tests for cross-version compatibility

mbabker:
Ref: #2708

We don't need to check for class presence on the ORM's attribute driver, we're well past the version that was introduced as a minimum requirement.  Other than that, just some more fixture generation/synchronization and adjusting test setup to always use attributes on PHP 8 instead of hard requiring the annotation driver.


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2743, opened 2023-12-19
Remove the `@internal` annotation from the attribute reader

mbabker:
`Gedmo\Mapping\Driver\AttributeReader` isn't really internal.  If you're running an application that doesn't actually use annotations (but the annotations library still gets installed because of dependencies like this one where there's still a hard coupling), the only way to actually configure this library to not use annotations is to use this `@internal` annotated class.  Which is also typehinted in a fairly decent number of public APIs.  So, let's just remove the annotation.


franmomu (approved):

franmomu:
thanks @mbabker! It wasn't easy to remove, so when the times comes we can deprecate those methods.

PR #2754, opened 2024-02-05
Standardize runtime deprecation handling to the `doctrine/deprecations` library

mbabker:
I've added some deprecations using Symfony's `trigger_deprecation()` function, others have done it with straight-up `@trigger_error()` calls.  This PR makes the runtime handling consistent and uses Doctrine's deprecations library to do so.

Why Doctrine's package?  It's a bit more feature-rich than just using one of the `trigger_*()` function calls with inbuilt support for a logger or ignoring select deprecations directly through its API compared to needing to hook into PHP's own error handling mechanism to deal with `E_USER_DEPRECATED` errors, and for package maintainers, it gives a way to selectively emit the runtime deprecation through its `Deprecation::triggerIfCalledFromOutside()` to avoid spamming unfixable deprecations from package internal code paths as well as providing its own PHPUnit integration for verify whether deprecations are emitted.

Plus, in recent Symfony versions, Doctrine's deprecation library is enabled by default in the FrameworkBundle and PHPUnit Bridge, so the tooling is already treated as a first-class citizen there.

### DIFF
--- a/src/Sortable/SortableListener.php
+++ b/src/Sortable/SortableListener.php
@@ -12,6 +12,7 @@ namespace Gedmo\Sortable;
 use Doctrine\Common\Comparable;
 use Doctrine\Common\EventArgs;
 use Doctrine\Common\Util\ClassUtils;
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Persistence\ObjectManager;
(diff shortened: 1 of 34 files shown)


src/Sortable/SortableListener.php
```diff
@@ -157,7 +158,9 @@ public function onFlush(EventArgs $args)
     /**
      * Update maxPositions as needed
      *
-     * @param LifecycleEventArgs $args
+     * @param 
+     
+     $args
```
franmomu, line 163:
```suggestion
     * @param LifecycleEventArgs $args
```

franmomu (approved):

franmomu:
thanks @mbabker!

PR #2757, opened 2024-02-08
Update TimestampableEntity.php for compat with ORM v3

andreybolonin:
https://github.com/doctrine-extensions/DoctrineExtensions/issues/2708#issuecomment-1931650181

### DIFF
--- a/src/Timestampable/Traits/TimestampableEntity.php
+++ b/src/Timestampable/Traits/TimestampableEntity.php
@@ -24,8 +24,6 @@ trait TimestampableEntity
      * @var \DateTime|null
      *
      * @Gedmo\Timestampable(on="create")
-     *
-     * @ORM\Column(type="datetime")
      */
     #[Gedmo\Timestampable(on: 'create')]
     #[ORM\Column(type: Types::DATETIME_MUTABLE)]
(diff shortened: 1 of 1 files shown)


mbabker:
This is an incorrect fix as it removes annotation support entirely.  The presence of annotations when installed alongside ORM 3.0 in and of itself isn't an issue, unless, your application is still using annotations for something outside the ORM and this package's mapping and those annotations are being read by that process.  In that case, you should add a call to `Doctrine\Common\Annotations\AnnotationReader::addGlobalIgnoredNamespace('Doctrine\ORM\Mapping')` so the annotation reader service in your application ignores the ORM's mapping classes.

As well, this package does not yet support ORM 3.0.  You should downgrade your application to ORM 2.18 and continue using that until all dependencies are compatible with the new major release.

mishac:
In that case shouldn't the composer.json indicate the conflict with ORM > 2?

mbabker:
> In that case shouldn't the composer.json indicate the conflict with ORM > 2?

Probably should've been done sooner, but #2759

franmomu:
Closing as this is not right as explained in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2757#issuecomment-1934260512

PR #2758, opened 2024-02-08
Add a compat trait to address B/C issues with EntityRepository

mbabker:
Ref: #2708

In ORM 3.0, the `EntityRepository::__call()` method signature is fully typehinted.  As it includes a `mixed` return type, that signature can't just be applied to the package right now due to PHP 7.4 still being supported.

To help with B/C, an internal trait is introduced that any entity repository provided by this package can use (thus far, it only affects the `NestedTreeRepository`, but if we find other gaps, the trait can be updated to account for those as well.


franmomu:
Thanks @mbabker! I guess we should use this strategy with more classes to deal with BC with ORM 3.

PR #2759, opened 2024-02-09
Add high versions to optional package conflicts

mbabker:
Probably should've been done much sooner, but this will add upper-version constraints to the conflicts for the DBAL and object manager packages.


franmomu (approved):

franmomu:
thanks @mbabker!

Issue #2700, opened 2023-09-20
Clock Mocking doesn't work from v3.12.0


rasstislav:
Hello,

[this commit](https://github.com/doctrine-extensions/DoctrineExtensions/commit/2ed66e372073369418e4c8457725b5d7cc54b30c) broke [Clock mocking](https://symfony.com/doc/current/components/phpunit_bridge.html#clock-mocking).

Is it possible to fix it?

Thanks

stof:
The future-proof way would be to support injecting a PSR-20 clock in the listener and using it to get the current time.

Chris53897:
@rasstislav Can this be closed as completed (see linked PR)?

rasstislav:
@Chris53897 my tests are still not working due to DateTime as @boedah also mentioned.

rasstislav:
@boedah, can you please test it as well?

mbabker:
> @Chris53897 my tests are still not working due to DateTime as @boedah also mentioned.

What tests specifically?  Without a specific use case, it's hard to know what it is that's causing your issue.  The last release supports injecting a PSR-20 where practical, with the only places using still using `new \DateTime()` being an ODM query filter where injecting a clock isn't really feasible (or if it is I guess that'd be through the parameters API, but not being an ODM user I couldn't tell you if that's good or bad), and the `setLoggedAt()` methods on the base log entry document/entity classes that it's recommended to extend from for that integration, and you can work around that with implementing your own log entry object and overloading those methods to use some other time source if need be.

phansys:
Closing as #2760 implemented a solution for this.
Thanks!

PR #2760, opened 2024-02-09
Add support for providing a PSR-20 clock to event adapters which create new DateTimeInterface objects

mbabker:
Refs: #2700, stof/StofDoctrineExtensionsBundle#469

This PR adds support for injecting a PSR-20 clock into event listeners and their event adapters to replace hardcoded `new \DateTime()` calls.  Similar to supporting injection of an annotation reader or cache pool, a setter is provided to inject the clock.  Then, when creating adapters via `MappedEventSubscriber::getEventAdapter()`, if the adapter implements the new `Gedmo\Mapping\Event\ClockAwareAdapterInterface`, this factory will inject the clock to the adapter if one was provided.

Basic example:

```php
$listener = new TimestampableListener();
$listener->setClock(new Clock());

// All the rest of the usual setup logic...
```

Since the PSR-20 implementations I looked at all require PHP 8, for testing purposes, a very simple test mock is added to the test suite and used in a couple of the test cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ a release.
 ### Added
 - SoftDeleteable: `Gedmo\SoftDeleteable\Event\PreSoftDeleteEventArgs` and
   `Gedmo\SoftDeleteable\Event\PostSoftDeleteEventArgs` classes.
+- Add support for injecting a `psr/clock` implementation into event adapters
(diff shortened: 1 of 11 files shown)


franmomu (approved):

mbabker:
> LGTM, can you please add an entry in the changelog?

Done!

franmomu:
thanks @mbabker! I think I'll make a release since ORM 3 is stable and we are starting to receive issues related to this.

PR #2761, opened 2024-02-09
Document expected EventArgs subtypes in listeners

mbabker:
The event listeners really aren't expected to receive the base `Doctrine\Common\EventArgs` object, it misses information that almost every listener needs (all of them expect an object manager, and many also are lifecycle event listeners expecting an object).  Since changing the type declarations in the signatures would be a B/C break, this PR adds the expected subtype through `@param` annotations.  Bouncing this off the ORM and MongoDB ODM, it looks like the intent is that all of the event sub-classes end up being a child of either `Doctrine\Persistence\Event\LifecycleEventArgs` or `Doctrine\Persistence\Event\ManagerEventArgs` so that is the information provided.

There's still a lot of missing type safety or sanity checks (i.e. always assuming `$event->getObjectManager()->getUnitOfWork()` will exist since that's not a part of the root object manager interface, but it's there for the managers that are supported here), but it's a small step in the right direction.

### DIFF
--- a/src/Uploadable/UploadableListener.php
+++ b/src/Uploadable/UploadableListener.php
@@ -12,7 +12,9 @@ namespace Gedmo\Uploadable;
 use Doctrine\Common\EventArgs;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
 use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
+use Doctrine\Persistence\Event\ManagerEventArgs;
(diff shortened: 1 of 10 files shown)


src/Uploadable/UploadableListener.php
```diff
@@ -12,7 +12,9 @@
 use Doctrine\Common\EventArgs;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\Persistence\Event\LifecycleEventArgs;
```
franmomu, line 15:
```suggestion
```

franmomu (approved):

franmomu:
thanks @mbabker!

PR #2765, opened 2024-02-12
Doctrine\ORM\Mapping\ClassMetadata::getFieldMapping() will return an object on ORM 3 instead of an array

mbabker:
Ref: #2708

Cherry-picks out changes related to moving from an array to the `Doctrine\ORM\Mapping\FieldMapping` in ORM 3.0.

The compat issues are baselined into the PHPStan config for now since we know the class doesn't exist on 2.18 (which is what SA is running against) and the entire point of this diff is to be a compat layer.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -291,6 +291,21 @@ parameters:
 			path: src/Sluggable/SluggableListener.php
 
 		-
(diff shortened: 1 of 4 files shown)


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2768, opened 2024-02-14
Change the placeholder for the sluggable extension when generating a slug for an identifier field

mbabker:
Ref: #2708

In https://github.com/doctrine/orm/pull/10785 (shipped in ORM 2.16.0) a check was added to the UnitOfWork to prevent ID collisions in its internal map, which was then changed to a deprecation in https://github.com/doctrine/orm/pull/10878 (shipped in ORM 2.16.1) due to its disruptiveness.  That check is back in 3.0.

This PR avoids the issue by changing the ID placeholder used by the sluggable extension when it is configured to generate slugs for an identifier field to a placeholder based on `uniqid()` and fixes the last PHPUnit error on my hack-it-together-ORM-3-support branch.

### DIFF
--- a/src/Sluggable/SluggableListener.php
+++ b/src/Sluggable/SluggableListener.php
@@ -257,7 +257,7 @@ class SluggableListener extends MappedEventSubscriber
         if ($config = $this->getConfiguration($om, $meta->getName())) {
             foreach ($config['slugs'] as $slugField => $options) {
                 if ($meta->isIdentifier($slugField)) {
-                    $meta->getReflectionProperty($slugField)->setValue($object, '__id__');
+                    $meta->getReflectionProperty($slugField)->setValue($object, uniqid('__sluggable_placeholder__'));
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Sluggable/SluggableListener.php
```diff
@@ -257,7 +257,7 @@ public function prePersist(EventArgs $args)
         if ($config = $this->getConfiguration($om, $meta->getName())) {
             foreach ($config['slugs'] as $slugField => $options) {
                 if ($meta->isIdentifier($slugField)) {
-                    $meta->getReflectionProperty($slugField)->setValue($object, '__id__');
+                    $meta->getReflectionProperty($slugField)->setValue($object, '__sluggable_placeholder_'.random_int(1, PHP_INT_MAX).'__');
```
Lustmored, line 260: How about using `uniqid()` instead of `random_int()`? I believe the former is created for exact this purpose - there needs to be uniqueness, but not necessarily cryptographic security. Chances of collision with `random_int` are practically 0, but still a change to consider.

The call would probably also get a bit cleaner: `uniqid('__sluggable_placeholder_')` (`'__'` at the end is unused, so not necessary)

mbabker: Yeah, that'd be fine.  Admittedly I went with the first solution that came to mind, which was `random_int()`, but that is a bit overkill here since it doesn't need cryptographic-backed uniqueness.

franmomu (approved):

franmomu:
Thanks @mbabker!

PR #2772, opened 2024-02-19
Deprecate annotations support

mbabker:
Now that the hard dependency is broken, the next major task to coincide with [the `doctrine/annotations` library being deprecated](https://github.com/doctrine/annotations/issues/485) would be deprecating annotations support in this package.  WIP for now because this is going to be quite a big deprecation and architectural task since the attribute drivers are all subclasses of their annotation counterparts so some thought will be needed into how to address that (as folks in the Symfony ecosystem will eventually come here with reports about it, even with the drivers all being flagged internal).

Other things to think about along the way include:

- In Symfony, the term "annotation" was deprecated in full and replaced with "attribute" in all relevant APIs, which included renaming and aliasing classes in "Annotation" namespaces and renaming methods (i.e. `setAnnotationReader()`), how in-depth does this deprecation need to go here?


franmomu:
I don't know if we can invert the inheritance and make the Annotation classes inherit from the Attribute ones

mbabker:
> I don't know if we can invert the inheritance and make the Annotation classes inherit from the Attribute ones

We'd have to flip it for the `AnnotationDriverInterface` and `AttributeDriverInterface` as well.  There are checks like [this one](https://github.com/doctrine-extensions/DoctrineExtensions/blob/8d658b4d22977e3b72f02bfe4e68b2df0ba586aa/src/Mapping/ExtensionMetadataFactory.php#L243-L257) targeting only the attribute drivers so those bits and bobs would need to be inverted as well to not break any other behaviors.

Conceivably it's possible, and the mapping drivers within this package are a lot more of an internal implementation than in the object managers themselves given the way `ExtensionMetadataFactory::getDriver()` works, so even though there are some potential B/C implications they aren't as impactful to downstream users as the changes that were made upstream.

mbabker:
Reversing the inheritance chain actually worked out a lot more smoothly than I expected it to.  The only possible hangup from an end-user perspective is if someone implemented a custom extension (as in fully added their own listener extending from `MappedEventSubscriber` and adding their own annotation/attribute mapping drivers), they would also have to flip their inheritance chains.  I would suggest the likelihood of that is pretty low, though.

franmomu:
> Reversing the inheritance chain actually worked out a lot more smoothly than I expected it to. 

nice 🎉 

> The only possible hangup from an end-user perspective is if someone implemented a custom extension (as in fully added their own listener extending from `MappedEventSubscriber` and adding their own annotation/attribute mapping drivers), they would also have to flip their inheritance chains. I would suggest the likelihood of that is pretty low, though.

I'm fine with this, I don't think a lot of people would have both annotations and attributes drivers in their custom extensions.

franmomu (approved):
LGTM

phansys (approved):

PR #2773, opened 2024-02-19
Misc. code doc updates

mbabker:
Just as advertised:

- Updates the trait doc blocks to remove references to PHP 5.4, we're a bit past that point 😉
- Updates the marker interface doc blocks to make it a bit more clear these are generic marker interfaces
- Updates the inline comments on the marker interfaces to stop using a very legacy annotation syntax


franmomu (approved):

franmomu:
thanks @mbabker, nice work!

PR #2089, opened 2020-01-27
Fix ODM SoftDeleteableFilter return type is missing fix Atlantic18/Do…

theyough:
…ctrineExtensions#2088

PR #2782, opened 2024-03-19
[SoftDeleteableFilter] Fix addFilterConstraint() return type is missing

cavasinf:
Resolve:
> Fatal error: Declaration of Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, $targetTableAlias) must be compatible with Doctrine\ORM\Query\Filter\SQLFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, string $targetTableAlias): string in /var/www/html/vendor/gedmo/doctrine-extensions/src/SoftDeleteable/Filter/SoftDeleteableFilter.php on line 54

https://github.com/doctrine/orm/blob/55c4845d5734bea5579424c2c2a7c5b97dbb4c60/src/Query/Filter/SQLFilter.php#L173

As DoctrineExtensions supports `php >= 7`, but was denied in #2089 before.

### DIFF
--- a/src/SoftDeleteable/Filter/SoftDeleteableFilter.php
+++ b/src/SoftDeleteable/Filter/SoftDeleteableFilter.php
@@ -51,7 +51,7 @@ class SoftDeleteableFilter extends SQLFilter
      *
      * @return string
      */
-    public function addFilterConstraint(ClassMetadata $targetEntity, $targetTableAlias)
+    public function addFilterConstraint(ClassMetadata $targetEntity, $targetTableAlias): string
     {
         $class = $targetEntity->getName();
         if (true === ($this->disabled[$class] ?? false)) {


mbabker:
It looks like you're trying to use this package with ORM 3.0 given the presence of the return type.  This package itself [isn't compatible with ORM 3.0](https://github.com/doctrine-extensions/DoctrineExtensions/issues/2708) yet, and technically adding this return type (or one of the other dozen or so now mandatory param or return types [I've identified working on ORM 3.0 compat](https://github.com/doctrine-extensions/DoctrineExtensions/compare/main...mbabker:DoctrineExtensions:orm-3-merge)) is a B/C break within this package.  So this PR on its own can't just be merged and shipped.

cavasinf:
OK, I though ORM 3.0 wasn't compatible because of annotation vs attributes left (#2708).
Closing this PR then.

mbabker:
> I though ORM 3.0 wasn't compatible because of annotation vs attributes left

That part should be fine as long as you use explicit config.  The feature detection bits in this package prefer annotations over attributes for B/C, so when a reader is not explicitly injected and your app environment has `doctrine/annotations` installed (either as your own explicit app dependency or it's still being pulled in as a transient dependency through one of your other dependencies), the package will create a new `Doctrine\Common\Annotations\AnnotationReader` instance without any configuration, and that would trigger annotation parsing errors [like this one](https://github.com/doctrine-extensions/DoctrineExtensions/issues/2708#issuecomment-1992111139).  So to avoid that if your app is using attributes, you can inject the `Gedmo\Mapping\Driver\AttributeReader` in and avoid the feature detection trying to decide what to use.

PR #2786, opened 2024-04-04
Add support for ORM3

ajgarlag:
Based on the incredible work of @oleg-andreyev on #2784, this PR adds support for ORM >= 3.0 keeping compatibility with ORM 2 and without removing Annotations or YAML support.


mbabker:
I need to rebuild the branch, but https://github.com/doctrine-extensions/DoctrineExtensions/compare/main...mbabker:DoctrineExtensions:orm-3-merge plus some version of https://github.com/doctrine-extensions/DoctrineExtensions/pull/2768 should be pretty close to what all is needed (I know the `MappedEventSubscriber.php` change should go away for a final PR, but I've been working through that branch for so long I don't remember if I threw any other convenience hacks at it), minus factoring for some PHP 7.4 compat breaks (union types are used in ORM 3).

The custom ORM XML mapping driver should be 100% unnecessary, they reverted the hard mapping requirements for that driver because of the inability to use a custom XSD making extensions with XML config impossible (follow https://github.com/doctrine/orm/issues/11123 for that, and the MongoDB ODM folks have an eye on that too since they have a similar issue with their XML mapping).

Seb33300:
I am using that branch in my project and it's working so far. (but I only use Blameable & Timestampable)
The only point I noticed is that compatibility with DBAL 4 is missing.

It could be nice to start merging PR related to ORM3 even if they're not 100% completed.
This will make it easy for everybody to start testing and report broken features.
For instance by creating a BETA version release.

franmomu:
`main` branch is now compatible with ORM 2 and 3 with the changes @mbabker pointed out, if would be nice if people give it a try before making a release, thanks @ajgarlag for the work you did here

Seb33300:
I will update my project to `dev-master` and let you know if I face any issue.

PR #2793, opened 2024-04-24
Update TimestampableEntity.php

featuriz:
FIX:: [remove double required annotation + attributes](https://github.com/symfony/framework-bundle/commit/f5bd3d9960b97b1c9f903d569152292aa3d67e0f)

### DIFF
--- a/src/Blameable/Traits/BlameableEntity.php
+++ b/src/Blameable/Traits/BlameableEntity.php
@@ -21,8 +21,6 @@ trait BlameableEntity
 {
     /**
      * @var string
-     * @Gedmo\Blameable(on="create")
-     * @ORM\Column(nullable=true)
      */
     #[ORM\Column(nullable: true)]
     #[Gedmo\Blameable(on: 'create')]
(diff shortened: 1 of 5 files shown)


mbabker:
This PR breaks compatibility with ORM 2.x and applications which use annotations.

featuriz:
Ok. I mentioned the problem. Update as per [Fabien Potencier](https://github.com/fabpot) [comment](https://github.com/symfony/framework-bundle/commit/f5bd3d9960b97b1c9f903d569152292aa3d67e0f), I mentioned earlier.

mbabker:
That fix was very specific to Symfony and the way it handles the `@required` annotation and `#[Required]` attribute in its DependencyInjection component.  That's not a generalized fix for every place where annotations and attributes are in place.

As mentioned in your other issue, you're trying to use ORM 3.x and this package is just not compatible with it at this point in time.  The appropriate solution if you 100% need this package right now is to downgrade and continue using ORM 2.x until all of the compatibility issues can be resolved.

featuriz:
Yes my friend @mbabker you are right. **That's not a generalized fix for every place where annotations and attributes are in place.**

In some cased I use [symplify/vendor-patches](https://github.com/symplify/vendor-patches) to override.

PR #2798, opened 2024-05-01
Rewrite the annotations reference

mbabker:
This is a first pass at bringing some of the documentation up-to-date and starts with the annotations reference page.  Right now it is incomplete and tries to act as a "how to set up your entity manager" guide as well (which better belongs in another "getting started" type page that doesn't exist right now, that'll come later).  With this PR, the annotations reference now covers all of the annotations and only acts as a reference guide as far as what annotations exist and how they are configured.

The approach I'm taking here kind of emulates the structure the [ORM's reference page](https://www.doctrine-project.org/projects/doctrine-orm/en/2.19/reference/annotations-reference.html) uses as far as describing the annotations and their configs.  Once this PR is finished, my intention with this aspect of the docs is to clone the page into an attributes reference page (again, similar to how the ORM has separate pages, which is also pertinent here as there are some minor differences in config between the two mapping types) and eventually update the other pages to just link over to this reference instead of trying to duplicate information (as many of the extension pages also have a similar bit of "here's the annotations and how you configure them" information).

It's probably best to just view the rendered page instead of trying to review the diff, so head over to https://github.com/mbabker/DoctrineExtensions/blob/annotation-attribute-docs/doc/annotations.md to see it in action.

### DIFF
--- a/doc/annotations.md
+++ b/doc/annotations.md
@@ -1,567 +1,1505 @@
-# Annotation reference
+# Annotations Reference
 
-Bellow you will find all annotation descriptions used in these extensions.
-There will be introduction on usage with examples. For more detailed usage
-on extensions, refer to their specific documentation.
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


elfantome (approved):

franmomu (approved):

franmomu:
thanks @mbabker, nice work!

PR #2801, opened 2024-05-20
Update SoftDeleteable to work with Doctrine 3.1

elfantome:
Update Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter::addFilterConstraint signature to be compatible with Doctrine\ORM\Query\Filter\SQLFilter::addFilterConstraint

### DIFF
--- a/src/SoftDeleteable/Traits/SoftDeleteableEntity.php
+++ b/src/SoftDeleteable/Traits/SoftDeleteableEntity.php
@@ -23,8 +23,6 @@ use Doctrine\ORM\Mapping as ORM;
 trait SoftDeleteableEntity
 {
     /**
-     * @ORM\Column(type="datetime", nullable=true)
-     *
      * @var \DateTime|null
      */
     #[ORM\Column(type: Types::DATETIME_MUTABLE, nullable: true)]
--- a/src/SoftDeleteable/Traits/SoftDeleteableEntity.php
+++ b/src/SoftDeleteable/Traits/SoftDeleteableEntity.php
@@ -23,6 +23,8 @@ use Doctrine\ORM\Mapping as ORM;
 trait SoftDeleteableEntity
 {
     /**
+     * @ORM\Column(type="datetime", nullable=true)
+     *
      * @var \DateTime|null
      */
     #[ORM\Column(type: Types::DATETIME_MUTABLE, nullable: true)]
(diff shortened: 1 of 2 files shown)


mbabker:
Unfortunately this change breaks compatibility with ORM 2.x.

```sh
Fatal error: Declaration of Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, string $targetTableAlias): string must be compatible with Doctrine\ORM\Query\Filter\SQLFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, $targetTableAlias) in /src/SoftDeleteable/Filter/SoftDeleteableFilter.php on line 54
```

elfantome:
> Unfortunately this change breaks compatibility with ORM 2.x.
> 
> ```shell
> Fatal error: Declaration of Gedmo\SoftDeleteable\Filter\SoftDeleteableFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, string $targetTableAlias): string must be compatible with Doctrine\ORM\Query\Filter\SQLFilter::addFilterConstraint(Doctrine\ORM\Mapping\ClassMetadata $targetEntity, $targetTableAlias) in /src/SoftDeleteable/Filter/SoftDeleteableFilter.php on line 54
> ```

In this case, maybe bumping the version number with a new release can solve the problem?

mbabker:
I just looked at the install stats.  It is way too early to consider dropping ORM 2.x support, and the volunteer resources don't exist to support two major version branches of this package.

There might be a way to change the signature in a way to support both versions (like only adding the return type without the param type), but based on numbers and resources only supporting ORM 3.x right now is not the right decision.

elfantome:
Makes sense.
I update my PR to only add a return type to the method to keep compatibility with ORM 2

mbabker (approved):

src/SoftDeleteable/Traits/SoftDeleteableEntity.php
```diff
@@ -23,8 +23,6 @@
 trait SoftDeleteableEntity
 {
     /**
-     * @ORM\Column(type="datetime", nullable=true)
```
Seb33300, line 26: Needs to be reverted

franmomu (approved):

franmomu:
thanks @elfantome!

PR #2804, opened 2024-05-28
Add an internal trait to allow cross-version compatibility for ORM hydrators

mbabker:
Ref: #2708

Similar to #2758 this adds a trait to help with cross-version compatibility, this time for the ORM hydrators and their changed signatures.  Flat-out adding return types would be treated as a B/C break, so that can't be done cleanly until a 4.0 release, so this trait helps to cover those changes in an easily replaced way when the bridge is no longer needed.


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2807, opened 2024-05-28
Bump versions in CI tools

mbabker:
Upgrades all the CI stuff.  Gist of the changes:

- Use PHP 8.3 for all analysis jobs or high builds (previously a mix of 8.1 or 8.2)
- Bump the Docker Compose config to use PHP 8.3
- Unpin the `ext/mongodb` version in the PHPStan build, latest release is G2G on the latest PHP version to my knowledge (and if it's not then we can use something more recent than 10 minor releases ago 😅)
- `isbang/compose-action` has a new major release

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -19,7 +19,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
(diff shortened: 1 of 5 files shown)


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2813, opened 2024-06-09
Add an attributes reference page

mbabker:
Continuation of #2798 by adding a new attributes reference page.  Also includes some minor tweaks and corrections on the annotations page.


franmomu (approved):

franmomu:
thanks @mbabker!

PR #2781, opened 2024-03-19
Run tests without `doctrine/annotations` installed

mbabker:
MongoDB ODM 2.7 dropped the hard dependency, so let's do a smoke test build without the annotations package installed here and see what happens...

PR #2814, opened 2024-06-09
Finalize ORM 3 Support

mbabker:
This should be the last bit that's needed to finish out #2708.

- ~~Needs #2781 to merge first to be able to build on the CI architecture (it's blowing up pretty badly right now because of the presence of `doctrine/annotations` and the auto-detection code in our `Gedmo\Mapping\MappedEventSubscriber` class)~~ Handled in the testing bootstrap instead
- ~~I haven't even looked at the PHPStan report yet, that can be dealt with once we have consistent green PHPUnit builds~~ Baseline refreshed

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   phpunit:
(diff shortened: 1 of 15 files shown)


franmomu (approved):

franmomu:
Thank you @mbabker for all the work you're doing!

PR #2821, opened 2024-06-24
3.16.0

franmomu:


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+
+## [3.16.0]
 ### Deprecated
 - Support for `doctrine/annotations` has been deprecated and will be removed in 4.0.
```
```suggestion
- Support for defining mapping information from annotations has been deprecated and will be removed in 4.0, use PHP attributes mapping instead.
```

phansys (changes requested):

CHANGELOG.md
phansys: BTW, I think we should also update this statement:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/ca713c5ddd26e0f7eb94b95fa1e863efcc1bbd7d/README.md?plain=1#L64

franmomu: sure, let me change it before releasing, anything else you think we should take a look at?

phansys: No @franmomu, that's all.

Thank you so much!

Issue #2502, opened 2022-08-08
[Loggable] ORM entity relies on deprecated field type


mbabker:
DBAL 3.4 [deprecated the array and object types](https://github.com/doctrine/dbal/blob/3.4.0/UPGRADE.md#deprecated-array-and-object-column-types).  The `Gedmo\Loggable\Entity\MappedSuperclass\AbstractLogEntry` schema uses the array type for its `data` column.  This is going to need migrating somehow.

(While the actual change isn't too bad to make here, it requires a data migration for all users of this package from a serialized string to its replacement (JSON column is the suggestion), which isn't a straightforward thing depending on the types of changes being logged)

chapterjason:
I already tried to overwrite the `LogEntry` and set the type to `Types::JSON`, which results in the following issue:

This is the content of the `data`  field in a `LogEntry`.

```
Array
(
    [content] => <div>test1</div>
    [publishedAt] => 
    [reviewedAt] => 
    [slug] => 
    [state] => draft
    [title] => test1
    [updatedAt] => Array
        (
            [date] => 2022-08-09 20:45:59.585784
            [timezone_type] => 3
            [timezone] => UTC
        )

    [publishedBy] => 
    [reviewedBy] => 
    [updatedBy] => Array
        (
            [uid] => ...
        )

)
```

Using revert results in the following error:

```
Cannot assign array to property App\Entity\Post::$updatedAt of type ?DateTimeImmutable
```

This means the `LogEntryRepository::mapValue` requires some rework. But I don't know how to dynamically convert all possible fields of the dbal type registry without a lot of custom implementation.

mbabker:
I knew the data migration was going to be tricky in the first place because of how the log entries get written.  They actually can also have objects if you're logging those types of fields (we have `odolbeau/phone-number-bundle` installed and log a couple of phone number fields, those are logged as a serialized `libphonenumber\PhoneNumber` instance).

So this almost turns into a bigger "re-evaluate how the logging extension operates" item because not only do you have to think through the data serialization, you have to handle data in all sorts of different states (`DateTimeInterface` objects look to be written as an array, serialized objects, and scalar values) when migrating between types, and whatever format the entry's written in has to be something that can be reversed back onto an entity for the revert workflow (which I honestly didn't realize exists until your comment).

chapterjason:
Experimented a bit and got it working for my use case, tests are also running. Not sure if I miss any cases.

[...]

PR #2822, opened 2024-06-24
Update README.md

andreybolonin:
According to https://github.com/doctrine-extensions/DoctrineExtensions/pull/2821

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ can be easily implemented using Mapping extension to handle the additional metad
 
 | Extensions Version | Compatible Doctrine ORM & Common Library |
 | --- | --- |
+| 3.16 | 3.0+ |
 | 2.4 | 2.5+ |
 | 2.3 | 2.2 - 2.4 |
 


mbabker:
I'm honestly half tempted to suggest to just close this unless you're willing to rebuild the table entirely (or just wipe out the table).  The heading "Compatible Doctrine ORM & Common Library" implies compatibility with `doctrine/orm` and `doctrine/common`, which don't have the same versioning scheme, and does nothing to note compatibility with `doctrine/mongodb-odm` or other required Doctrine libraries (like the DBAL or Persistence packages which are more relevant here than `doctrine/common`).

Personally, I think I'd just update the README to note the supported versions as of now and not try to keep a matrix for what older versions did or did not support (a lot of that can be pulled out of the `composer.json` file for folks who really need it), so instead of a table I'd use this list:

- DBAL: `^3.2`
- ORM: `^2.14` or `^3.0`
- MongoDB ODM: `^2.3`

(And yes, the DBAL callout is important because of #2502 which relates to DBAL features and not the ORM)

franmomu:
thanks @andreybolonin, I've updated https://github.com/doctrine-extensions/DoctrineExtensions/pull/2821 with @mbabker's suggestion. I don't think the older versions are useful there.

PR #2825, opened 2024-06-25
Introduce the Revisionable extension

mbabker:
Ref: #2502

As the issue notes, the loggable extension is incompatible with ORM 4.x due to the removal of the array field type.  The issue also has a patch for migrating to JSON, but because of the need for a data migration, it's not a patch that can be easily dropped in.  Enter the new revisionable extension.

Mostly the same thing as the loggable extension, except for changing the way the data is collected for the history object's data field.  For the ORM, the data was stored as a serialized PHP array, which while it works, is less than optimal:

```sh
a:4:{s:5:"title";s:5:"Title";s:9:"publishAt";O:17:"DateTimeImmutable":3:{s:4:"date";s:26:"2024-06-24 23:00:00.000000";s:13:"timezone_type";i:3;s:8:"timezone";s:3:"UTC";}s:11:"author.name";s:8:"John Doe";s:12:"author.email";s:12:"john@doe.com";}
```

The new extension uses the mapping layers of the ORM and ODM to store the data in its database representation, and for the ORM, as a native JSON field:

```json
{
	"title": "Title",
	"publishAt": "2024-06-24 23:00:00",
	"author.name": "John Doe",
	"author.email": "john@doe.com"
}
```

Another benefit to introducing a new extension is that the old and new versions can live side-by-side and allows for a data migration so long as your log entry data can be unserialized back into PHP and its info mapped to a database value (https://gist.github.com/mbabker/1879a3e55feac953e23cb2f025654052 was something I quickly hacked into the example app code as a proof of concept, a full-on tool could be built out using the logic here).

There are some other extras baked into this branch which generally help improve things too, including:

- Adding templates for the config arrays in the extension listeners
- Adding new methods to the `WrapperInterface` implementations (and `@method` annotated on the interface itself to add to 4.0) to handle mapping values to PHP and database formats for each manager

Differences between the two extensions include:

- Removing the `prePersistLogEntry()` hook that existed in the loggable listener, the Doctrine event manager can be used here without needing to have an open class
- Uses a `Revisionable` attribute instead of `Loggable` at the class level, and uses a `KeepRevisions` attribute instead of the `Versioned` attribute at the property level; for migrating, you're basically looking at a handful of lines changed no matter the mapping (changing a class import and annotation/attribute when using that mapping, XML goes from `<gedmo:loggable/>` to `<gedmo:revisionable/>` and `<gedmo:versioned/>` to `<gedmo:keep-revisions/>`)
- Fully typed to the extent PHP 7.4 allows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- Introduced the `Revisionable` extension as a modern replacement to the `Loggable` extension (#2825)
+
 ### Changed
 - All: Removed the dollar sign from the generated cache ID for extension metadata to ensure only characters mandated by [PSR-6](https://www.php-fig.org/psr/psr-6/#definitions) are used, improving compatibility with caching implementations with strict character requirements (#2978)
 
@@ -48,7 +51,6 @@ a release.
 - SoftDeleteable: Resolved a bug where a soft-deleted object isn't remove from the ObjectManager (#2930)
 
 ### Added
-- Introduced the `Revisionable` extension as a modern replacement to the `Loggable` extension (#2825)
 - IP address provider for use with extensions with IP address references (#2928)
 
 ## [3.19.0] - 2025-02-24
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ a release.
 - SoftDeleteable: Resolved a bug where a soft-deleted object isn't remove from the ObjectManager (#2930)
 
 ### Added
+- Introduced the `Revisionable` extension as a modern replacement to the `Loggable` extension (#2825)
 - IP address provider for use with extensions with IP address references (#2928)
 
 ## [3.19.0] - 2025-02-24
--- a/src/Revisionable/Document/Repository/RevisionRepository.php
+++ b/src/Revisionable/Document/Repository/RevisionRepository.php
@@ -124,13 +124,15 @@ class RevisionRepository extends DocumentRepository
             $mapping = $documentMeta->getFieldMapping($field);
 
             // Fill the embedded document
-            if ($documentWrapper->isEmbeddedAssociation($field) && null !== $value) {
-                assert(class_exists($mapping['targetDocument']));
+            if ($documentWrapper->isEmbeddedAssociation($field)) {
+                if (null !== $value) {
+                    assert(class_exists($mapping['targetDocument']));
 
-                $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
-                $document = $embeddedMetadata->newInstance();
-                $this->fillDocument($document, $value);
-                $value = $document;
+                    $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
+                    $document = $embeddedMetadata->newInstance();
+                    $this->fillDocument($document, $value);
+                    $value = $document;
+                }
             } elseif ($documentMeta->isSingleValuedAssociation($field)) {
                 assert(class_exists($mapping['targetDocument']));
 
--- a/src/Revisionable/Document/Repository/RevisionRepository.php
+++ b/src/Revisionable/Document/Repository/RevisionRepository.php
@@ -124,15 +124,13 @@ class RevisionRepository extends DocumentRepository
             $mapping = $documentMeta->getFieldMapping($field);
 
             // Fill the embedded document
-            if ($documentWrapper->isEmbeddedAssociation($field)) {
-                if (!empty($value)) {
-                    assert(class_exists($mapping['targetDocument']));
+            if ($documentWrapper->isEmbeddedAssociation($field) && null !== $value) {
+                assert(class_exists($mapping['targetDocument']));
 
-                    $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
-                    $document = $embeddedMetadata->newInstance();
-                    $this->fillDocument($document, $value);
-                    $value = $document;
-                }
+                $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
+                $document = $embeddedMetadata->newInstance();
+                $this->fillDocument($document, $value);
+                $value = $document;
             } elseif ($documentMeta->isSingleValuedAssociation($field)) {
                 assert(class_exists($mapping['targetDocument']));
 
--- a/src/Revisionable/Document/Repository/RevisionRepository.php
+++ b/src/Revisionable/Document/Repository/RevisionRepository.php
@@ -114,7 +114,7 @@ class RevisionRepository extends DocumentRepository
         assert($documentMeta instanceof ClassMetadata);
 
         $config = $this->getListener()->getConfiguration($this->getDocumentManager(), $documentMeta->getName());
-        $fields = $config['versioned'];
+        $fields = $config['versionedFields'];
 
         foreach ($data as $field => $value) {
             if (!in_array($field, $fields, true)) {
--- a/src/Revisionable/Document/Repository/RevisionRepository.php
+++ b/src/Revisionable/Document/Repository/RevisionRepository.php
@@ -142,14 +142,12 @@ class RevisionRepository extends DocumentRepository
             }
 
             $documentWrapper->setPropertyValue($field, $value);
-            unset($fields[$field]);
+            unset($fields[array_search($field, $fields, true)]);
         }
 
-        /*
         if (count($fields)) {
-            throw new UnexpectedValueException(sprintf('Could not fully revert document %s to version %d.', $documentMetadata->getName(), $version));
+            throw new UnexpectedValueException(sprintf('Could not fully revert document %s.', $documentMeta->getName()));
         }
-        */
     }
 
     /**
--- a/src/Revisionable/Document/Repository/RevisionRepository.php
+++ b/src/Revisionable/Document/Repository/RevisionRepository.php
@@ -1,0 +1,182 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Revisionable\Document\Repository;
+
+use Doctrine\ODM\MongoDB\Mapping\ClassMetadata;
+use Doctrine\ODM\MongoDB\Repository\DocumentRepository;
+use Gedmo\Exception\RuntimeException;
+use Gedmo\Exception\UnexpectedValueException;
+use Gedmo\Revisionable\Document\MappedSuperclass\AbstractRevision;
+use Gedmo\Revisionable\Revisionable;
+use Gedmo\Revisionable\RevisionableListener;
+use Gedmo\Tool\Wrapper\MongoDocumentWrapper;
+
+/**
+ * The RevisionRepository has some useful functions to interact with revisions.
+ *
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ *
+ * @template T of Revisionable|object
+ *
+ * @template-extends DocumentRepository<AbstractRevision<T>>
+ */
+class RevisionRepository extends DocumentRepository
+{
+    /**
+     * The revisionable listener associated with the document manager for the {@see AbstractRevision} document.
+     *
+     * @var RevisionableListener<T>|false|null
+     */
+    private $listener = false;
+
+    /**
+     * Loads all revisions for the given document
+     *
+     * @param T $document
+     *
+     * @return list<AbstractRevision<T>>
+     */
+    public function getRevisions(object $document): array
+    {
+        $documentWrapper = new MongoDocumentWrapper($document, $this->getDocumentManager());
+
+        $documentId = (string) $documentWrapper->getIdentifier(false, true);
+        $documentClass = $documentWrapper->getMetadata()->getName();
+
+        $qb = $this->createQueryBuilder();
+        $qb->field('revisionableId')->equals($documentId);
+        $qb->field('revisionableClass')->equals($documentClass);
+        $qb->sort('version', 'DESC');
+
+        return $qb->getQuery()->getIterator()->toArray();
+    }
+
+    /**
+     * Reverts the given document to the requested version, restoring all versioned fields to the state of that revision.
+     *
+     * Callers to this method will need to persist and flush changes to the document.
+     *
+     * @param T            $document
+     * @param positive-int $version
+     *
+     * @throws UnexpectedValueException
+     */
+    public function revert(object $document, int $version = 1): void
+    {
+        $documentWrapper = new MongoDocumentWrapper($document, $this->getDocumentManager());
+
+        $documentMetadata = $documentWrapper->getMetadata();
+        $documentId = (string) $documentWrapper->getIdentifier(false, true);
+        $documentClass = $documentMetadata->getName();
+
+        $qb = $this->createQueryBuilder();
+        $qb->field('revisionableId')->equals($documentId);
+        $qb->field('revisionableClass')->equals($documentClass);
+        $qb->field('version')->lte($version);
+        $qb->sort('version', 'ASC');
+
+        $revisions = $qb->getQuery()->getIterator()->toArray();
+
+        if ([] === $revisions) {
+            throw new UnexpectedValueException(sprintf('Could not find any revisions for version %d of document %s.', $version, $documentClass));
+        }
+
+        $data = [[]];
+
+        while ($revision = array_shift($revisions)) {
+            $data[] = $revision->getData();
+        }
+
+        $data = array_merge(...$data);
+
+        $this->fillDocument($document, $data);
+    }
+
+    /**
+     * Fills a document's versioned fields with the given data
+     *
+     * @param T                    $document
+     * @param array<string, mixed> $data
+     */
+    protected function fillDocument(object $document, array $data): void
+    {
+        $documentWrapper = new MongoDocumentWrapper($document, $this->getDocumentManager());
+
+        $documentMeta = $documentWrapper->getMetadata();
+
+        assert($documentMeta instanceof ClassMetadata);
+
+        $config = $this->getListener()->getConfiguration($this->getDocumentManager(), $documentMeta->getName());
+        $fields = $config['versioned'];
+
+        foreach ($data as $field => $value) {
+            if (!in_array($field, $fields, true)) {
+                continue;
+            }
+
+            $mapping = $documentMeta->getFieldMapping($field);
+
+            // Fill the embedded document
+            if ($documentWrapper->isEmbeddedAssociation($field)) {
+                if (!empty($value)) {
+                    assert(class_exists($mapping['targetDocument']));
+
+                    $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
+                    $document = $embeddedMetadata->newInstance();
+                    $this->fillDocument($document, $value);
+                    $value = $document;
+                }
+            } elseif ($documentMeta->isSingleValuedAssociation($field)) {
+                assert(class_exists($mapping['targetDocument']));
+
+                $value = $value ? $this->getDocumentManager()->getReference($mapping['targetDocument'], $value) : null;
+            } else {
+                $value = $value ? $documentWrapper->convertToPHPValue($value, $documentMeta->getTypeOfField($field)) : null;
+            }
+
+            $documentWrapper->setPropertyValue($field, $value);
+            unset($fields[$field]);
+        }
+
+        /*
+        if (count($fields)) {
+            throw new UnexpectedValueException(sprintf('Could not fully revert document %s to version %d.', $documentMetadata->getName(), $version));
+        }
+        */
+    }
+
+    /**
+     * Get the revisionable listener associated with the document manager for the {@see AbstractRevision} document.
+     *
+     * @throws RuntimeException if the listener is not found
+     *
+     * @return RevisionableListener<T>
+     */
+    private function getListener(): RevisionableListener
+    {
+        if ($this->listener instanceof RevisionableListener) {
+            return $this->listener;
+        }
+
+        if (false === $this->listener) {
+            foreach ($this->getDocumentManager()->getEventManager()->getAllListeners() as $listeners) {
+                foreach ($listeners as $listener) {
+                    if ($listener instanceof RevisionableListener) {
+                        return $this->listener = $listener;
+                    }
+                }
+            }
+
+            $this->listener = null;
+        }
+
+        throw new RuntimeException('The revisionable listener was not registered to the document manager.');
+    }
+}
--- a/src/Revisionable/Mapping/Driver/Attribute.php
+++ b/src/Revisionable/Mapping/Driver/Attribute.php
@@ -95,14 +95,14 @@ class Attribute extends AbstractAnnotationDriver
                     continue;
                 }
 
-                $config['versioned'][] = $field;
+                $config['versionedFields'][] = $field;
             }
         }
 
         // Validate configuration
         if (!$meta->isMappedSuperclass && $config) {
             // The revisionable flag must be set, except for embedded models, and the versioned config should be a non-empty array
-            if (isset($config['versioned']) && !isset($config['revisionable']) && !$this->isEmbed($meta)) {
+            if (isset($config['versionedFields']) && !isset($config['revisionable']) && !$this->isEmbed($meta)) {
                 throw new InvalidMappingException(sprintf('Class "%s" has "%s" annotated fields but is missing the "%s" class annotation.', $meta->getName(), Versioned::class, Revisionable::class));
             }
         }
@@ -130,7 +130,7 @@ class Attribute extends AbstractAnnotationDriver
                     continue;
                 }
 
-                $config['versioned'][] = $embeddedField;
+                $config['versionedFields'][] = $embeddedField;
             }
         }
 
--- a/src/Revisionable/Mapping/Driver/Attribute.php
+++ b/src/Revisionable/Mapping/Driver/Attribute.php
@@ -103,7 +103,7 @@ class Attribute extends AbstractAnnotationDriver
         if (!$meta->isMappedSuperclass && $config) {
             // The revisionable flag must be set, except for embedded models, and the versioned config should be a non-empty array
             if (isset($config['versioned']) && (!$this->isEmbed($meta) && !isset($config['revisionable']))) {
-                throw new InvalidMappingException(sprintf("Class '%s' has '%s' annotated fields but is missing the '%s' class annotation.", $meta->getName(), Versioned::class, Revisionable::class));
+                throw new InvalidMappingException(sprintf('Class "%s" has "%s" annotated fields but is missing the "%s" class annotation.', $meta->getName(), Versioned::class, Revisionable::class));
             }
         }
 
--- a/src/Revisionable/Mapping/Driver/Attribute.php
+++ b/src/Revisionable/Mapping/Driver/Attribute.php
@@ -13,8 +13,8 @@ use Doctrine\ODM\MongoDB\Mapping\ClassMetadata as MongoDBDOMClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadata as ORMClassMetadata;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use Gedmo\Exception\InvalidMappingException;
+use Gedmo\Mapping\Annotation\KeepRevisions;
 use Gedmo\Mapping\Annotation\Revisionable;
-use Gedmo\Mapping\Annotation\Versioned;
 use Gedmo\Mapping\Driver\AbstractAnnotationDriver;
 
 /**
@@ -60,7 +60,7 @@ class Attribute extends AbstractAnnotationDriver
         foreach ($class->getProperties() as $property) {
             $field = $property->getName();
 
-            if ($this->reader->getPropertyAnnotation($property, Versioned::class)) {
+            if ($this->reader->getPropertyAnnotation($property, KeepRevisions::class)) {
                 if ($meta->isCollectionValuedAssociation($field)) {
                     throw new InvalidMappingException(sprintf('Cannot version field %s::$%s, collection valued associations are not supported.', $meta->getName(), $field));
                 }
@@ -103,7 +103,7 @@ class Attribute extends AbstractAnnotationDriver
         if (!$meta->isMappedSuperclass && $config) {
             // The revisionable flag must be set, except for embedded models, and the versioned config should be a non-empty array
             if (isset($config['versionedFields']) && !isset($config['revisionable']) && !$this->isEmbed($meta)) {
-                throw new InvalidMappingException(sprintf('Class "%s" has "%s" annotated fields but is missing the "%s" class annotation.', $meta->getName(), Versioned::class, Revisionable::class));
+                throw new InvalidMappingException(sprintf('Class "%s" has "%s" annotated fields but is missing the "%s" class annotation.', $meta->getName(), KeepRevisions::class, Revisionable::class));
             }
         }
 
@@ -121,7 +121,7 @@ class Attribute extends AbstractAnnotationDriver
     private function inspectEmbeddedForVersioned(string $field, array $config, ORMClassMetadata $meta): array
     {
         foreach ((new \ReflectionClass($meta->embeddedClasses[$field]['class']))->getProperties() as $property) {
-            if ($this->reader->getPropertyAnnotation($property, Versioned::class)) {
+            if ($this->reader->getPropertyAnnotation($property, KeepRevisions::class)) {
                 $embeddedField = $field.'.'.$property->getName();
 
                 if (isset($meta->embeddedClasses[$embeddedField])) {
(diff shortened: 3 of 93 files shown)


simoheinonen:
Having no username set in the `RevisionableListener` doesn't work because `->setUsername()` expects a string even though it is nullable in the database

mbabker:
> Having no username set in the `RevisionableListener` doesn't work because `->setUsername()` expects a string even though it is nullable in the database

Thanks, I've updated everything to account for that.

simoheinonen:
What's the status of this PR? Anything I can do to help/test? Would like to see it merged soon!

mbabker:
> What's the status of this PR? Anything I can do to help/test? Would like to see it merged soon!

I don't think I have anything big to add to this PR at this point.  It really just needs testing and making sure the idea is solid (especially as part of the rationale for a new extension versus trying to make a version of loggable that is DBAL 4 friendly is allowing both the old and new data to live side-by-side so an app can either keep that old data by either polyfilling the deprecated DBAL array type or migrating it into the new extension).

mustanggb:
I started trying to test/use this in an ODM configuration.

Notables so far:
- RevisionableListener is final - I was previously extending LoggableListener, but perhaps will be able to make the necessary changes in a RevisionInterface implementation instead.
- RevisionInterface "sets" has lots of `void` returns - In LogEntryInterface these weren't hardcoded, perhaps not the end of the world.
- RevisionInterface has a lot of string that are no longer `nullable` - Again, probably fine for now.
- Missing `StofDoctrineExtensionsBundle` integration - Biggest stumpling block so far as event listeners aren't triggering at all.
- No documentation/examples - i.e. `doc/revisionable.md` and updated `doc/annotations.md`

So at the moment I've not been able to create/trigger any revisions, I guess there is a way to enable this without `StofDoctrineExtensionsBundle`, but if you were able to create a `dev` version with support then I'll be able to test further.

mbabker:
> RevisionableListener is final - I was previously extending LoggableListener, but perhaps will be able to make the necessary changes in a RevisionInterface implementation instead.

The loggable listener has to be open because of its `prePersistLogEntry` hook point; with this implementation, I decided to rely on the events that the Doctrine object managers emit and get rid of that, and as a result, I've landed on a final class.

Maybe it's too soon to go with a hard final listener, but I'm not going to take the "no I'm not going to open it up for inheritance" standpoint if there's a legitimate use case that a hard final blocks.  Maybe somewhere down the road, something can be figured out for how interfaces can be built for the listeners (as right now the hook points into Doctrine are all Doctrine event listeners, so the interface is more so specifying the required events to subscribe to and less what a public-ish API looks like).

> RevisionInterface "sets" has lots of void returns - In LogEntryInterface these weren't hardcoded, perhaps not the end of the world.

The new extension is more strictly typed given it's being written as new code with the PHP 7.4 minimum in mind, compared to all of the other extensions which were written in PHP 5 times; `LogEntryInterface` already uses `@return void` annotations so `RevisionInterface` solidifies this with native return types.

> RevisionInterface has a lot of string that are no longer nullable - Again, probably fine for now.

I wanted to make the revision model a little more strict in terms of having valid state, hence the introduction of the static `RevisionInterface::createRevision()` constructor and the listener using that over `new Revision()`; the three fields this effects are the action, version, and logged at timestamp (all of which can be initialized in those constructors), the rest of the fields do remain nullable as the model is designed for everything else to support null values (you don't have to have a user for a revision, you don't have to log the data state for a revision, etc.).

> Missing `StofDoctrineExtensionsBundle` integration - Biggest stumpling block so far as event listeners aren't triggering at all.

That'd have to be done after this PR landed.  Trying to shoot a PR off over there would just result in a PR with constantly failing builds, it's easier to handle that update after this change lands.

For manual config in a Symfony app, you can take the loggable listener service in https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/doc/frameworks/symfony.md#extensions-compatible-with-all-managers and change that to the revisionable listener (it'll hook into the same events so it's just changing the service ID and class name)

> No documentation/examples - i.e. `doc/revisionable.md` and updated `doc/annotations.md`

It'll get done before this merges.  I did call out the needed mapping changes in the PR description, but the gist of it is you switch `@Gedmo\Loggable` to `@Gedmo\Revisionable` for annotations and attributes, and if you're using the `logEntryClass` param to use a custom model, change that to `revisionClass`.  The `@Gedmo\Versioned` annotation/attribute gets reused, with the goal here being to make the config migration as minimal as possible.

mustanggb:
If action is no longer nullable does it make sense set a default then, to avoid `getAction()` complaining?

e.g.
```
  #[ODM\Field(name: 'action', type: Type::STRING)]
- protected ?string $action = null;
+ protected string $action = self::ACTION_CREATE;
```
(or whatever, if the syntax is slightly off)

mbabker:
> If action is no longer nullable does it make sense set a default then, to avoid `getAction()` complaining?

Probably; this has all been iterated over a few times so little pain points like that are bound to have slipped in by accident.

mbabker:
> > If action is no longer nullable does it make sense set a default then, to avoid `getAction()` complaining?
> 
> Probably; this has all been iterated over a few times so little pain points like that are bound to have slipped in by accident.

This has been updated.

mustanggb:
> For manual config in a Symfony app, you can take the loggable listener service and change that to the revisionable listener

For reference this is what I'm using for now.
Also required changing `doctrine.event_listener` to `doctrine_mongodb.odm.event_listener`.

```
services:
  gedmo.mapping.driver.attribute:
    class: Gedmo\Mapping\Driver\AttributeReader

  gedmo.listener.revisionable:
    class: Gedmo\Revisionable\RevisionableListener
    tags:
      - { name: doctrine_mongodb.odm.event_listener, event: 'onFlush' }
      - { name: doctrine_mongodb.odm.event_listener, event: 'loadClassMetadata' }
      - { name: doctrine_mongodb.odm.event_listener, event: 'postPersist' }
    calls:
      - [ setAnnotationReader, [ '@gedmo.mapping.driver.attribute' ] ]
```

mustanggb:
Next problem, there seems to be a circular issue with ODM embedded documents.

1. To appear in revision data the field in the embedded document requires versioned attribute.
2. But this gives a mapping exception about missing revisionable attribute.
3. Adding the revisionable attribute gives another mapping exception that embedded objects can't have a revisionable attribute.

In fairness the same problem exists with loggable, I guess it worked at some point, then got broken.

I would suggest that embedded documents should require versioned attributes against fields, but not revisionable/loggable attributes against the class.

And in testing (commenting out the revisionable/loggable attribute checks) this works.
i.e. revision (and log) data gets set for embedded documents.

For searchability the checks, that I'm suggesting should be exempted for embedded documents are:
```
throw new InvalidMappingException(sprintf("Class '%s' is mapped as an embedded object and cannot specify the revision class property.", $meta->getName()));
```
```
throw new InvalidMappingException("Class must be annotated with Loggable annotation in order to track versioned fields in class - {$meta->getName()}");
```

mbabker:
Do you have an example you can share where it's not working? I'm not a ODM user, but I went through the effort to get MongoDB running in my setup for a couple OSS projects so I was able to build out the test fixtures for this new code, which include embedded documents.

One of the differences between the ODM and ORM is how they track embedded objects. The ORM inlines embeds as part of the entity itself, while the ODM tracks them as managed objects in the unit of work. An earlier iteration tried to not use the Revisionable attribute on embedded documents, but because of that difference between the two object managers, the embedded document wasn't being handled in the event listener.

mbabker:
I'll play with that a bit and see what I can come up with.

> The difference I notice is in the database they're all strings for loggable, whereas for revisionable they are the correct field type.

That's the other big change in this new extension.  With loggable, it would just take the values from the objects and serialize the array.  In the new code, I'm going through the `Type::getType($type)->convertToDatabaseValue($value);` APIs to log everything as its database representation (so no more serialized DateTime objects in the log data, as an example, it'll be a formatted string).

mbabker:
OK, I think I have this back to not requiring the revisionable attribute (or similar config in XML/YAML) on embedded models.  The fields that should be versioned still need to be configured as such, but that looks to be the same with the current loggable extension so this should be OK.

mustanggb:
Fantastic, that appears to work perfectly... for revisionable.

Which in theory should be fine, except I think the situation is:
- ✔️ Revisionable + ORM
- ✔️ Loggable + ORM
- ✔️ Revisionable + ODM
- ❌ Loggable + ODM

As an embedded document with versioned attributes gives the error:

```
Class must be annotated with Loggable annotation in order to track versioned fields in class - App\Document\Price
```

I get that it's probably kind of out of scope to add features to loggable in this PR, but as it's reusing the versioned attribute, and loggable and revisionable probably need to exist in tandem for a period of time, what would you suggest as the best resolution?

My one line change to `isClassAnnotationInValid()` might not be the full solution, but it does resolve it for my use-case.

---

On second thought, what you're probably expecting to happen is that the two exist in tandem in code only, and aren't actually used together. Meaning ODM embed's aren't supported by loggable (and probably never were), but are supported by revisionable, so you'd have to migrate, disabling loggable in the process.

mbabker:
In my mind, the two extensions can live side-by-side (which also gives a way to migrate data across if you need to, like the hacked together gist in the OP shows can be done), but a model shouldn't be both loggable and revisionable at the same time.  That does get trickier with embedded models because of the behavioral differences between object managers, but if tweaking the loggable extension makes the migration more practical, then we should do that as well.

As a separate PR (because it sounds like the limitations with loggable and the ODM are already there and addressing that shouldn't rely on this PR landing anytime soon), I'd say to make the changes you're suggesting in the mapping drivers for the loggable extension, take the `MongoDBODMMappingTestCase` from this PR, and set up a `LoggableMongoDBODMMappingTest` (similar to the `RevisionableMongoDBODMMappingTest` in this PR) and all the relevant fixtures to make sure the mapping behaviors are working as expected with the ODM.

jhogervorst:
@mbabker Hey 👋 Great work on this new extension!

Do you have an indication of this PR's status, or an idea when it could get merged?

In my project we'll be needing something like this soon. And because `Loggable` is getting end-of-life, I'm following this PR curiously. It would be great if we could start with `Revisionable`, so we won't have to do any migration in the future.

I hope you don't mind me asking this directly 😊 I've sent a contribution through GitHub Sponsors for your work on this!

Coffee2CodeNL:
Eagerly awaiting this, any progress?

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- Introduced the revisionable extension as a modern replacement to the loggable extension (#2825)
```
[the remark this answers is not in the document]
mbabker: Applied while rebasing

src/Revisionable/Document/Repository/RevisionRepository.php
```diff
@@ -0,0 +1,182 @@
...
+        foreach ($data as $field => $value) {
+            if (!in_array($field, $fields, true)) {
+                continue;
+            }
+
+            $mapping = $documentMeta->getFieldMapping($field);
+
+            // Fill the embedded document
+            if ($documentWrapper->isEmbeddedAssociation($field)) {
+                if (!empty($value)) {
```
phansys, line 128: What are the values allowed for `$value`? As `empty()` will return `true` for falsy values like `false` and `0`, I'd try to avoid using this function in favor of a stricter comparison.

src/Revisionable/Document/Repository/RevisionRepository.php
```diff
@@ -0,0 +1,182 @@
...
+
+                    $embeddedMetadata = $this->getDocumentManager()->getClassMetadata($mapping['targetDocument']);
+                    $document = $embeddedMetadata->newInstance();
+                    $this->fillDocument($document, $value);
+                    $value = $document;
+                }
+            } elseif ($documentMeta->isSingleValuedAssociation($field)) {
+                assert(class_exists($mapping['targetDocument']));
+
+                $value = $value ? $this->getDocumentManager()->getReference($mapping['targetDocument'], $value) : null;
```
phansys, line 139: Could we use a stricter condition around `$value`?

src/Revisionable/Document/Repository/RevisionRepository.php
```diff
@@ -0,0 +1,182 @@
...
+            } else {
+                $value = $value ? $documentWrapper->convertToPHPValue($value, $documentMeta->getTypeOfField($field)) : null;
+            }
+
+            $documentWrapper->setPropertyValue($field, $value);
+            unset($fields[$field]);
+        }
+
+        /*
```
phansys, line 148: What is the intention for this comment?

src/Revisionable/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,172 @@
...
+
+                $config['versioned'][] = $field;
+            }
+        }
+
+        // Validate configuration
+        if (!$meta->isMappedSuperclass && $config) {
+            // The revisionable flag must be set, except for embedded models, and the versioned config should be a non-empty array
+            if (isset($config['versioned']) && (!$this->isEmbed($meta) && !isset($config['revisionable']))) {
+                throw new InvalidMappingException(sprintf("Class '%s' has '%s' annotated fields but is missing the '%s' class annotation.", $meta->getName(), Versioned::class, Revisionable::class));
```
phansys, line 106:
```suggestion
                throw new InvalidMappingException(sprintf('Class "%s" has "%s" annotated fields but is missing the "%s" class annotation.', $meta->getName(), Versioned::class, Revisionable::class));
```

src/Revisionable/Mapping/Driver/Attribute.php
```diff
@@ -0,0 +1,172 @@
...
+                }
+
+                $config['versioned'][] = $field;
+            }
+        }
+
+        // Validate configuration
+        if (!$meta->isMappedSuperclass && $config) {
+            // The revisionable flag must be set, except for embedded models, and the versioned config should be a non-empty array
+            if (isset($config['versioned']) && (!$this->isEmbed($meta) && !isset($config['revisionable']))) {
```
phansys, line 105: Move the cheaper check first and remove the superfluos parentheses.

```suggestion
            if (isset($config['versioned']) && !isset($config['revisionable']) && !$this->isEmbed($meta)) {
```

src/Revisionable/Mapping/Driver/Xml.php
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Mapping driver for the revisionable extension which reads extended metadata from an XML mapping document for a revisionable class.
+ *
+ * @author Boussekeyt Jules <jules.boussekeyt@gmail.com>
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ * @author Miha Vrhovnik <miha.vrhovnik@gmail.com>
+ *
+ * @internal
+ */
+class Xml extends BaseXml
```
phansys, line 27: Any reason not to make this final?

tests/Gedmo/Mapping/Fixture/Document/EmbeddedRevisionable.php
```diff
@@ -0,0 +1,34 @@
...
+use Doctrine\ODM\MongoDB\Types\Type;
+use Gedmo\Mapping\Annotation as Gedmo;
+
+/**
+ * @author Fabian Sabau <fabian.sabau@socialbit.de>
+ *
+ * @ODM\EmbeddedDocument
+ */
+#[ODM\EmbeddedDocument]
+class EmbeddedRevisionable
```
phansys, line 24:
```suggestion
final class EmbeddedRevisionable
```

tests/Gedmo/Timestampable/ChangeTest.php
```diff
@@ -132,9 +132,7 @@ public function getDateValue($meta, $field): ?\DateTime
     }
 }
 
-/**
- * @phpstan-extends AbstractTrackingListener<array, TimestampableAdapter>
- */
+/** @phpstan-extends AbstractTrackingListener<array, TimestampableAdapter> */
```
phansys, line 135: This change seems unrelated to this PR.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: I'd have to inspect what's actually getting stored in different scenarios to see what's going on here, but this is largely copy/pasted from the loggable repository and I'm not a heavy MongoDB user so I'm not sure if stuff that'll cause issues with `empty()` is actually getting logged.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: Maybe a `!== null` would work in this branch?  It'd be another one of those "we gotta test it thoroughly" things, and TBH there's only a very basic "does it work without erroring" smoke test for the `revert()` capabilities so we'd need to build out some good test fixtures here.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: It's been there ever since https://github.com/doctrine-extensions/DoctrineExtensions/pull/1212 was accepted in the log entry repository.  The ORM repository throws when it couldn't process all fields, no idea why the MongoDB repository has the check commented out.

tests/Gedmo/Mapping/Fixture/Document/EmbeddedRevisionable.php
mbabker: The MongoDB ODM doesn't have an issue with final embedded objects?  (I know there are quirks with final in some places in the OMs but I don't remember what the deal is with embeddable)

tests/Gedmo/Timestampable/ChangeTest.php
mbabker: Yeah that's definitely from different iterations of messing with the generics and rebasing things 😅

src/Revisionable/Mapping/Driver/Xml.php
mbabker: Not really, added the keyword.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: For my reference (this is a bug in loggable and this implementation)

Adding a `dump('fields list start', $fields);` after `$fields = $config['versioned']` gives:

```sh
1 "fields list start"
2 array:5 [
  0 => "subject"
  1 => "message"
  2 => "writtenAt"
  3 => "article"
  4 => "author"
]
```

And `dump('fields list end', $fields);` before the commented block:

```sh
1 "fields list end"
2 array:5 [
  0 => "subject"
  1 => "message"
  2 => "writtenAt"
  3 => "article"
  4 => "author"
]
```

`unset($fields[$field]);` can't work because the fields list isn't keyed by name.  The ORM repository uses `unset($fields[array_search($field, $fields, true)]);` so this change needs to be tested for both.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker:
```sh
1 "isEmbeddedAssociation"
2 array:2 [
  "name" => "John Doe"
  "email" => "john@doe.com"
]
```

The only time we enter this branch with the current test fixtures it has an array with the embedded document's revisioned data.  Given that we recursively get into another `fillDocument()` method call and the `$value` is passed to the typehinted `array $data` argument, we might be able to go to a `$value !== []` check.  This will need more testing though to validate the assumption.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: This is where me not knowing the MongoDB ODM well doesn't help any.

```
1 "isSingleValuedAssociation"
2 "67ce41458a6433683e0c45e6"
```

This is the only time the branch gets reached in the `RevisionableDocumentTest`.  Looking at the comment fixture, it looks like this is going to be the `ReferenceOne` relation to the article, so I guess `$value` in this case is the document identifier?  If that's the case then this could most likely work as a `!== null` check, assuming that it goes from a non-null to a null value when a non-required reference is removed.

tests/Gedmo/Mapping/Fixture/Document/EmbeddedRevisionable.php
phansys: I'm not sure, but I don't have enough time either.
So for now, let's keep it extensible.

src/Revisionable/Document/Repository/RevisionRepository.php
phansys: Friendly ping @franmomu :pray:

Based on your experience, maybe you have a chance to help with this.

phansys (commented):
We should add a keyword for this new extension at [`composer.json`](https://github.com/doctrine-extensions/DoctrineExtensions/blob/692cd8aac45ea517d60a712895e09bbacecd8ee4/composer.json#L17-L18).

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: I added more tests here and so far it's either an array or null.  So switched to an explicit `null !== $value` check for the moment, but if this causes problems then this can be loosened up.

src/Revisionable/Document/Repository/RevisionRepository.php
mbabker: The revisionable repository's updated with this change and the commented code in the loggable repository is removed (funny enough the loggable ORM repository also had this commented out, at this point it's just easier to remove the commented code and not try to enable it there and just ship this with it enabled).

src/Revisionable/RevisionableListener.php
```diff
@@ -0,0 +1,390 @@
...
+    {
+        $om = $ea->getObjectManager();
+        $wrapped = AbstractWrapper::wrap($object, $om);
+        $meta = $wrapped->getMetadata();
+        $config = $this->getConfiguration($om, $meta->getName());
+        $uow = $om->getUnitOfWork();
+        $newValues = [];
+
+        foreach ($ea->getObjectChangeSet($uow, $object) as $field => $changes) {
+            if (empty($config['versioned']) || !in_array($field, $config['versioned'], true)) {
```
phansys, line 289: Do you think we can replace all the added conditions using `empty()` with its stricter variants?

I'd like to reduce the work we must do later, narrowing conditions like this in the existing codebase.

mbabker: Yeah, I'll see what I can do with it.

antoniovj1:
Hi! Do you need some help with this?

Thanks!!

mbabker:
> Not sure if bug or feature but I can't have have only `#[Loggable] + #[Versioned]` in an entity but it forces me to use both `#[Loggable] + #[Revisionable] + #[Versioned]`
> 
> ```
> Class "App\Entity\VendorOrder" has "Gedmo\Mapping\Annotation\Versioned" annotated fields but is missing the "Gedmo\Mapping\Annotation\Revisionable" class annotation.
> ```

I'd need to see an entity to look at this more.  The loggable mapping didn't change any and the revisionable mapping only looks for the `Revisionable` and `Versioned` attributes.

simoheinonen:
> > Not sure if bug or feature but I can't have have only `#[Loggable] + #[Versioned]` in an entity but it forces me to use both `#[Loggable] + #[Revisionable] + #[Versioned]`
> > ```
> > Class "App\Entity\VendorOrder" has "Gedmo\Mapping\Annotation\Versioned" annotated fields but is missing the "Gedmo\Mapping\Annotation\Revisionable" class annotation.
> > ```
> 
> I'd need to see an entity to look at this more. The loggable mapping didn't change any and the revisionable mapping only looks for the `Revisionable` and `Versioned` attributes.

```yaml
# services.yaml
    gedmo.mapping.driver.attribute:
        class: Gedmo\Mapping\Driver\AttributeReader

    gedmo.listener.revisionable:
        class: Gedmo\Revisionable\RevisionableListener
        tags:
            - { name: doctrine.event_listener, event: 'onFlush' }
            - { name: doctrine.event_listener, event: 'loadClassMetadata' }
            - { name: doctrine.event_listener, event: 'postPersist' }
        calls:
             - [ setAnnotationReader, [ '@gedmo.mapping.driver.attribute' ] ]

# doctrine.yaml
orm:
  # ...
  entity_managers:
    default:
      # ...
      mappings:
        # ...
        gedmo_loggable:
          type: attribute
          prefix: Gedmo\Loggable\Entity
          dir: "%kernel.project_dir%/vendor/gedmo/doctrine-extensions/src/Loggable/Entity/"
          alias: GedmoLoggable # (optional) it will default to the name set for the mappingmapping
          is_bundle: false
        revisionable:
          type: attribute
          alias: GedmoRevisionable
          prefix: Gedmo\Revisionable\Entity
          dir: "%kernel.project_dir%/vendor/gedmo/doctrine-extensions/src/Revisionable/Entity"
          is_bundle: false

# stof_doctrine_extensions.yaml
stof_doctrine_extensions:
    default_locale: en_US
    orm:
        default:
            timestampable: true
            blameable: true
            loggable: true
            sortable: true
```



```php
<?php

declare(strict_types=1);

namespace App\Entity;

use App\Repository\VendorRepository;
use Doctrine\DBAL\Types\Types;
use Doctrine\ORM\Mapping\Column;
use Doctrine\ORM\Mapping\GeneratedValue;
use Doctrine\ORM\Mapping\Id;
use Gedmo\Mapping\Annotation\Loggable;
use Gedmo\Mapping\Annotation\Versioned;

#[Loggable]
#[ORM\Table(name: 'vendors')]
#[ORM\Entity(repositoryClass: VendorRepository::class)]
class Vendor
{
    #[Column(name: 'vendor_id', type: Types::INTEGER)]
    #[Id]
    #[GeneratedValue(strategy: 'IDENTITY')]
    private ?int $vendorId = null;

    #[Versioned]
    #[Column(name: 'name', type: Types::STRING, length: 255, unique: true)]
    private string $name;

    public function getVendorId(): ?int
    {
        return $this->vendorId;
    }

    public function getName(): string
    {
        return $this->name;
    }

    public function setName(string $name): void
    {
        $this->name = $name;
    }
}
```
⚠️  doesn't work but errors with 
```
Class "App\Entity\Vendor" has "Gedmo\Mapping\Annotation\Versioned" annotated fields but is missing the "Gedmo\Mapping\Annotation\Revisionable" class annotation.
```

If I add `#[Revisionable]` (also to all other entities having `#[Loggable]`) then (both of them) works:
```php
<?php

declare(strict_types=1);

namespace App\Entity;

use App\Repository\VendorRepository;
use Doctrine\DBAL\Types\Types;
use Doctrine\ORM\Mapping\Column;
use Doctrine\ORM\Mapping\GeneratedValue;
use Doctrine\ORM\Mapping\Id;
use Gedmo\Mapping\Annotation\Loggable;
use Gedmo\Mapping\Annotation\Revisionable;
use Gedmo\Mapping\Annotation\Versioned;

#[Loggable]
#[Revisionable]
#[ORM\Table(name: 'vendors')]
#[ORM\Entity(repositoryClass: VendorRepository::class)]
class Vendor
{
    #[Column(name: 'vendor_id', type: Types::INTEGER)]
    #[Id]
    #[GeneratedValue(strategy: 'IDENTITY')]
    private ?int $vendorId = null;

    #[Versioned]
    #[Column(name: 'name', type: Types::STRING, length: 255, unique: true)]
    private string $name;

    public function getVendorId(): ?int
    {
        return $this->vendorId;
    }

    public function getName(): string
    {
        return $this->name;
    }

    public function setName(string $name): void
    {
        $this->name = $name;
    }
}
```

I can try to set up a full reproducer next week if this isn't enough to figure it out

simoheinonen:
> The only thing I can think of without building a full reproducer is that the config being passed by reference is causing an unexpected side-effect, adding a `dump($config)` before [this line](https://github.com/mbabker/DoctrineExtensions/blob/922d934f4d1f2f15be92143e5a21521382a0fb3c/src/Revisionable/Mapping/Driver/Attribute.php#L106) would confirm that.

image

mbabker:
Apparently, I can't use the same config key name (versioned) in both extensions since it looks like the `ExtensionMetadataFactory::getExtensionMetadata()` method doesn't pass a "clean" config array into the mapping drivers (no idea if I'd call this a bug or a feature at this point, but here we are).

I'll have to think about ways to improve this because the issue's really a symptom of a bigger design concern overall.

mbabker:
I've changed the config key, so that conflict shouldn't come up anymore.

simoheinonen:
> I've changed the config key, so that conflict shouldn't come up anymore.

Unfortunately doesn't seem to help. Exactly same thing as in https://github.com/doctrine-extensions/DoctrineExtensions/pull/2825#issuecomment-3144761628

Dumping the failing config: 
image

mbabker:
> Unfortunately doesn't seem to help. Exactly same thing as in [#2825 (comment)](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2825#issuecomment-3144761628)

I get what the mapping drivers are trying to do here (warn if you use `#[Versioned]` on a property without the `#[Loggable]` or `#[Revisionable]` attributes), but the only way around that is to introduce an entirely new attribute (similar for XML and (deprecated) YAML)) just to not use the existing config names.  And that is frustrating.

mbabker:
> I wanted to kindly ask if you could share some visibility on the status of this PR. Do you have an idea of when it might be merged, and are there specific blockers preventing it from moving forward?

I've gone as far with this as I can and it really just needs help testing from users.  I don't have merge rights here so I can't personally do anything else with this right now other than keep it up-to-date, but in my mind since the bulk of it is a clone of the loggable extension with the adjustments in the data storage layer (switching from the DBAL's deprecated array type to JSON and switching from PHP serialized values to database normalized values through the MongoDB ODM and DBAL types APIs) it should be pretty much ready to go.  The worst issues this PR's surfaced have been more design quirks with the library overall than bugs in the new extension, so to me that's also a good sign things should be ready to go.

Cafeine42:
Hello, 

It has now been a little over three weeks since we deployed the extension to production, and we have not encountered any issues. 
For certain relationships, we replicate the associated primary keys in JSON columns for traceability purposes only. 

Thank you again for your excellent work.

gdbonino:
I don't know what impact the new extension has on the rest of the library, but if there's no impact, it would be really interesting to reintegrate it.

Kaaly:
Any idea when this extension gonna be released ?
thx !

SebLevDev:
Any idea when this extension gonna be released ?

I'm waiting for the merge; otherwise, I'll have to switch to https://github.com/DamienHarper/auditor-bundle or other repo

Chris53897:
Another (temporary) solution in userland code, is to register `ArrayType` and `ObjectType` as a DoctrineType.
We adjusted this code a bit and i am not sure if both are necessary. 
This makes it possible to update to the newest versions.

```doctrine:
    dbal:
        types:
            array: App\Doctrine\ArrayType
            object: App\Doctrine\ObjectType
```


https://raw.githubusercontent.com/doctrine/dbal/refs/tags/2.13.9/lib/Doctrine/DBAL/Types/ArrayType.php
https://raw.githubusercontent.com/doctrine/dbal/refs/tags/2.13.9/lib/Doctrine/DBAL/Types/ObjectType.php

mbabker:
> Another (temporary) solution in userland code, is to register `ArrayType` and `ObjectType` as a DoctrineType.
> We adjusted this code a bit and i am not sure if both are necessary.
> This makes it possible to update to the newest versions.

You only need the array type to use the Loggable extension.

Kaaly:
Any news on a release date for this feature ?

gdbonino:
Haha, we're all desperately waiting for the release :)

mislavjakopovic:
Great work, looking forward to getting it released soon!

For anyone using Symfony bundle and would like to quickly run this, we've made a small fork as @mbabker suggested.
https://github.com/teamherode/StofDoctrineExtensionsBundle

You just need to update your `composer.json`:
```
    "require": {
        "gedmo/doctrine-extensions": "dev-revisionable as 3.22.0",
        "stof/doctrine-extensions-bundle": "dev-revisionable-v1.15.3 as v1.15.3",
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/mbabker/DoctrineExtensions"
        },
        {
            "type": "vcs",
            "url": "https://github.com/teamherode/StofDoctrineExtensionsBundle"
        }
    ]
```

We'll do our share of testing, and hopefully use this on production soon (even via fork until it's released).

simoheinonen:
@franmomu @phansys

Can we please get some sort of update on this? What needs to happen in order to get this merged and released?

There are a lot of people waiting for a release and this PR has been open for almost two years now.

Any project using `Loggable` is blocked from updating to Symfony 8, since Symfony 8 requires DoctrineBundle 3.x which only supports DBAL 4, and `Loggable` is not compatible with DBAL 4 / ORM 3.

Kaaly:
Personally, I switched to something else.
The impression that things aren't moving forward even though the community is ready to help, there's no feedback.
Too bad, but we were at a point where we couldn't wait any longer.

gdbonino:
> Personally, I switched to something else. The impression that things aren't moving forward even though the community is ready to help, there's no feedback. Too bad, but we were at a point where we couldn't wait any longer.

What are you using now?

Kaaly:
> > Personally, I switched to something else. The impression that things aren't moving forward even though the community is ready to help, there's no feedback. Too bad, but we were at a point where we couldn't wait any longer.
> 
> What are you using now?

https://github.com/rcsofttech85/AuditTrailBundle

gdbonino:
> > > Personally, I switched to something else. The impression that things aren't moving forward even though the community is ready to help, there's no feedback. Too bad, but we were at a point where we couldn't wait any longer.
> > 
> > 
> > What are you using now?
> 
> https://github.com/rcsofttech85/AuditTrailBundle

I've been looking at it too. It's very comprehensive. For now, the only thing that's stopped me from using it is how new it is and the few downloads it's had so far.

microHoffman:
I have been using this PR from @mbabker in my project for last 6+ months and so far it's working good. It would be great to have this functionality merged to the main repo, or if this repo is kinda dead, then create some fork that should be used instead?

PR #2833, opened 2024-06-30
Rewrite the framework integration documentation

mbabker:
Just as advertised in the title.  Best reviewed by [viewing the folder](https://github.com/mbabker/DoctrineExtensions/tree/framework-docs/doc/frameworks).

If anyone's using Laminas and has a better idea for this, please send pull requests.  I'm going very heavily off their documentation and the default config code in each Doctrine module repository having not worked with the framework myself.


doc/frameworks/symfony.md
```diff
@@ -0,0 +1,481 @@
...
+
+## Registering Filters
+
+### Soft Deleteable Filter
+
+When using the [Soft Deleteable](../softdeleteable.md) extension, a filter is available which allows configuring whether
+soft-deleted objects are included in query results.
+
+> [!NOTE]
+> The Symfony bundles default filters to not being enabled, these examples will enable the filters by default.
```
phansys, line 335: The intended message here isn't clear enough to me.
Which bundles are you referring? "doctrine/doctrine-bundle" and "doctrine/mongodb-odm-bundle"?

mbabker: Right.  Basically the `$config->addFilter()` APIs that these configs get transformed into (and same for the Laminas config) only register the filter classes to the object managers, they don't get enabled by default.  The Symfony bundles wire extra calls to enable them globally with their respective `ManagerConfigurator` classes.

phansys:
```suggestion
> The default configuration in the Symfony bundles don't enable the filters. These examples show how to globally enable them.
```

mbabker: Updated, thanks!

doc/frameworks/symfony.md
```diff
@@ -0,0 +1,481 @@
...
+            - [ setAnnotationReader, [ '@annotation_reader' ] ]
+```
+
+## Registering Mapping Configuration
+
+When using the [Loggable](../loggable.md), [Translatable](../translatable.md), or [Tree](../tree.md) extensions, you will
+need to register the mappings for these extensions to your object managers.
+
+> [!NOTE]
+> These extensions only provide mappings through annotations or attributes; if using annotations, you will need to ensure the deprecated [`doctrine/annotations`](https://www.doctrine-project.org/projects/annotations.html) library is installed and enabled.
```
phansys, line 245: AFAIK, the deprecation mentioned here operates on the library implementation, [not on the library itself](https://github.com/doctrine/annotations/blob/94f40ad7ecbc6931958faa8a57c48dce5da2b468/README.md?plain=1#L1-L5).

>⚠️ PHP 8 introduced [attributes](https://www.php.net/manual/en/language.attributes.overview.php), which are a native replacement for annotations. As such, this library is considered feature complete, and should receive exclusively bugfixes and security fixes.

phansys: I understand.
IMO, from our perspective, the important thing in our documentation is our deprecation for the annotation mapping driver, regardless of the library deprecation.
https://github.com/doctrine-extensions/DoctrineExtensions/blob/e85560ed96f977b8c29428a99222cb2ef2f0e80d/README.md?plain=1#L64-L65

mbabker: OK, last commit should (hopefully) make this a bit more clear across all pages I've updated so far.

phansys: Thank you!

franmomu (approved):

doc/frameworks/laminas.md
```diff
@@ -0,0 +1,369 @@
...
+use Psr\Clock\ClockInterface;
+use Psr\Container\ContainerInterface;
+
+return [
+    'service_manager' => [
+        'invokables' => [
+            'gedmo.mapping.driver.attribute' => AttributeReader::class,
+        ],
+        'factories' => [
+            'gedmo.listener.blameable' => function (ContainerInterface $container, string $requestedName): BlameableListener {
```
phansys, line 62: I didn't work with this integration, so this is more a question than a suggestion.

As these functions are not using `$this`, could they be declared as `static`?

phansys (approved):

doc/frameworks/laminas.md
mbabker: …maybe? 😅

I don't use Laminas either, so I just went off the docs and some stuff I saw in the Doctrine modules to see if I got this right enough.  It'd be great if there were a Laminas user that could help verify and improve this, but I'm reasonably sure what's here should at least work.

mbabker: I finally took the time to actually set up a Laminas app and validate the docs page.  What's here works.

Declaring the Closures as static works as well (since they don't use `$this`), but, none of the examples on https://docs.laminas.dev/laminas-servicemanager/v4/configuring-the-service-manager/ document them this way so for the sake of simplicity and to stay aligned with the Laminas docs I'm just going to leave these as-is for now.  Folks can optimize this config by using static closures or implementing factory classes if they so desire, but what's here is enough to get everything installed and functional.

phansys:
Thank you @mbabker!

Issue #2827, opened 2024-06-25
ORM 3 deprecations


zorn-v:
#### Package

<details><summary>show</summary>
<p>

```
$ composer show --latest gedmo/doctrine-extensions

name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.16.0
released : 2024-06-24, this week
latest   : v3.16.0 released 2024-06-24, this week
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : http://gediminasm.org/
source   : [git] https://github.com/doctrine-extensions/DoctrineExtensions.git 7832a3059bdaf14433a2a2515d533c6a16e58520
dist     : [zip] https://api.github.com/repos/doctrine-extensions/DoctrineExtensions/zipball/7832a3059bdaf14433a2a2515d533c6a16e58520 7832a3059bdaf14433a2a2515d533c6a16e58520
path     : /mnt/M.2/Develop/mark/prokat/prokat/vendor/gedmo/doctrine-extensions
names    : gedmo/doctrine-extensions

support
email : gediminas.morkevicius@gmail.com
issues : https://github.com/doctrine-extensions/DoctrineExtensions/issues
source : https://github.com/doctrine-extensions/DoctrineExtensions/tree/v3.16.0
wiki : https://github.com/Atlantic18/DoctrineExtensions/tree/main/doc

autoload
psr-4
Gedmo\ => src/

requires
behat/transliterator ^1.2
doctrine/collections ^1.2 || ^2.0
doctrine/common ^2.13 || ^3.0
doctrine/deprecations ^1.0
doctrine/event-manager ^1.2 || ^2.0
doctrine/persistence ^2.2 || ^3.0
php ^7.4 || ^8.0
psr/cache ^1 || ^2 || ^3
psr/clock ^1
symfony/cache ^5.4 || ^6.0 || ^7.0

requires (dev)
doctrine/annotations ^1.13 || ^2.0
doctrine/cache ^1.11 || ^2.0
doctrine/dbal ^3.2
doctrine/doctrine-bundle ^2.3
doctrine/mongodb-odm ^2.3
doctrine/orm ^2.14.0 || ^3.0
[...]

PR #2834, opened 2024-07-01
[Timestampable] Prevent deprecated ArrayAccess on FieldMapping in doctrine/orm 3

cngJo:
During an upgrade to `doctrine/orm` 3 together with `gedmo/doctrine-extensions` 3.16.1, I noticed a lot of deprecation messages that seem to originate in this package, rather than my application code ;) 

I verified this change by applying a non-backwards compatible version to in my `vendor` directory directly, and the deprecation messages went away.

---

I introduced a small helper method to avoid code duplication around the `instanceof` check. This can be removed when support for `doctrine/orm` 2 is dropped somewhere in the future and replaced with:

```diff
-$this->getType($mapping);
+$mapping->type;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Fixed
+- Prevent depected ArrayAccess on FieldMapping in Timestampable ORM with `doctrine/orm` 3
 
 ## [3.16.1]
 ### Fixed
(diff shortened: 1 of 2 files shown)


cngJo:
Ah sorry ... I just found #2827 after the fact. Feel free to just close this PR if you're not interested in this change 🙂

mbabker:
> Feel free to just close this PR if you're not interested in this change 🙂

The issue isn't necessarily whether we're interested in this type of change or not, but with as widespread as use of the mapping data in this package is, and how it needs to account for at least 3 different shapes (ORM 2.x arrays, ORM 3.x objects, and ODM 2.x arrays), my personal feeling is it just ends up adding a lot of code to this package which affects runtime performance with the only benefit of (trying to) be deprecation-free on a new install/update, especially as similar conditionals have to be introduced in a few dozen places.  That's the reason I just used https://github.com/doctrine-extensions/DoctrineExtensions/issues/2827#issuecomment-2188696396 to not have to think about it in my own applications.

cngJo:
This was fixed with https://github.com/doctrine-extensions/DoctrineExtensions/pull/2889, so this PR is no longer needed.

PR #2836, opened 2024-07-02
[Translatable] Ignore notInsertable columns…

dgrothaus-mc:
in postPersist hook.

Respect the column's metadata and don't attempt to insert a column that's flagged to be not-insertable.

When a custom Translation class configures a field that is flagged as not insertable

```php
#[ORM\Column(
    insertable: false
)]
```

then don't attempt to insert the field in the postPersist hook.

While this works with MySql and MariaDB, inserting a generated column in Postgres throws an error.

```error
cannot insert a non-DEFAULT value
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Fixed
+- Translatable: Ignore notInsertable columns in postPersist hook
 
 ## [3.16.1]
 ### Fixed
--- a/src/Translatable/Mapping/Event/Adapter/ODM.php
+++ b/src/Translatable/Mapping/Event/Adapter/ODM.php
@@ -139,9 +139,15 @@ final class ODM extends BaseAdapterODM implements TranslatableAdapter
         $data = [];
 
         foreach ($meta->getReflectionProperties() as $fieldName => $reflProp) {
-            if (!$meta->isIdentifier($fieldName)) {
-                $data[$meta->getFieldMapping($fieldName)['name']] = $reflProp->getValue($translation);
+            if ($meta->isIdentifier($fieldName)) {
+                continue;
             }
+            $fieldMappings = $meta->getFieldMapping($fieldName);
+            if (($fieldMappings['notSaved'] ?? false) === true) {
+                continue;
+            }
+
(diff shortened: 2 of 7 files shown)


mbabker:
Would you be able to add unit tests covering this change?

Also, the ODM part of this needs an update.  The [`#[Field]`](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/attributes-reference.html#field) attribute uses a `notSaved` property (with that being the key written to the mapping array) compared to the ORM's `insertable` property and `notInsertable` mapping key.

dgrothaus-mc:
> Would you be able to add unit tests covering this change?
> 
> Also, the ODM part of this needs an update. The [`#[Field]`](https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/attributes-reference.html#field) attribute uses a `notSaved` property (with that being the key written to the mapping array) compared to the ORM's `insertable` property and `notInsertable` mapping key.

Sure. I'll add a unit test and force-push the branch with the changes to ODM.

dgrothaus-mc:
Sorry for coming back to you so late. I've been OOO for longer than anticipated.

Could you give me a hint on how to trigger the test for the ODM part?
The ORM part was quite easy. As there's already a test with a custom Translation class https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/tests/Gedmo/Translatable/Fixture/PersonTranslation.php I could add a field to that class to trigger the ORM part of the change.

As far as I can see, there's currently no existing ODM test for a custom Translation Document, that extends https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/src/Translatable/Document/Translation.php
Should I just add a completely new test with all classes required to use a custom Translation Document?

mbabker:
> Should I just add a completely new test with all classes required to use a custom Translation Document?

If you can, that'd be great.  But if it's all too complex (there is a lot of stuff in this package only tested through the ORM), then we can come back around to improving the test coverage later.

It looks like there's already a `TranslatableDocumentTest` and some fixtures in https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/tests/Gedmo/Translatable/Fixture/Document so maybe all we need is a new test case in that class and whatever fixture updates are involved.

PR #2837, opened 2024-07-06
Deprecate annotation-specific properties in the Sluggable extension annotation/attribute classes

mbabker:
When attributes support was added, the way slug handlers were configured was changed between annotations and attributes.  With annotations support now fully deprecated, it doesn't make sense to continue supporting annotation-specific config paths, so this PR deprecates the annotation style slug handler configuration.  Specifically, this will:

- Deprecate the `Slug` attribute's `$handlers` property, slug handlers are now configured through property-level attributes
- Deprecate the `SlugHandlerOption` class entirely, it is unused by the attribute driver in favor of providing the options as an associative array on the `SlugHandler` attribute

Below example (taken right out of the docs) is the same configuration for a field using both annotations and attributes to give a visual reference.

```php
/**
 * @ORM\Column(type="string", unique=true)
 * @Gedmo\Slug(
 *   fields={"title"},
 *   handlers={
 *     @Gedmo\SlugHandler(
 *       class="Gedmo\Sluggable\Handler\TreeSlugHandler",
 *       options={
 *         @Gedmo\SlugHandlerOption(name="parentRelationField", value="category"),
 *         @Gedmo\SlugHandlerOption(name="separator", value="/")
 *       }
 *     )
 *   }
 * )
 */
#[ORM\Column(type: Types::STRING, unique: true)]
#[Gedmo\Slug(fields: ['title'])]
#[Gedmo\SlugHandler(class: TreeSlugHandler::class, options: ['parentRelationField' => 'category', 'separator' => '/'])]
public ?string $slug = null;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ a release.
 
 ## [Unreleased]
 ### Deprecated
-- Annotation-specific mapping parameters for the sluggable extension (#2837)
+- Sluggable: Annotation-specific mapping parameters (#2837)
 
 ### Fixed
 - Fix regression with `doctrine/dbal` >= 4.0 that caused MariaDB to improperly attempt LONGTEXT casting in `TranslationWalker` (issue #2887)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Deprecated
+- Annotation-specific mapping parameters for the sluggable extension (#2837)
+
 ### Fixed
 - Fix regression with `doctrine/dbal` >= 4.0 that caused MariaDB to improperly attempt LONGTEXT casting in `TranslationWalker` (issue #2887)
 
--- a/src/Mapping/Annotation/Slug.php
+++ b/src/Mapping/Annotation/Slug.php
@@ -44,14 +44,20 @@ final class Slug implements GedmoAnnotation
     public string $separator = '-';
     public string $prefix = '';
     public string $suffix = '';
-    /** @var SlugHandler[] */
+
+    /**
+     * @var SlugHandler[]
+     *
+     * @deprecated since gedmo/doctrine-extensions 3.18
+     */
     public $handlers = [];
+
     public string $dateFormat = 'Y-m-d-H:i';
 
     /**
      * @param array<string, mixed> $data
      * @param string[]             $fields
-     * @param SlugHandler[]        $handlers
+     * @param SlugHandler[]        $handlers @deprecated since since gedmo/doctrine-extensions 3.18
      */
     public function __construct(
         array $data = [],
--- a/doc/attributes.md
+++ b/doc/attributes.md
@@ -474,7 +474,8 @@ Optional Parameters:
 
 - **suffix** - An optional suffix for the generated slug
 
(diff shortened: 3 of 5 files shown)


src/Mapping/Annotation/Slug.php
```diff
@@ -44,14 +44,20 @@ final class Slug implements GedmoAnnotation
...
+     */
     public $handlers = [];
+
     public string $dateFormat = 'Y-m-d-H:i';
 
     /**
      * @param array<string, mixed> $data
      * @param string[]             $fields
-     * @param SlugHandler[]        $handlers
+     * @param SlugHandler[]        $handlers @deprecated since since gedmo/doctrine-extensions 3.17
```
phansys, line 60: I didn't use `@deprecated` annotation on top of others annotations. Can you confirm it generates the expected `E_USER_DEPRECATED` error?

mbabker: The `@deprecated` annotation on its own won't trigger a runtime deprecation here, and the attribute driver already ignores this property entirely.  So I think we've got as much as we can get out of this one without adding a check that ends up throwing an exception in the attribute driver if this is a non-empty array.

phansys: Thanks for the explanation.
If this annotation is not used to trigger an `E_USER_DEPRECATED` error, I assume the only purpose here is documentary. If this is the case, I think we should keep the docblock structure as standardized as possible (avoiding more than 1 annotation per line).
Perhaps it is enough to add a description that explains the status of this argument.

mbabker: I just changed it to not use the annotation.  I've seen some projects handle param deprecations that way, but it's still not really a solved problem in PHP (the closest thing I could find in phpDocumentor is https://github.com/phpDocumentor/phpDocumentor/issues/1604 and how to deal with virtual method/property deprecations, but neither it nor any of the WIP PSR's for doc blocks get into this scenario).

phansys (approved):

franmomu (approved):

src/Mapping/Annotation/Slug.php
```diff
@@ -44,14 +44,20 @@ final class Slug implements GedmoAnnotation
     public string $separator = '-';
     public string $prefix = '';
     public string $suffix = '';
-    /** @var SlugHandler[] */
+
+    /**
+     * @var SlugHandler[]
+     *
+     * @deprecated since since gedmo/doctrine-extensions 3.17
```
phansys, line 51:
```suggestion
     * @deprecated since gedmo/doctrine-extensions 3.17
```

phansys (changes requested):

phansys (commented):
Thanks @mbabker!

Could you please add an entry at `CHANGELOG.md` for these deprecations?

mbabker:
> Could you please add an entry at `CHANGELOG.md` for these deprecations?

Added

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ a release.
 ---
 
 ## [Unreleased]
+### Deprecated
+- Annotation-specific mapping parameters for the sluggable extension (#2837)
```
phansys, line 22:
```suggestion
- Sluggable: Annotation-specific mapping parameters (#2837)
```

PR #2851, opened 2024-08-18
Add generics for the event subscriber configuration arrays and adapters

mbabker:
Extracted from #2825

There is already an attempt to document the returned configs and event adapters in the event subscribers using `@method` annotations, but this can be quirky with IDEs (`@method` is read as a virtual method as an example) and there were still some baseline PHPStan issues from the existing setup.  This PR formalizes that through generics by providing templates for the config and adapter interfaces and updating each listener to fill those in (where that info already exists, there are a few extensions where a type for the config still needs to be generated so a generic array is still specified).

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -286,21 +286,6 @@ parameters:
 			path: src/References/ReferencesListener.php
 
 		-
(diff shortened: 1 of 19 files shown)


phansys:
Thanks @mbabker!

PR #2860, opened 2024-08-29
Bump behat/transliterator to ^1.3

Kaaly:
Composer say project is compatible with PHP 7.4 but require `behat/transliterator:^1.2` which is not compatible with PHP 7.4 so bump `behat/transliterator` to ^1.3 min.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "require": {
         "php": "^7.4 || ^8.0",
-        "behat/transliterator": "^1.2",
+        "behat/transliterator": "^1.3",
         "doctrine/collections": "^1.2 || ^2.0",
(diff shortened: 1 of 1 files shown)


mbabker:
Looking at [the diff between 1.2 and 1.3](https://github.com/Behat/Transliterator/compare/v1.2.0...v1.3.0), the only thing that changed was replacing deprecated syntax.  Functionally, there's no real difference between the two versions.

(Not saying there's no reason to accept this, but the older version of the transliterator library (which the odds of getting installed outside of a `composer update --prefer-lowest` type of run should be pretty much zero) would at least work on PHP 7.4)

stof:
Also, `^1.3` and `^1.2` has the same upper bound. So this PR is not adding support for anything. It only removes support for some versions.

PR #2873, opened 2024-09-30
[SoftDeleteable] Make deleted value configurable for SoftDeleteableFilter.php

bartholdbos:


bartholdbos:
@phansys We have added some tests like the existing ones, does this meet the requirements?

mbabker:
You mind describing the use case here?  Also, just on a glance over this PR it seems like a lot more changes than this would be needed considering the extension is designed around using nullable date columns for the deleted value, so widening it to support non-date values in place of null feels like it would need more changes.

You've also only updated the ORM's filter class, the ODM's should be updated as well to keep both implementations working the same way.

As for the tests, new test fixtures should be added which cover the new configuration option and not just adding it to the existing tests.

RubenKluft:
@mbabker Do you consider new test satisfactory?
Otherwise I'd like some guidance on what exactly you'd like to see tested and how.

jaaplallie:
@phansys @mbabker Hello, we assume you guys have a backlog with items to work through, but could you give an indication on how / when we can proceed? Thank you in advance :)

bartholdbos:
@phansys Could you maybe take a look at this? Thanks in advance

PR #2897, opened 2024-11-20
[Tree] Allow uuid as path source in materialized path strategy

andreakeesys:
Fixes the following error in Symfony 7.* during cache warmup, with an entity that implements Tree with the materialized path strategy  and uses Uuidv7 as primary key:

```
In Attribute.php line 260: 

Tree PathSource field - [id] type is not valid. It can be any of the integer 
variants, double, float or string in class - App\Entity\<MyEntity>
```

### DIFF
--- a/tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
+++ b/tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
@@ -1,0 +1,171 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Doctrine Behavioral Extensions package.
+ * (c) Gediminas Morkevicius <gediminas.morkevicius@gmail.com> http://www.gediminasm.org
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Gedmo\Tests\Tree;
+
+use Doctrine\Common\EventManager;
+use Gedmo\Tests\Tool\BaseTestCaseORM;
+use Gedmo\Tests\Tree\Fixture\MPCategoryUuid;
+use Gedmo\Tree\TreeListener;
+use Symfony\Component\Uid\UuidV4;
+
+/**
+ * These are tests for Tree behavior when using Uuid as primary key and TreePathSource.
+ *
+ * @author Gustavo Falco <comfortablynumb84@gmail.com>
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ * @author Andrea Bergamasco <a.bergamasco@keesystem.com>
+ */
+final class MaterializedPathUuidORMTest extends BaseTestCaseORM
+{
+    /**
+     * @var array<string, mixed>
+     */
+    protected array $config;
+
+    protected TreeListener $listener;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->listener = new TreeListener();
+
+        $evm = new EventManager();
+        $evm->addEventSubscriber($this->listener);
+
+        $this->getDefaultMockSqliteEntityManager($evm);
+
+        $meta = $this->em->getClassMetadata(MPCategoryUuid::class);
+        $this->config = $this->listener->getConfiguration($this->em, $meta->getName());
+    }
+
+    public function testInsertUpdateAndRemove(): void
+    {
+        // Insert
+        $category = $this->createCategory();
+        $category->setTitle('1');
+        static::assertNotNull($category->getId());
+        $category2 = $this->createCategory();
+        $category2->setTitle('2');
+        static::assertNotNull($category2->getId());
+        $category3 = $this->createCategory();
+        $category3->setTitle('3');
+        static::assertNotNull($category3->getId());
+        $category4 = $this->createCategory();
+        $category4->setTitle('4');
+        static::assertNotNull($category4->getId());
+
+        $category2->setParent($category);
+        $category3->setParent($category2);
+
+        $this->em->persist($category4);
+        $this->em->persist($category3);
+        $this->em->persist($category2);
+        $this->em->persist($category);
+        $this->em->flush();
+
+        $this->em->refresh($category);
+        $this->em->refresh($category2);
+        $this->em->refresh($category3);
+        $this->em->refresh($category4);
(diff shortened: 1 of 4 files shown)


phansys (commented):
Thank you @KeeSystemOrg!
Is there a way to test the expected behavior in order to avoid regressions?

andreakeesys:
Hi @phansys  I just pushed a PR that contains a new fixture and test class, when removing the change in Validator.php the test breaks with the same error reported in the PR description.

andreakeesys:
Hi @phansys I downgraded Uuid to V4 for PHP 7.4 compatibility and added some CS fixes.

src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
```diff
@@ -72,25 +72,25 @@ public function __construct(DocumentManager $em, UnitOfWork $uow, ClassMetadata
     }
 
     /**
-     * Sets the RepositoryUtilsInterface instance
```
phansys, line 75: Why do we need this change?

phansys (commented):
In order to keep the changes as narrow as possible regarding the PR goal and to ease the review process, please remove the unrelated changes.

src/Tree/Document/MongoDB/Repository/AbstractTreeRepository.php
andreakeesys: Ok! It has been suggested by PHPCS after I rebased main, it was one of the breaking points during CI checks.

andreakeesys: Cleanup done

.docker/php/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-ARG PHP_VERSION=8.2-cli
+ARG PHP_VERSION=8.4-cli
```
phansys, line 3: Could you please perform a rebase?

phansys: `git pull --rebase upstream main && git push --force origin main` should be enough to remove any "duplication" between the changes introduced in this PR, where `upstream` is the remote pointing to https://github.com/doctrine-extensions/DoctrineExtensions and `origin` the remote pointing to https://github.com/KeeSystemOrg/DoctrineExtensions.

andreakeesys: Done, the rebase was not enough to fix, but I amended the duplicated commit that indeed contained my changes and force-pushed it. Now it's feb082ce. Thanks!

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * @param array<int|string, int|string|UuidV4|null> $sources
+     */
+    private function generatePath(array $sources): string
+    {
+        $path = '';
+
+        foreach ($sources as $id) {
+            $path .= $id.$this->config['path_separator'];
```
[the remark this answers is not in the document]
andreakeesys: This depends on `path_ends_with_separator` in the listener configuration, there's also a `path_starts_with_separator` option and both are used in [AbstractMaterializedPath](https://github.com/doctrine-extensions/DoctrineExtensions/blob/09c61fd9149cb1eb3bfb61edb2604e01df8624d2/src/Tree/Strategy/AbstractMaterializedPath.php#L202). I don't think it can be bypassed without side effects.
I updated `generatePath() `to use the same settings retrieved from the listener configuration.

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
```diff
@@ -0,0 +1,172 @@
...
+        return [
+            MPCategoryUuid::class,
+        ];
+    }
+
+    private function createCategory(): MPCategoryUuid
+    {
+        $class = MPCategoryUuid::class;
+
+        return new $class();
```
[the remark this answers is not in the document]
andreakeesys: Thanks, fixed in 0182d2e8

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
```diff
@@ -0,0 +1,174 @@
...
+ * @author Gustavo Falco <comfortablynumb84@gmail.com>
+ * @author Gediminas Morkevicius <gediminas.morkevicius@gmail.com>
+ * @author Andrea Bergamasco <a.bergamasco@keesystem.com>
+ */
+final class MaterializedPathUuidORMTest extends BaseTestCaseORM
+{
+    /**
+     * @var array<string, mixed>
+     */
+    protected $config;
```
phansys, line 32:
```suggestion
    private array $config;
```

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * @var TreeListener
+     */
+    protected $listener;
```
phansys, line 37:
```suggestion
    private TreeListener $listener;
```

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
andreakeesys: Thanks, fixed in e316a1ed

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
andreakeesys: Thanks, fixed in e316a1ed

phansys: Please, remove the docblock. It does not add any value here.

phansys:
Triggering the CI after the merge of #2900.

tests/Gedmo/Tree/MaterializedPathUuidORMTest.php
andreakeesys: Thanks, fixed in 3352422b440732a01220eb8b4b307df878a7886d

phansys:
Thank you @andreakeesys!

PR #2901, opened 2024-11-29
[Uploadable] Remove deprecation warning when calling is_file with a null value

JoolsMcFly:
Hi!

We've been getting lots of deprecation warning (_Deprecated: is_file(): Passing null to parameter #1 ($filename) of type string is deprecated_) coming from `src/Uploadable/UploadableListener::removeFile`:
```php
public function removeFile($filePath)
{
    if (is_file($filePath)) {
        return @unlink($filePath);
    }

     return false;
}
```

This PR simply adds a check for null:
```php
if (!is_null($filePath) && is_file($filePath)) {
```

Thanks for reviewing!

### DIFF
--- a/src/Uploadable/UploadableListener.php
+++ b/src/Uploadable/UploadableListener.php
@@ -389,7 +389,7 @@ class UploadableListener extends MappedEventSubscriber
      */
     public function removeFile($filePath)
     {
-        if (is_file($filePath)) {
+        if (!is_null($filePath) && is_file($filePath)) {
             return @unlink($filePath);
         }
 


mbabker:
Do you have a stack trace showing how this method is being called with null?  The doc block says that only strings should be passed in, so something else is already `_doing_it_wrong()` by calling this method with a null value (and if it's something in this package, that should be updated as well).

JoolsMcFly:
Hi @mbabker .

Here's what we get in Sentry:

image

I guess we could add a check in `postFlush`  as well?

mbabker:
That's helpful as it leads me to two other minor issues:

- The protected `getFilePathFieldValue()` method isn't guaranteed to return a string if it's reading direct from a class property (so that return should be at least `string|null`)
- The protected `addFileRemoval()` method should probably make sure `getFilePathFieldValue()` returns a non-empty value before writing to the `$this->pendingFileRemovals` list

Making changes around those two methods in theory should ensure that the listener isn't calling `removeFile()` with a totally bad parameter.

JoolsMcFly:
Hey, thanks for your comment. Do you want me to make changes you listed in my PR?

mbabker:
Yes, that'd be the more correct fix to make here.

JoolsMcFly:
Hey @mbabker .

Sorry for the delay!

I've implemented what you mentioned but I left the check I originally added as nothing really prevents `removeFile` from being called with a null parameter.

WDYT?

src/Uploadable/UploadableListener.php
```diff
@@ -389,7 +389,7 @@ public function processFile(AdapterInterface $ea, $object, $action)
      */
     public function removeFile($filePath)
     {
-        if (is_file($filePath)) {
+        if (!is_null($filePath) && is_file($filePath)) {
```
phansys, line 392:
```suggestion
        if (null !== $filePath && is_file($filePath)) {
```

PR #2904, opened 2024-12-16
Make the doctrine/common dependency optional

derrabus:
Doctrine ORM has moved away from the Doctrine Common proxy implementation. Yet, this package still forces the installation of doctrine/common.

The only class that we need from doctrine/common is `ClassUtils` which can resolve proxy class names. However, if doctrine/common is not installed, we can be sure that we won't encounter any Common proxies and fall back to `get_class()`. This class introduces a new utility method that does exactly that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,6 @@
         "php": "^7.4 || ^8.0",
         "behat/transliterator": "^1.2",
         "doctrine/collections": "^1.2 || ^2.0",
-        "doctrine/common": "^2.13 || ^3.0",
         "doctrine/deprecations": "^1.0",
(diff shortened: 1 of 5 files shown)


phansys:
Thanks @derrabus!

IIUC, I see other uses that also need to be replaced. By instance:
https://github.com/doctrine-extensions/DoctrineExtensions/blob/389be08a96eb1f3a5aea94f924e25da20eeeeb59/src/Translatable/Mapping/Event/Adapter/ORM.php#L12

phansys (approved):

phansys:
Thank you @derrabus!

PR #2908, opened 2025-01-03
[Timestampable] - Add types to createdAt and updatedAt

giosal:
Based on the requirements of our project to be compatible with PHPStan Level 10, I had to add the types to createdAt and updatedAt, as we're using Timestampable in some of our entities

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ a release.
 
 ## [Unreleased]
 ### New
-- Add return types to createdAt and updatedAt
+- Timestampable: Add types to createdAt and updatedAt
 ### Deprecated
 - Sluggable: Annotation-specific mapping parameters (#2837)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### New
+- Add return types to createdAt and updatedAt
 ### Deprecated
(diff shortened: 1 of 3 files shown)


phansys (changes requested):
Thank you @giosal!

For BC reasons, we can not introduce these changes in a minor version.

If these traits are used alongside an interface covering the same methods, this will force the implementation to be updated.

For the PHPStan level bump, I think we can use #2902 as starting point, allowing the changes here to be implemented in the future.

phansys:
@mbabker, I'd like to hear your thoughts on this.

mbabker:
IMO this is a good 4.0 change but isn't safe to ship as part of a minor release (plus the PR as is right now is incomplete as it doesn't update the ODM trait).

The traits here are purely convenience features and not required for an application to use, so for apps that do need strict type declarations, they can inline this code and update it as needed.

Chris53897:
If this is planned for 4.0, does it make sense to change it to DateTimeImmutable for php type and doctrine?

phansys:
>@phansys if I understood correctly, the changes you request are to update the ODM trait?

As @mbabker said, the ODM implementation should be updated too.

>Otherwise, thank you for your review and your acceptance of the idea even if for a major version

Yes, these changes should be present in the next MAJOR release, as for BC reasons, they can not be introduced in a MINOR release: https://3v4l.org/q0qFe#v7.4.0.

Issue #2843, opened 2024-07-17
Blameable and Frankenphp creates a new entity 


tuxes3:
### Environment
- [FrankenPHP](https://frankenphp.dev/)
#### Package

<details><summary>show</summary>
<p>

```
$ composer show --latest gedmo/doctrine-extensions
name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.15.0
released : 2024-02-12, 5 months ago
latest   : v3.16.1 released 2024-06-25, 3 weeks ago
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : http://gediminasm.org/
source   : [git] https://github.com/doctrine-extensions/DoctrineExtensions.git 2a89103f4984d8970f3855284c8c04e6e6a63c0f
dist     : [zip] https://api.github.com/repos/doctrine-extensions/DoctrineExtensions/zipball/2a89103f4984d8970f3855284c8c04e6e6a63c0f 2a89103f4984d8970f3855284c8c04e6e6a63c0f
path     : /var/www/vendor/gedmo/doctrine-extensions
names    : gedmo/doctrine-extensions

support
email : gediminas.morkevicius@gmail.com
issues : https://github.com/doctrine-extensions/DoctrineExtensions/issues
source : https://github.com/doctrine-extensions/DoctrineExtensions/tree/v3.15.0
wiki : https://github.com/Atlantic18/DoctrineExtensions/tree/main/doc

autoload
psr-4
Gedmo\ => src/

requires
behat/transliterator ^1.2
doctrine/collections ^1.2 || ^2.0
doctrine/common ^2.13 || ^3.0
doctrine/deprecations ^1.0
doctrine/event-manager ^1.2 || ^2.0
doctrine/persistence ^2.2 || ^3.0
php ^7.4 || ^8.0
psr/cache ^1 || ^2 || ^3
psr/clock ^1
symfony/cache ^5.4 || ^6.0 || ^7.0

requires (dev)
doctrine/annotations ^1.13 || ^2.0
doctrine/cache ^1.11 || ^2.0
doctrine/dbal ^3.2
doctrine/doctrine-bundle ^2.3
doctrine/mongodb-odm ^2.3
[...]

PR #2914, opened 2025-02-01
Add support for an actor provider for extensions which use a user reference

mbabker:
Ref: #2843

This is kind of a fresh take on #2074 which is seemingly more relevant with stuff like FrankenPHP out there nowadays.  For now I went with a single interface that can be used by all extensions (since loggable (and the WIP revisionable) have a similar thing), but since the extensions also have different requirements (blameable supports strings and associations, loggable only supports strings), maybe this needs more refinement.  Also for now there aren't any implementations shipped in this library, what goes on within a provider is probably better handled either in the framework integrations or directly within an application.

Draft for now to get feedback on the idea and be able to get docs set up.


stof:
This seems like a good idea.

One benefit (once I implement the actor provider in StofDoctrineExtensionsBundle) is that this will avoid reading the authenticated user early in the request (requiring to potentially start the session), even in requests that won't ever record a blameable change.

src/Tool/ActorProviderInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Gedmo\Tool;
+
+/**
+ * Interface for a provider for an actor for extensions supporting actor/user references.
+ */
+interface ActorProviderInterface
+{
+    /**
+     * @return object|string
```
stof, line 18: should we support returning `null` for the case where we don't have an actor or should we throw ? If throwing, what exception should be used ?

stof: BlameableListener supports `null`, so I suggest using `@return object|string|null`

tests/Gedmo/TestActorProvider.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    private $actor;
+
+    /**
+     * @param object|string $actor
+     */
+    public function __construct($actor)
+    {
+        if (!is_string($actor) && !is_object($actor)) {
+            throw new \InvalidArgumentException(sprintf('The actor must be a string or object, "%s" given', gettype($actor)));
```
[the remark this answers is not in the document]
mbabker: Updated.

src/Tool/ActorProviderInterface.php
mbabker: Yeah I thought about that the other day, null is definitely needed here and it's updated with that.

stof:
@mbabker what is the plan for this PR ? Are you going to merge it ? I got a new report on the bundle that would also be solved by migrating to the actor provider.

mbabker:
I went ahead and did a quick docs update, should be ready for review/merge now.

tests/Gedmo/TestActorProvider.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    private $actor;
+
+    /**
+     * @param object|string|null $actor
+     */
+    public function __construct($actor)
+    {
+        if (!is_string($actor) && !is_object($actor) && null !== $actor) {
+            throw new \TypeError(sprintf('The actor must be a string, an object, or null, "%s" given', gettype($actor)));
```
phansys, line 29:
```suggestion
            throw new \TypeError(sprintf('The actor must be a string, an object, or null, "%s" given.', gettype($actor)));
```

src/Loggable/LoggableListener.php
```diff
@@ -229,6 +242,39 @@ protected function getLogEntryClass(LoggableAdapter $ea, $class)
...
+
+            if (method_exists($actor, 'getUsername')) {
+                return (string) $actor->getUsername();
+            }
+
+            if (method_exists($actor, '__toString')) {
+                return $actor->__toString();
+            }
+
+            throw new InvalidArgumentException(\sprintf('The loggable extension requires the actor provider to return a string or an object implementing the "getUserIdentifier", "getUsername", or "__toString" methods. "%s" cannot be used as an actor.', get_class($actor)));
```
phansys, line 272: Since the actor is not injected as an argument here, I think `UnexpectedValueException` is better for this case.

phansys, line 272:
```suggestion
            throw new InvalidArgumentException(\sprintf('The loggable extension requires the actor provider to return a string or an object implementing the "getUserIdentifier()", "getUsername()", or "__toString()" methods. "%s" cannot be used as an actor.', get_class($actor)));
```

mbabker: Updated.

mbabker: Updated while rebasing.

tests/Gedmo/TestActorProvider.php
mbabker: Updated while rebasing.

phansys (approved):

phansys:
Thank you @mbabker!

phansys:
@stof, this was released under [3.19.0](https://github.com/doctrine-extensions/DoctrineExtensions/releases/tag/v3.19.0).

Thanks!

PR #2924, opened 2025-02-12
Bump CS-Fixer to latest minor, update config

mbabker:
PHP-CS-Fixer doesn't need to have an extremely wide range of supported versions, so this updates the tool to require the latest minor release to close that gap down significantly.  I've also updated the config to:

- Enable the experimental parallel runner, which with 800 files being processed doesn't seem like a terrible thing
- Removed the explicit `@PHP71Migration` rules as they're implied in the `@PHP74` rules
- Updated the `@PHPUnitMigration` rule from 8.4 to 9.1 which ensures certain renamed assertions use the newer name (pretty sure this was already done anyway since there weren't any changes locally after this)
- Updated the `nullable_type_declaration_for_default_null_value` rule to stop using deprecated configuration

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -28,13 +28,12 @@ $finder = PhpCsFixer\Finder::create()
     ]);
 
 return (new PhpCsFixer\Config())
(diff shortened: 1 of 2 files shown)


phansys:
Thank you @mbabker!

PR #2925, opened 2025-02-15
[Sluggable] Use public property instead of deprecated array access

odolbeau:
This property is already available in doctrine/orm 2.14.0 which is the lowest supported version ([see here](https://github.com/doctrine/orm/blob/2.14.0/lib/Doctrine/ORM/Mapping/Embedded.php#L22)) so we can easily use it instead of [the deprecated array access](https://github.com/doctrine/orm/pull/11211).

### DIFF
--- a/src/Sluggable/Mapping/Driver/Attribute.php
+++ b/src/Sluggable/Mapping/Driver/Attribute.php
@@ -82,7 +82,7 @@ class Attribute extends AbstractAnnotationDriver
         // Embedded entity
         if (property_exists($meta, 'embeddedClasses') && $meta->embeddedClasses) {
             foreach ($meta->embeddedClasses as $propertyName => $embeddedClassInfo) {
-                $embeddedClass = new \ReflectionClass($embeddedClassInfo['class']);
+                $embeddedClass = new \ReflectionClass($embeddedClassInfo->class);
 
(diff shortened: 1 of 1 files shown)


mbabker:
This one's not quite the same thing.  That `Embedded` attribute isn't what's being read here, it's the [`EmbeddedClassMapping` DTO](https://github.com/doctrine/orm/blob/3.0.0/src/Mapping/EmbeddedClassMapping.php) and that was introduced in 3.0 only.

odolbeau:
Oups! Sorry for the noise! :see_no_evil: 
By curiosity, what's the reason to support so many versions?

mbabker:
ORM 2.x and 3.x are still both maintained upstream, and a lot of the code also "just works" for supporting the MongoDB ODM.  So even closing the version ranges isn't exactly simplifying too much in this library's code in all honesty (especially with ORM 3.x having DTOs and the MongoDB ODM having arrays in the metadata APIs), I just did it in one PR and without flat out dropping ORM 2.x (which IMO would be a mistake and the resources aren't here to maintain two branches with support for both versions) there really isn't much else to gain.

odolbeau:
This is perfectly clear, thanks for your time and for this lib! :pray:

PR #2928, opened 2025-02-26
Add an IP address provider for extensions referencing IP addresses

### LINKED ISSUE #2927: [IpTraceable] Add a provider API to get the current IP
## Feature Request

The ActorProviderInterface solved the issue with frankenphp where the PHP process can handle multiple requests for different users in the same PHP process (which might leak a value set with `setUserValue` previously).
The IpTraceable extension suffers from the same kind of leaks for the current IP. It would be great to have a similar provider interface for the current IP to solve this case.

mbabker:
Fixes #2927

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- IP address provider for use with extensions with IP address references (#2928)
 
 ## [3.19.0] - 2025-02-24
 ### Added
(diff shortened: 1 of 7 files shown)


stof (approved):

phansys (approved):

phansys:
Thank you @mbabker!

stof:
@mbabker can you do a new release containing that feature ? This would allow me to require that new version as the min version of the extensions in my bundle to migrate to that feature (solving the compat with the worker mode of frankenphp)

PR #2929, opened 2025-03-06
Add support for blameable on remove

oojacoboo:
This PR adds support for blameable on `remove`/deletion of an entity.  This can be very useful, when used in conjunction with soft-delete.  It obviously is useless without soft-delete.

Docs have been updated a tests added.


src/AbstractTrackingListener.php
```diff
@@ -48,6 +48,7 @@ public function getSubscribedEvents()
             'prePersist',
             'onFlush',
             'loadClassMetadata',
+            'preRemove',
```
mbabker, line 51: Adding the subscribed event here means that the IP traceable and timestampable listeners are also getting "on remove" support.  Is this intended?

doc/blameable.md
```diff
@@ -4,10 +4,19 @@ The **Blameable** behavior automates the update of user information on your Doct
 
 ## Index
 
-- [Getting Started](#getting-started)
-- [Configuring Blameable Objects](#configuring-blameable-objects)
-- [Using Traits](#using-traits)
-- [Logging Changes For Specific Actions](#logging-changes-for-specific-actions)
+- [Blameable Behavior Extension for Doctrine](#blameable-behavior-extension-for-doctrine)
```
mbabker, line 7: Just a personal suggestion, but I'd leave the index as it was where only the top level headers are listed.  Since this project isn't using an external docs tool and they're being rendered in GitHub, GitHub's native outline feature keeps this a little more maintainable (GitHub will automatically outline/index the whole document, all we have to do is link to the main sections).

src/AbstractTrackingListener.php
oojacoboo: Makes sense, and no that was necessarily intentional.  Is that an issue though?

mbabker: It could be, as this change implies all extensions (the ones in this library and in the off chance someone wrote their own) using this listener need to support the Doctrine `preRemove` event.  But as is, there isn't a way to configure that for the other extensions.  So if the intent is to only add it to blameable, I'd just do a `public function getSubscribedEvents() { return array_merge(parent::getSubscribedEvents(), ['preRemove']; }` override to not have to worry about it.

doc/blameable.md
oojacoboo: Reverted the TOC back to a simplified version

oojacoboo:
Couple things here:

- I don't understand why PHP 7.x is being supported still.  Why can't people that need 7.4 support use an older version of this lib?  How long do you guys intend on trying to continue to support it?  It makes it difficult trying to properly type this lib when supporting such a old version.  [PHP isn't even supporting PHP 8.0 anymore](https://www.php.net/supported-versions.php).  Annotations (only attributes supported) should be entirely dropped from this lib IMO, for all future versions.
- What's going on with the PHPStan errors?  I didn't make changes to these files.

phansys:
>What's going on with the PHPStan errors? I didn't make changes to these files.

You are [removing `@phpstan-template` annotations](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2929/files#diff-549857e13d9ecdad914283e2d364dd791db3a55328d207516ffb5f823a4e3257L44), so errors [like this](https://github.com/doctrine-extensions/DoctrineExtensions/actions/runs/13907238607/job/39320813376?pr=2929#step:5:15) are expected:

```
Error: Generic type Gedmo\Mapping\MappedEventSubscriber<array, Gedmo\Mapping\Event\AdapterInterface> in PHPDoc tag @extends specifies 2 template types, but class Gedmo\Mapping\MappedEventSubscriber supports only 1: TConfig
```

oojacoboo:
@phansys I've resolved the PHPStan issues.  I'm a bit confused with what's going on with PHPStan on this lib, though.  There is a `phpstan-baseline.neon` file that's over 1200 lines of ignores.  Also, does PHPStan not read the default phpdoc annotations now (`@param`, `@return`, etc)?

Also, IMO, it'd be a good idea to drop support for some of these unsupported versions of PHP.  It's holding back to the lib and making updates and codestyle much more difficult.

Please let me know if anything else is needed here - should be ready.

oojacoboo:
@mbabker @phansys Can we get this merged?

oojacoboo:
@phansys I've merged in the latest changes.  There weren't any conflicts though, so it wasn't actually necessary.  But that's been done.  I'm unsure about this failing test though.  That seems to be preexisting, unless I'm misunderstanding something.

PR #2930, opened 2025-03-06
Resolved a bug where a soft-deleted object isn't remove from the ObjectManager

oojacoboo:
Currently, after you remove an entity, it's still queryable by the pk, as it's cached in the `ObjectManager`.  This PR fixes this bug by removing it `postFlush`.  See test for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ a release.
 
 ## [Unreleased]
 
+### Fixed
+- SoftDeleteable: Resolved a bug where a soft-deleted object isn't remove from the ObjectManager (#2930)
+
 ## [3.19.0] - 2025-02-24
 ### Added
 - Actor provider for use with extensions with user references (#2914)
@@ -103,7 +106,7 @@ a release.
 - Dropped support for doctrine/dbal < 3.2
 
 ### Deprecated
-- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager()` and `getObject()` on EventArgs that do not implement `getObjectManager()` and `getObject()` (such as old EventArgs implementing `getEntityManager()` and `getEntity()`) 
+- Calling `Gedmo\Mapping\Event\Adapter\ORM::getObjectManager()` and `getObject()` on EventArgs that do not implement `getObjectManager()` and `getObject()` (such as old EventArgs implementing `getEntityManager()` and `getEntity()`)
 - Calling `Gedmo\Uploadable\Event\UploadableBaseEventArgs::getEntityManager()` and `getEntity()`. Call `getObjectManager()` and `getObject()` instead.
 
 ## [3.13.0] - 2023-09-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ a release.
 ---
 
 ## [Unreleased]
-
 ### Fixed
 - SoftDeleteable: Resolved a bug where a soft-deleted object isn't remove from the ObjectManager (#2930)
 
--- a/src/SoftDeleteable/SoftDeleteableListener.php
+++ b/src/SoftDeleteable/SoftDeleteableListener.php
@@ -51,6 +51,14 @@ class SoftDeleteableListener extends MappedEventSubscriber
      */
     public const POST_SOFT_DELETE = 'postSoftDelete';
 
+
+    /**
+     * Objects soft-deleted on flush.
+     *
+     * @var array<object>
+     */
+    private array $softDeletedObjects = [];
+
     /**
      * @return string[]
      */
@@ -59,6 +67,7 @@ class SoftDeleteableListener extends MappedEventSubscriber
         return [
             'loadClassMetadata',
             'onFlush',
+            'postFlush',
         ];
     }
 
@@ -102,7 +111,7 @@ class SoftDeleteableListener extends MappedEventSubscriber
 
                     $evm->dispatchEvent(
                         self::PRE_SOFT_DELETE,
-                        $preSoftDeleteEventArgs
+                        $preSoftDeleteEventArgs,
                     );
                 }
 
@@ -129,7 +138,26 @@ class SoftDeleteableListener extends MappedEventSubscriber
                         $postSoftDeleteEventArgs
                     );
                 }
+
+                $this->softDeletedObjects[] = $object;
             }
+        }
+    }
+
+    /**
+     * Detach soft-deleted objects from object manager.
+     *
+     * @param EventArgs $args
+     *
+     * @return void
+     */
+    public function postFlush(EventArgs $args)
+    {
+        $ea = $this->getEventAdapter($args);
+        $om = $ea->getObjectManager();
+        foreach ($this->softDeletedObjects as $index => $object) {
+            $om->detach($object);
+            unset($this->softDeletedObjects[$index]);
         }
     }
 
--- a/src/SoftDeleteable/SoftDeleteableListener.php
+++ b/src/SoftDeleteable/SoftDeleteableListener.php
@@ -147,8 +147,6 @@ class SoftDeleteableListener extends MappedEventSubscriber
     /**
      * Detach soft-deleted objects from object manager.
      *
-     * @param EventArgs $args
-     *
      * @return void
      */
     public function postFlush(EventArgs $args)
--- a/src/SoftDeleteable/SoftDeleteableListener.php
+++ b/src/SoftDeleteable/SoftDeleteableListener.php
@@ -51,7 +51,6 @@ class SoftDeleteableListener extends MappedEventSubscriber
      */
     public const POST_SOFT_DELETE = 'postSoftDelete';
 
-
     /**
      * Objects soft-deleted on flush.
      *
--- a/tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
+++ b/tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
@@ -522,6 +522,42 @@ final class SoftDeleteableEntityTest extends BaseTestCaseORM
         static::assertCount(0, $data);
     }
 
+    public function testSoftDeletedObjectIsRemovedPostFlush(): void
+    {
+        $repo = $this->em->getRepository(Article::class);
+        $commentRepo = $this->em->getRepository(Comment::class);
+
+        $comment = new Comment();
+        $commentValue = 'Comment 1';
+        $comment->setComment($commentValue);
+
+        $art0 = new Article();
+        $field = 'title';
+        $value = 'Title 1';
+        $art0->setTitle($value);
+        $art0->addComment($comment);
+
+        $this->em->persist($art0);
+        $this->em->flush();
+
+        $art = $repo->findOneBy([$field => $value]);
+
+        $this->assertNull($art->getDeletedAt());
+        $this->assertNull($comment->getDeletedAt());
+        $this->assertCount(1, $art->getComments());
+
+        $this->em->remove($comment);
+
+        // The Comment has been marked for removal, but not yet flushed.  This means the
+        // Comment should still be available.
+        $this->assertInstanceOf(Comment::class, $commentRepo->find($comment->getId()));
+
+        $this->em->flush();
+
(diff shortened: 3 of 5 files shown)


oojacoboo:
@phansys it'd be nice if the tests could be run on new commits.  Also, I believe the CHANGELOG will need to be something you commit, as I don't have the target version.

phansys:
>@phansys it'd be nice if the tests could be run on new commits.

Tests are running.

>Also, I believe the CHANGELOG will need to be something you commit, as I don't have the target version.

Please, read the [Changelog](https://github.com/doctrine-extensions/DoctrineExtensions/blob/ecb4b8b87df823fdf169e36f13ef2ba0fd0d2637/CONTRIBUTING.md#changelog) section at `CONTRIBUTING.md`.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ a release.
 
 ## [Unreleased]
```
phansys, line 21:
```suggestion
```

tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
```diff
@@ -522,6 +522,42 @@ public function testShouldFilterBeQueryCachedCorrectlyWhenToggledForEntity(): vo
...
+
+        $art = $repo->findOneBy([$field => $value]);
+
+        static::assertNull($art->getDeletedAt());
+        static::assertNull($comment->getDeletedAt());
+        static::assertCount(1, $art->getComments());
+
+        $this->em->remove($comment);
+
+        // The Comment has been marked for removal, but not yet flushed.  This means the
```
phansys, line 551:
```suggestion
        // The Comment has been marked for removal, but not yet flushed. This means the
```

tests/Gedmo/SoftDeleteable/SoftDeleteableEntityTest.php
```diff
@@ -522,6 +522,42 @@ public function testShouldFilterBeQueryCachedCorrectlyWhenToggledForEntity(): vo
...
+        // Comment should still be available.
+        static::assertInstanceOf(Comment::class, $commentRepo->find($comment->getId()));
+
+        $this->em->flush();
+
+        // Now that we've flushed, the Comment should no longer be available and should return null
```
phansys, line 557:
```suggestion
        // Now that we've flushed, the Comment should no longer be available and should return null.
```

src/SoftDeleteable/SoftDeleteableListener.php
```diff
@@ -129,10 +137,27 @@ public function onFlush(EventArgs $args)
                         $postSoftDeleteEventArgs
                     );
                 }
+
+                $this->softDeletedObjects[] = $object;
```
phansys, line 141: More than a suggestion, this is a question or a request for analysis.

Given storing the list of deleted objects in this array could lead to high memory consumption, do you evaluated a way to mitigate this by using the object references (like `spl_object_id(), `spl_object_hash()`, or a combination of the model class and its id)?

I don' t know if this can produce better results, but I think this is a consideration we should take into account.

oojacoboo: Being that these objects are unset (cleared from memory) `postFlush`, I really don't see it as much of a concern - certainly not enough to introduce `spl_object_id` complexity and the potential to expose bugs through that.

phansys (commented):
@mbabker, could you please provide your feedback here?

Thanks.

mbabker:
I don't use soft deletes in my Doctrine based projects so at face value this seems fine to me.

What's the current behavior if you do a `$em->getRepository(SoftDeleteable::class)->find($pk);` when the soft-deleted record hasn't been previously loaded?  I'm not 100% sure where the query filter is supposed to kick in but if calling `find()` when a soft-deleted record has never been loaded into the object manager is supposed to return null, then this would definitely fix the bug of it returning a soft-deleted record because the object manager is aware of it.

src/SoftDeleteable/SoftDeleteableListener.php
mbabker: These models shouldn't be held in memory for too long anyway short of something catastrophic happening between the `onFlush` event and the `postFlush` event.  So it really boils down to what happens if you try re-using a closed object manager (since the ORM's UoW is closing the EM and rolling back the active transaction on fail at least).

oojacoboo:
@mbabker You're correct in your assumption.  If it's not already cached in the `ObjectManager`, it would return `null`.  This fixes a bug around the cache.  We've been using it in production for years on a fork of this lib.  I'm trying to get this lib caught up with all the patches we've made over the years, now that the `ActorProvider` implementation is in place.

phansys:
@oojacoboo, could you please rebase this PR and fix the conflicts?

oojacoboo:
The PR says there aren't any conflicts with the base branch.

phansys:
>The PR says there aren't any conflicts with the base branch.


This is what I get from the GitHub UI:

image

oojacoboo:
image

Why would you be rebasing it and not merging?  You can squash and merge.

phansys:
Thank you @oojacoboo!

phansys:
>Why would you be rebasing it and not merging? You can squash and merge.

Good point. Squash was working with no conflicts.
Thanks!

rotdrop:
> Currently, after you remove an entity, it's still queryable by the pk, as it's cached in the `ObjectManager`. This PR fixes this bug by removing it `postFlush`. See test for more details.

One question: what about the soft-deleteable filter? Should the object be removed from the ObjectManager if the soft-deleteable filter is *disabled*? Shouldn't it stay in the cache in that case?

oojacoboo:
@rotdrop if that filter is disabled... that'd be a real database delete, no?  Why would it remain in the`ObjectManager` if you deleted it?  If you need access to values post-delete, you should assign them to variables prior.

rotdrop:
> @rotdrop if that filter is disabled... that'd be a real database delete, no? Why would it remain in the`ObjectManager` if you deleted it? If you need access to values post-delete, you should assign them to variables prior.

No, I think you misunderstood my point. The "soft-deleteable" filter is the filter which actually *enables* filtering out soft-deleted entities. If that filter is disabled (and it is disabled by default) then select statements will just retrieve entities as usual. Only if the soft-deleteable entity filter is enabled then soft-deleted entities are omitted from the result set.

So by "the filter" I do *not* mean the SoftDeleteable extension, but the soft-deleteable entity filter, please compare with the documentation here:

https://github.com/doctrine-extensions/DoctrineExtensions/blob/ea1d37586b8e4bae2a815feb38b177894b12c44c/doc/softdeleteable.md?plain=1#L32

Hence: *without* the filter being enabled the entities show IMHO stay in the cache.

rotdrop:
> @rotdrop if that filter is disabled... that'd be a real database delete, no? Why would it remain in the`ObjectManager` if you deleted it? If you need access to values post-delete, you should assign them to variables prior.

@oojacoboo Perhaps this PR #2939 clarifies what I am talking about.

oojacoboo:
I see what you mean now - regarding the filter and being enabled/disabled.

Honestly, I don't see the difference.  I get where you're coming from and I see your logic, but I think it's more of a personal use case, or a flawed way of looking at it.

Deleting an entity, regardless of whether soft-delete is enabled/disabled or even setup in the first place, should remove the object from the ObjectManager.  You're deleting it... that's an explicit intention.  Leaving it in the `ObjectManager` is a flawed side-effect.

That said, if you have the `soft-deleteable` filter disabled, your subsequent call will just fetch it again.  I do not think that's a reason to leave it in the `ObjectManager` though.  The filter is a query filter, meaning it's purpose is to affect the results of a "fetch"/query on the database, and the results of that.  It's not intended to alter your state or change the behavior of state.

rotdrop:
@oojacoboo I do not think that the scenario is that clear. Also, if we talk about state, explicit intent and that calling remove should remove the entity from the object manager: this still does not happen until you call flush(). However, flush() is meant to sync the state of the object manager with the database, flush() is not a cache garbage collector.

Also, this  entire SoftDeleteable extension already breaks "explicit intent", actually, the whole purpose of this plugin is to alter the meaning of remove(). So, I am not convinced that the way this PR #2930 handles things is desirable.

oojacoboo:
`flush` should also be seen as an abstraction for the `EntityManager`, to flush/update state.  The `EntityManager` is abstracted from the DBAL layer already.

I won't argue that `SoftDeletable` is a broken antipattern.  That's been well discussed, and I'm not going to disagree.  However, many people are stuck using it.  But that's no reason to try and increase cohesion.

bobvandevijver:
I'm torn by this change. From one point of view I agree with the reasoning behind this change, but from another point I do not. Simply detaching the soft-deleted entity during a request/message context seems to introduce more issues than what it solves.

As it is required to call `flush()` to get the softdeletable entity state updated (to set the deleted at time and possibly the actor), you are forced to take that route if you need those values for further handling down the line. However, with this change that is no longer possible, which forces you to first disable the filter again to retrieve the same object that was already there. My use case here is to create an additional log entry after the object has been removed, which references the removed object.

In our projects we have acted on knowing that a softdeleted object is still available within the same request/message context, as you know it won't be fully deleted in the end.

Next to that, issues #2942, #2947 & #2955 seem to be directly caused by this change. Especially #2955 will be almost impossible to fully solve.

So, I believe this needs to be either reverted to solves the issues that have risen, or to make this behaviour configurable.

<details>
<summary>If anyone is interested, here is a compiler pass to remove the `postFlush` handler when using the stof bundle.</summary>

```php
use Exception;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;

class DisablePostFlushDetachCompilerPass implements CompilerPassInterface
{

  const TAG_NAME = 'doctrine.event_listener';

  public function process(ContainerBuilder $container): void
  {
    $definition = $container->getDefinition('stof_doctrine_extensions.listener.softdeleteable');
    if (!$definition->hasTag(self::TAG_NAME)) {
      throw new Exception('Doctrine event listener not configured?');
    }

    // Remove the postFlush tag
    $tags = $definition->getTags();
    $tags[self::TAG_NAME] = array_filter(
      $tags[self::TAG_NAME],
      static fn (array $tag): bool => ($tag['event'] ?? null) !== 'postFlush'
    );

    // Update the listener tags
    $definition->clearTags();
    $definition->setTags($tags);
  }
}

// Register in your bundle with the following:
// Needs to be run before RegisterEventListenersAndSubscribersPass is run
// $container->addCompilerPass(new DisablePostFlushDetachCompilerPass(),  priority: 10);
```

</details>

rotdrop:
> I'm torn by this change. From one point of view I agree with the reasoning behind this change, but from another point I do not. Simply detaching the soft-deleted entity during a request/message context seems to introduce more issues than what it solves.
> 
> As it is required to call `flush()` to get the softdeletable entity state updated (to set the deleted at time and possibly the actor), you are forced to take that route if you need those values for further handling down the line. However, with this change that is no longer possible, which forces you to first disable the filter again to retrieve the same object that was already there. My use case here is to create an additional log entry after the object has been removed, which references the removed object.
> 
> In our projects we have acted on knowing that a softdeleted object is still available within the same request/message context, as you know it won't be fully deleted in the end.
> 
> Next to that, issues #2942, #2947 & #2955 seem to be directly caused by this change. Especially #2955 will be almost impossible to fully solve.
> 
> So, I believe this needs to be either reverted to solves the issues that have risen, or to make this behaviour configurable.
> If anyone is interested, here is a compiler pass to remove the `postFlush` handler when using the stof bundle.

Also I am still not convinced, although @oojacoboo had the last word so far. But though I did not answer him yet I really do not agree with his last comment in this affair.

So here is a feature request: please revert #2930

Please let us start a discussion and let us come to a better documented agreement about how "soft-deleteable" should behave.

bobvandevijver:
@kopfsalat Yeah, sounds related to this change. You can confirm it by using the CompilerPass I posted in my previous comment here: if using that solves your issue you know for sure.

oojacoboo:
@kopfsalat Have you tried disabling the query filter before the second flush that's attempting to query the database for that parent record?

kopfsalat:
@oojacoboo Hi, I tried it now and it does not prevent the error.

Issue #2582, opened 2023-02-15
[Tree][NestedSet] Invalid tree when multiple roots are inserted in one flush


sgehrig:
Using the following simple Tree

```php
#[ORM\Entity(repositoryClass: OURepository::class)]
#[ORM\Table(name: 'ous')]
#[ORM\Index(columns: ['lft', 'rgt'], name: 'idx_tree')]
#[Gedmo\Tree(type: 'nested')]
class OU
{
    #[ORM\Id]
    #[ORM\Column('id', 'guid')]
    private string $id;

    #[ORM\ManyToOne(targetEntity: OU::class, inversedBy: 'children')]
    #[ORM\JoinColumn(name: 'parent', referencedColumnName: 'id', nullable: true, onDelete: 'CASCADE')]
    #[Gedmo\TreeParent]
    private ?self $parent = null;

    #[ORM\Column(name: 'lft', type: 'integer', options: ['unsigned' => true])]
    #[Gedmo\TreeLeft]
    private int $left = 1;

    #[ORM\Column(name: 'lvl', type: 'integer', options: ['unsigned' => true])]
    #[Gedmo\TreeLevel]
    private int $level = 0;

    #[ORM\Column(name: 'rgt', type: 'integer', options: ['unsigned' => true])]
    #[Gedmo\TreeRight]
    private int $right = 2;

    #[ORM\OneToMany(mappedBy: 'parent', targetEntity: OU::class)]
    #[ORM\OrderBy(['left' => 'ASC'])]
    private Collection $children;

    public function __construct(string $id, ?self $parent = null)
    {
        $this->id = $id;
        $this->children = new ArrayCollection();
        $this->parent = $parent;
        if ($parent) {
            $parent->children->add($this);
        }
    }
}
```

I create multiple roots in one `flush()`:

```php
$ou1 = new OU('id1', null);
$ou11 = new OU('id11, $ou1);
$ou2 = new OU('id2', null);
$ou21 = new OU('id21, $ou2);
```

Now there's a corrupt tree in the database:

| id   | parent | lft | lvl | rgt |
|------|--------|-----|-----|-----|
| id1  |        | 1   | 0   | 8   |
| id11 | id1    | 2   | 1   | 7   |
| id2  |        | 3   | 0   | 6   |
| id21 | id2    | 4   | 1   | 5   |

If I split the two roots into two separate `flush()`s, the tree is valid:

| id   | parent | lft | lvl | rgt |
[...]

PR #2932, opened 2025-03-18
[Tree][NestedSet] Fix inserting multiple roots in one flush

sgehrig:
This PR tries to solve the issue described in #2582. Currently it only provides a test case that reproduces the issue and the observations made in https://github.com/doctrine-extensions/DoctrineExtensions/issues/2582#issuecomment-1961459995 (persisting roots first) and in https://github.com/doctrine-extensions/DoctrineExtensions/issues/2582#issue-1585812779 (doing two flushes).

`\Gedmo\Tests\Tree\Issue\Issue2582Test::testInsertTwoRootsInOneFlush` fails as expected right now because it is a direct replication of the issue. `\Gedmo\Tests\Tree\Issue\Issue2582Test::testInsertTwoRootsInOneFlushRootsFirst` and `\Gedmo\Tests\Tree\Issue\Issue2582Test::testInsertTwoRootsInTwoFlushes` both succeed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ a release.
 ### Changed
 - All: Removed the dollar sign from the generated cache ID for extension metadata to ensure only characters mandated by [PSR-6](https://www.php-fig.org/psr/psr-6/#definitions) are used, improving compatibility with caching implementations with strict character requirements (#2978)
 
+### Fixed
+- Tree: Fixed inserting multiple root nodes in one flush operation with the nested set strategy in certain circumstances (#2582)
+
 ## [3.22.0] - 2025-12-13
 ### Added
 - Support for Symfony 8
--- a/src/Tree/Strategy/ORM/Nested.php
+++ b/src/Tree/Strategy/ORM/Nested.php
@@ -407,6 +407,11 @@ class Nested implements Strategy
                 $wrapped->setPropertyValue($config['right'], $right);
             }
             $newRoot = $parentRoot;
+
+            if (!isset($this->treeEdges[$meta->getName()])) {
+                $this->treeEdges[$meta->getName()] = $this->max($em, $config['useObjectClass'], $newRoot) + 1;
+            }
+            $this->treeEdges[$meta->getName()] += 2;
         } elseif (!isset($config['root'])
             || ($meta->isSingleValuedAssociation($config['root']) && null !== $parent && ($newRoot = $meta->getFieldValue($node, $config['root'])))) {
             if (!isset($this->treeEdges[$meta->getName()])) {
--- a/tests/Gedmo/Tree/Issue/Issue2582Test.php
+++ b/tests/Gedmo/Tree/Issue/Issue2582Test.php
@@ -57,7 +57,7 @@ class Issue2582Test extends BaseTestCaseORM
                 $expected[$i],
                 [
                     $a->getId(),
-                    $a->getParent()?->getId(),
+                    $a->getParent() ? $a->getParent()->getId() : null,
                     $a->getLeft(),
                     $a->getLevel(),
                     $a->getRight(),
@@ -93,7 +93,7 @@ class Issue2582Test extends BaseTestCaseORM
                 $expected[$i],
                 [
                     $a->getId(),
-                    $a->getParent()?->getId(),
+                    $a->getParent() ? $a->getParent()->getId() : null,
                     $a->getLeft(),
                     $a->getLevel(),
                     $a->getRight(),
@@ -130,7 +130,7 @@ class Issue2582Test extends BaseTestCaseORM
                 $expected[$i],
                 [
                     $a->getId(),
-                    $a->getParent()?->getId(),
+                    $a->getParent() ? $a->getParent()->getId() : null,
                     $a->getLeft(),
                     $a->getLevel(),
                     $a->getRight(),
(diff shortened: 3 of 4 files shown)


sgehrig:
The change in `\Gedmo\Tree\Strategy\ORM\Nested` fixes the issue and keeps all tests *green*. I am not sure however if this is in any shape or form correct.

When debugging the failing test case I noticed that the `\Gedmo\Tree\Strategy\ORM\Nested::$treeEdges` property is not updated when `\Gedmo\Tree\Strategy\ORM\Nested::updateNode` handles the second non-root node.

sgehrig:
@phansys Is there anything I need to do to get this one merged? Thanks a lot for your help.

phansys:
Thanks for your contribution @sgehrig! 

Could you please add the corresponding changelog entries? See https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/CONTRIBUTING.md#changelog.

sgehrig:
Is there any chance to have this merged?

sgehrig:
Branch has been rebased. 
Is there any chance to have this one merged?

sgehrig:
Still a valid bug.

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ a release.
 ### Changed
 - All: Removed the dollar sign from the generated cache ID for extension metadata to ensure only characters mandated by [PSR-6](https://www.php-fig.org/psr/psr-6/#definitions) are used, improving compatibility with caching implementations with strict character requirements (#2978)
 
+### Fixed
+- Tree: Fixed inserting multiple root nodes in one flush operation with the nested set strategy in certain circumstances (#2582)
```
phansys: Thanks for the addition @sgehrig!

Please, replace "certain circumstances" with the concrete cases considered in this PR to unambiguously clarify the scenario where this fix applies.
This way, users receiving this update will be able to determine whether or not their own applications are affected by the bug.

sgehrig: @phansys Done.

src/Tree/Strategy/ORM/Nested.php
```diff
@@ -406,6 +406,9 @@ public function updateNode(EntityManagerInterface $em, $node, $parent, $position
                 $wrapped->setPropertyValue($config['right'], $right);
             }
             $newRoot = $parentRoot;
+
+            $this->treeEdges[$meta->getName()] ??= $this->max($em, $config['useObjectClass'], $newRoot) + 1;
+            $this->treeEdges[$meta->getName()] += 2;
```
phansys: This advances the edge counter by a literal `2`, but this branch opens a gap of `$treeSize` (see the `shiftRL($em, $config['useObjectClass'], $start, $treeSize, $parentRoot)` call above). For the cases under test `$treeSize` is always `2` - a freshly persisted leaf starts at `left=1, right=2`, so `treeSize = right - left + 1 = 2` - which is why the tests pass. But for a non-root node that already carries a subtree (`$treeSize > 2`), `+= 2` under-advances the counter and the next root will overlap again, reintroducing the bug this PR fixes.

Suggest matching the gap the branch actually opens:

```suggestion
            $this->treeEdges[$meta->getName()] += $treeSize;
```

This keeps every current test green (there `$treeSize === 2`) while making the increment correct for the general case rather than incidentally correct for single-leaf children.

src/Tree/Strategy/ORM/Nested.php
```diff
@@ -406,6 +406,9 @@ public function updateNode(EntityManagerInterface $em, $node, $parent, $position
                 $wrapped->setPropertyValue($config['right'], $right);
             }
             $newRoot = $parentRoot;
+
+            $this->treeEdges[$meta->getName()] ??= $this->max($em, $config['useObjectClass'], $newRoot) + 1;
```
phansys: Nice that this mirrors the root branch's seeding. One thing worth confirming: `$treeEdges` is keyed only by class name (`$meta->getName()`), while `max()` here is seeded with `$newRoot` (`= $parentRoot`). For this PR's fixture that is fine because `OU` has no `@Gedmo\TreeRoot` column, so `$newRoot` is `null` and numbering is global. It would be good to verify (or note) that the class-keyed cache still behaves for the separate-root-column configuration, which is not exercised by the tests here.

tests/Gedmo/Tree/Issue/Issue2582Test.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $this->listener = new TreeListener();
+
+        $evm = new EventManager();
+        $evm->addEventSubscriber($this->listener);
+
+        $this->getDefaultMockSqliteEntityManager($evm);
+    }
+
+    public function testInsertTwoRootsInOneFlush(): void
```
phansys: To lock in the general case behind the `+= $treeSize` change, it would help to add a scenario where a non-root node with an existing subtree is inserted before another root in the same flush (i.e. `$treeSize > 2`), and - if feasible - one variant using a dedicated `@Gedmo\TreeRoot` column. That would cover the two paths the current fixtures do not.

phansys (commented):
Your instinct here is right to flag. The approach - advancing `treeEdges` in the child branch so a later root does not allocate an overlapping range - is the correct layer and the correct idea: the child branch grows the tree via `shiftRL(..., $treeSize, ...)`, and the edge counter needs to reflect that growth. The one thing that makes it "works but I am not sure it is correct" is the hardcoded `+= 2`: it equals `$treeSize` only for fresh leaf children, which is all the tests cover. Using `+= $treeSize` makes it model the actual gap and generalises to subtree inserts. With that change (plus a test for a `treeSize > 2` insert), this moves from incidentally-correct to provably-correct. Left a couple of inline notes with the specifics.

PR #2935, opened 2025-03-19
[WIP] Bump the requirements to PHP 8.1

mbabker:
According to the Packagist stats there was a large fall off PHP 7 users after the 3.14 release, which is when the requirement was bumped up to PHP 7.4.  Since then, PHP 7.4 and 8.0 combined have had pretty small numbers on newer release branches and is generally now under 5% of the install base.  So this PR bumps the requirement to PHP 8.1.

The overwhelming majority of the code changes are from adjusting the Rector and PHP-CS-Fixer configs to apply PHP 8.1 rulesets, with a second pass specifically on the tests to add type declarations there, then only minor corrections or tweaks by my own review along the way.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,21 +40,21 @@
         "docs": "https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/doc"
     },
     "require": {
-        "php": "^7.4 || ^8.0",
-        "doctrine/collections": "^1.2 || ^2.0",
+        "php": "^8.1",
+        "doctrine/collections": "^1.6.5 || ^2.0",
         "doctrine/deprecations": "^1.0",
         "doctrine/event-manager": "^1.2 || ^2.0",
-        "doctrine/persistence": "^2.2 || ^3.0 || ^4.0",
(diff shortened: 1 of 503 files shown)


composer.json
```diff
@@ -40,20 +40,20 @@
         "wiki": "https://github.com/Atlantic18/DoctrineExtensions/tree/main/doc"
     },
     "require": {
-        "php": "^7.4 || ^8.0",
-        "behat/transliterator": "^1.2",
-        "doctrine/collections": "^1.2 || ^2.0",
+        "php": "^8.1",
+        "behat/transliterator": "^1.3",
+        "doctrine/collections": "^1.6.5 || ^2.0",
```
mbabker, line 45: Just for reference, all the changes to the minimum versions are either to get to a functional set of dependencies on PHP 8.1 (the transliterator package has errors on older versions, `doctrine/common` doesn't support some PHP 8.1 reflection changes until 3.1) or change to match the versions currently being installed on the `prefer-lowest` build.  Considering how old some of these releases are it shouldn't be a big deal, I hope.

PR #2957, opened 2025-06-02
Add a deprecations baseline to ignore selected deprecations in tests

mbabker:
Symfony 7.3 deprecates the lazy ghosts functionality in the VarExporter component in favor of the native PHP lazy objects features.  As this is an unfixable deprecation from this package's perspective, the indirect deprecation notices this raises can be ignored.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -67,7 +67,7 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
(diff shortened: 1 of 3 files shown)


phpunit.xml.dist
```diff
@@ -67,7 +67,7 @@
         </testsuite>
     </testsuites>
     <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0&amp;max[indirect]=21" />
```
phansys, line 70: Do you think we could use a more specific approach for the target cases, instead of the number of occurrences?
I mean, something like a [baseline](https://symfony.com/doc/current/components/phpunit_bridge.html#baseline-deprecations).

Issue #2955, opened 2025-05-22
Change in behavior with SoftDeleteable in v3.20


HypeMC:
#### Package

<details><summary>show</summary>
<p>

```
$ composer show --latest gedmo/doctrine-extensions
name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.20.0
released : 2025-04-04, 1 month ago
latest   : v3.20.0 released 2025-04-04, 1 month ago
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : http://gediminasm.org/
source   : [git] https://github.com/doctrine-extensions/DoctrineExtensions.git ea1d37586b8e4bae2a815feb38b177894b12c44c
dist     : [zip] https://api.github.com/repos/doctrine-extensions/DoctrineExtensions/zipball/ea1d37586b8e4bae2a815feb38b177894b12c44c ea1d37586b8e4bae2a815feb38b177894b12c44c
path     : /7.2-gedmo-bug/vendor/gedmo/doctrine-extensions
names    : gedmo/doctrine-extensions

support
docs : https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/doc
issues : https://github.com/doctrine-extensions/DoctrineExtensions/issues
source : https://github.com/doctrine-extensions/DoctrineExtensions/tree/v3.20.0

autoload
psr-4
Gedmo\ => src/

requires
behat/transliterator ^1.2
doctrine/collections ^1.2 || ^2.0
doctrine/deprecations ^1.0
doctrine/event-manager ^1.2 || ^2.0
doctrine/persistence ^2.2 || ^3.0 || ^4.0
php ^7.4 || ^8.0
psr/cache ^1 || ^2 || ^3
psr/clock ^1
symfony/cache ^5.4 || ^6.0 || ^7.0

requires (dev)
doctrine/annotations ^1.13 || ^2.0
doctrine/cache ^1.11 || ^2.0
doctrine/common ^2.13 || ^3.0
doctrine/dbal ^3.7 || ^4.0
doctrine/doctrine-bundle ^2.3
doctrine/mongodb-odm ^2.3
doctrine/orm ^2.20 || ^3.3
[...]

Issue #2947, opened 2025-05-12
Changes made to an entity during the `PRE_SOFT_DELETE` event are no longer taken into account.


ker0x:
#### Package

<details><summary>show</summary>
<p>

```
name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.20.0
released : 2025-04-04, 1 month ago
latest   : v3.20.0 released 2025-04-04, 1 month ago
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : http://gediminasm.org/
source   : [git] https://github.com/doctrine-extensions/DoctrineExtensions.git ea1d37586b8e4bae2a815feb38b177894b12c44c
dist     : [zip] https://api.github.com/repos/doctrine-extensions/DoctrineExtensions/zipball/ea1d37586b8e4bae2a815feb38b177894b12c44c ea1d37586b8e4bae2a815feb38b177894b12c44c
path     : /home/niamor/dev/guestspot/vendor/gedmo/doctrine-extensions
names    : gedmo/doctrine-extensions

support
docs : https://github.com/doctrine-extensions/DoctrineExtensions/tree/main/doc
issues : https://github.com/doctrine-extensions/DoctrineExtensions/issues
source : https://github.com/doctrine-extensions/DoctrineExtensions/tree/v3.20.0

autoload
psr-4
Gedmo\ => src/

requires
behat/transliterator ^1.2
doctrine/collections ^1.2 || ^2.0
doctrine/deprecations ^1.0
doctrine/event-manager ^1.2 || ^2.0
doctrine/persistence ^2.2 || ^3.0 || ^4.0
php ^7.4 || ^8.0
psr/cache ^1 || ^2 || ^3
psr/clock ^1
symfony/cache ^5.4 || ^6.0 || ^7.0

requires (dev)
doctrine/annotations ^1.13 || ^2.0
doctrine/cache ^1.11 || ^2.0
doctrine/common ^2.13 || ^3.0
doctrine/dbal ^3.7 || ^4.0
doctrine/doctrine-bundle ^2.3
doctrine/mongodb-odm ^2.3
doctrine/orm ^2.20 || ^3.3
[...]

Issue #2942, opened 2025-05-05
Migration to 3.20 breaks our test suite


gnutix:
#### Package

<details><summary>show</summary>
<p>

```
$ composer show --latest gedmo/doctrine-extensions
name     : gedmo/doctrine-extensions
descrip. : Doctrine behavioral extensions
keywords : Blameable, behaviors, doctrine, extensions, gedmo, loggable, nestedset, odm, orm, sluggable, sortable, timestampable, translatable, tree, uploadable
versions : * v3.19.0
released : 2025-02-24, 2 months ago
latest   : v3.20.0 released 2025-04-04, 1 month ago
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : http://gediminasm.org/
source   : [git] https://github.com/doctrine-extensions/DoctrineExtensions.git 5b0b8a442d19e6701ae64535dc08f7944e2895d2
dist     : [zip] https://api.github.com/repos/doctrine-extensions/DoctrineExtensions/zipball/5b0b8a442d19e6701ae64535dc08f7944e2895d2 5b0b8a442d19e6701ae64535dc08f7944e2895d2
path     : /app/vendor/gedmo/doctrine-extensions
names    : gedmo/doctrine-extensions

support
email : gediminas.morkevicius@gmail.com
issues : https://github.com/doctrine-extensions/DoctrineExtensions/issues
source : https://github.com/doctrine-extensions/DoctrineExtensions/tree/v3.19.0
wiki : https://github.com/Atlantic18/DoctrineExtensions/tree/main/doc

autoload
psr-4
Gedmo\ => src/

requires
behat/transliterator ^1.2
doctrine/collections ^1.2 || ^2.0
doctrine/deprecations ^1.0
doctrine/event-manager ^1.2 || ^2.0
doctrine/persistence ^2.2 || ^3.0 || ^4.0
php ^7.4 || ^8.0
psr/cache ^1 || ^2 || ^3
psr/clock ^1
symfony/cache ^5.4 || ^6.0 || ^7.0

requires (dev)
doctrine/annotations ^1.13 || ^2.0
doctrine/cache ^1.11 || ^2.0
doctrine/common ^2.13 || ^3.0
doctrine/dbal ^3.7 || ^4.0
doctrine/doctrine-bundle ^2.3
doctrine/mongodb-odm ^2.3
doctrine/orm ^2.20 || ^3.3
[...]

PR #2958, opened 2025-06-04
Revert PR #2930

### LINKED ISSUE #2956: Please revert PR #2930
See discussion on #2930 and related bug reports.

HypeMC:
Closes #2956

As mentioned in [this discussion](https://github.com/doctrine-extensions/DoctrineExtensions/pull/2930#issuecomment-2902427288), the original PR seems to have introduced a few new issues: #2942, #2947, and #2955.

Maybe it makes sense to revert it for now, at least until there's a fix for the original problem that doesn't cause new ones. At the moment, it feels like it's creating more trouble than it's solving.


kopfsalat:
Hi,
I would like to add Issue #2963 to the list. For now, we can use the CompilerPass, but I would much prefer #2930 to be reverted as it seems to have too many side effects.

rotdrop:
@phansys As you merged #2930, could you please also have a look and undo it? Kind Thanks!

HypeMC:
> On the listener, it probably doesn't need the constructor parameter and the setter for the option, just the setter would be fine (and consistent with all the other bits that can be configured on a listener).

@mbabker I can remove it if needed. I included it because the `ReferencesListener` defines both a constructor argument and an adder method.
 
> If it's possible, having some kind of failing test case over the exact issues that the changes here cover would be really helpful going forward because the changes in behavior that this post-flush change made starts getting deep into the internals of how the listener interacts with the object managers, and it's probably something that shouldn't be accidentally changed in a new way that breaks things later.

@mbabker The test I added fails when the new behavior is enabled.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
+### Added
+- SoftDeleteable: Add option to enable or disable handling of the `postFlush` event (#2958)
```
phansys, line 22: IIUC, this PR should also have at least an entry for the `Changed` and `Fixed` sections, because the added option changes the way the previous fix behaves, while fix the issues mentioned in the description.

Could you please confirm?

HypeMC: Hi. Yes you're right. I've updated the changelog.

phansys (approved):

phansys:
Thank you @HypeMC!

Issue #1021, opened 2014-03-21
TranslationWalker does not run when Query Cache is enabled


flip111:
I had a bug with translatable, it was firing queries even though registering the walker and hinting the right locale was done properly. After tracing the problem i came to the conclusion that Query Caching should be disabled for queries using Translatable.
To disable the query cache on a query: `$query->useQueryCache(false);`

Reasoning:
If query cache is used the Parser is never invoked:
https://github.com/doctrine/doctrine2/blob/5c44dd38234f2e143e376ac94d2e7f945d7f9328/lib/Doctrine/ORM/Query.php#L248

If the Parser does not run then no custom output walkers are checked:
https://github.com/doctrine/doctrine2/blob/4f585a3d6372ef1115a30663fe0005f611decd76/lib/Doctrine/ORM/Query/Parser.php#L357

I think the difference between a `HINT_CUSTOM_TREE_WALKERS` (SQL Walker) and `HINT_CUSTOM_OUTPUT_WALKER` (output walker) is that the first changes the actual DQL query and the latter always has to be done afterwards.

This is possibly a limitation of Doctrine

To do: gedmo translatable should automatically turn off query caching when it needs to register an output walker

acasademont:
Yep, same here. Even though the query is executed correctly with the translation joins the "skipOnLoad" method is not called and the translation objects are fetched...

l3pp4rd:
hints are used in cache key, it also is mentioned in the documentation. should be clear, read carefully

acasademont:
@l3pp4rd I have already read the docs,  and I have even contributed extensively in the past to the Translatable component and I perfectly know how it works, no need to be rude here.

It has nothing to do with hints. When the query is cached the custom translatable hydrators are not executed (these hydrators are set on the `TranslationWalker` and the walker is obviously never run when the query is already cached) and therefore the `setSkipOnLoad` method is not called and the `TranslatableListener` does again all the work.

The workaround is to manually set the custom translatable hydrators when executing the query.

So I would kindly ask you to reopen the issue.

flip111:
@l3pp4rd 
I believe only the original query is cached (inside the querycache, not the resultcache). The post-processing done by hints on the query yields a new query ... and so i think this new query (with join for translate table) is NOT cached in the query cache. If this is not true `hints are used in cache key` please be more explicit to the actual code where the modified query is cached.

[...]

PR #2960, opened 2025-06-18
[Translateable] Document the usage of refresh query hint if cache is used

stollr:
This addition to the documentation helps to prevent for running into the issue #1021.

This is the issue in short:
If query caching is enabled the `TranslationWalker` won't be used anymore once the query is cached. Therefore the `Query::HINT_REFRESH`, which is normally set by the `TranslationWalker`, won't be set automatically.
This leads to the issue, that once a translatable entity is loaded a second query with another locale won't update the translatable properties in the entity.

Adding the `Query::HINT_REFRESH` manually to the query will fix the issue.

### DIFF
--- a/doc/translatable.md
+++ b/doc/translatable.md
@@ -506,6 +506,11 @@ $query->setHint(
     \Gedmo\Translatable\TranslatableListener::HINT_FALLBACK,
     1 // fallback to default values in case if record is not translated
 );
+// refresh entities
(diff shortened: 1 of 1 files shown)


phansys:
I am closing and reopening this PR to trigger fresh CI workflows.

phansys:
Thank you @stollr!

PR #2978, opened 2025-08-22
[SoftDeleteable] Remove dollar sign from cache key

adambalint-srg:
This PR removes dollar sign ($) from the default cache key. It's not a globally reserved characters in filenames, but there are cases when it can cause problems, as I read. This is why e.g. [Laminas's filesystem cache disables this character.](https://github.com/laminas/laminas-cache-storage-adapter-filesystem/blob/3.2.x/src/FilesystemOptions.php#L26) I think using `__` instead of `_$` isn't a problem, and won't cause any problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ a release.
 ---
 
 ## [Unreleased]
-### Fixed
-- SoftDeleteable: Remove dollar sign ($) from the default cache key of soft deleteable metadata - to be compatible with stricter cache key naming conventions (see https://www.php-fig.org/psr/psr-6/#definitions) (#2978)
+### Changed
+- All: Removed the dollar sign from the generated cache ID for extension metadata to ensure only characters mandated by [PSR-6](https://www.php-fig.org/psr/psr-6/#definitions) are used, improving compatibility with caching implementations with strict character requirements (#2978)
 
 ## [3.22.0] - 2025-12-13
 ### Added


VincentLanglet:
Hi, do you have any example when it's an issue ?

adambalint-srg:
In my application (Laminas+Doctrine stack) I'm using this repo's extensions, and for Doctrine I'm using Laminas's filesystem cache. So the cacheId extension metadata will be used in file system cache. Now it throws an error. I've created a quick example where you can reproduce:

```json
{
    "name": "test/test",
    "require": {
		"php": "^8.3",
		"laminas/laminas-cache-storage-adapter-filesystem": "^3.1.0",
		"gedmo/doctrine-extensions": "^3.20.1"
	}
}
```

```php
<?php

declare(strict_types=1);

include 'vendor/autoload.php';

use Gedmo\Mapping\ExtensionMetadataFactory;
use Laminas\Cache\Storage\Adapter\Filesystem;

$cacheId = ExtensionMetadataFactory::getCacheId('Foo', 'Bar');
//cacheId is 'Foo_$BAR_CLASSMETADATA';

$fileSystemCache = new Filesystem();
$fileSystemCache->addItem($cacheId, ['some' => 'data']);
//Uncaught Laminas\Cache\Exception\InvalidArgumentException: The key 'Foo_$BAR_CLASSMETADATA' doesn't match against pattern '/^[a-z0-9_\+\-\.]*$/Di'
```

mbabker:
Since this library requires a PSR-6 cache, and from what I can tell the Laminas cache doesn't directly implement that PSR (it does provide [a decorator for this](https://docs.laminas.dev/laminas-cache/v4/psr6/)), is there an issue when you use their PSR-6 decorator with this library?  (On a side note, I'd say the `ExtensionMetadataFactory::getCacheId()` really should've been marked internal and not used externally or in the exact way that reproducer snippet is written, but it's a little late for that now)

From a practical perspective, there isn't an issue with this PR (and presumably improves PSR-6 compatibility by ensuring only characters the spec explicitly says an implementation must support are used).  But, it seems that the MongoDB ODM [also uses a dollar sign in its cache keys](https://github.com/doctrine/mongodb-odm/blob/2a9118a6266b29f778fbfde5e4acb19a19cc25fc/lib/Doctrine/ODM/MongoDB/Mapping/ClassMetadataFactory.php#L57) which would imply that ODM and Laminas Cache are incompatible for the same reason.

adambalint-srg:
@mbabker 
I'm using PSR-6 decorator in my application. But this decorator does nothing with the cache keys. 

```php
/** @var StorageAdapterFactoryInterface $storageFactory */
		$storageFactory = $container->get(StorageAdapterFactoryInterface::class);

		/** @var StorageInterface&FlushableInterface $fileCache */
		$fileCache = $storageFactory->create('filesystem', $options);

		return new CacheItemPoolDecorator($fileCache);
```

The direct use of the `getCacheId` is only for this example, I don't use it directly.

phansys (commented):
Thank you @adambalint-srg!

As you are presenting this change as a fix, please add an entry in the `Fixed` section of the changelog, including an explanation about what it fixes.

See: https://github.com/doctrine-extensions/DoctrineExtensions/blob/main/CONTRIBUTING.md#changelog.

Thanks also to @mbabker for the detailed analysis.

VincentLanglet:
It's on @phansys side, I dont have any right on this project

adambalint-srg:
@phansys I did a changelog update after this unresolved comment (9f129431ebf97764d18389dceb876a0e4279390f), but didn't get any update or feedback about this.

phansys:
> @phansys Currently I can't see any commits to rebase. Yesterday I did a conflict resolve, and it says now my branch is ahead of main, and not behind. Or I don't see something?

This PR currently contains more commits than those related to your changes:
image

This situation introduces conflicts that prevent the PR to be merged using the **Rebase** strategy:

image

PR #2994, opened 2025-10-06
[SoftDeleteable] Document the `$handlePostFlushEvent` option

HypeMC:
Docs for #2958.


stof:
@phansys it would be great to have this merged so that the new option is documented.

PR #3003, opened 2025-10-29
Run tests with MongoDB ODM using native laxy objects, fix other deprecations

mbabker:
As advertised


tests/bootstrap.php
```diff
@@ -45,3 +46,8 @@
 }
 
 Type::addType('uuid', UuidType::class);
+
+// Ignore unfixable deprecations
+Deprecation::ignoreDeprecations(
+    'https://github.com/doctrine-extensions/DoctrineExtensions/pull/2772', // Ignore annotations deprecations from self
+);
```
mbabker: This is useful if you're running with `<env name="DOCTRINE_DEPRECATIONS" value="trigger"/>` in a custom `phpunit.xml` to see more of the triggered deprecations

tests/Gedmo/Sortable/Fixture/Node.php
```diff
@@ -15,7 +15,7 @@
 use Gedmo\Sortable\Entity\Repository\SortableRepository;
 
 /**
- * @author Charles J. C. Elling, 2017-07-31
```
phansys: @mbabker, do you know what rule is changing these annotations?

BTW, I think we should separate this change in a different PR to confirm if it must be kept.

PR #3031, opened 2026-01-20
Address deprecations from the Collections package in the nested tree strategy

mbabker:
Fixes deprecated use of Criteria with raw field access not enabled and use of the deprecated order constant (as feasible)


umulmrum:
I just came across this problem and add a comment so the PR doesn't get auto-closed. It would be great if this could be finished/merged :slightly_smiling_face:

phansys:
Thank you @mbabker for the PR and @sdespont for the reporting the issue!

PR #3033, opened 2026-01-28
Replace `ClassMetadata::getReflectionProperty` with `(get|set)FieldValue`

### LINKED ISSUE #3032: ReflectionProperty is deprecated in ORM 3.5+
`ReflectionProperty` in ORM si deprecated since 3.5: https://github.com/doctrine/orm/pull/12083
The replacement is `PropertyAccessor`, which is available since 3.4: https://github.com/doctrine/orm/pull/11659

`PropertyAccessor` has `setValue` and `getValue` methods, which should be sufficient for all uses in this library. I performaed a quick scan of the codebase and found only `->setAccessible(true)` as an other method called onto `ReflectionProperty` (which are obtained via `ClassMetaData::getReflectionProperty`).

Would it be acceptable to create a small static wrapper like this?

```php

function getProperty(ClassMetadata $meta): ReflectionProperty|PropertyAccessor|null
{
    if (method_exists(ClassMetadata::class, 'getPropertyAccessor')) {
        return $meta->getPropertyAccessor();
    }

    return $meta->getReflectionProperty();
}
```

...and use it everywhere in the codebase to solve this issue?

Jean85:
This avoids a new deprecation.

Fixes #3032

### DIFF
--- a/tests/Gedmo/Mapping/Mock/Extension/Encoder/EncoderListener.php
+++ b/tests/Gedmo/Mapping/Mock/Extension/Encoder/EncoderListener.php
@@ -83,11 +83,16 @@ class EncoderListener extends MappedEventSubscriber
         $om = $ea->getObjectManager();
         $meta = $om->getClassMetadata(get_class($object));
         $uow = $om->getUnitOfWork();
+
+        if (!method_exists($meta, 'getFieldValue') || !method_exists($meta, 'setFieldValue')) {
+            throw new \RuntimeException('ClassMetadata does not have (get|set)FieldValue() methods: '.get_class($meta));
+        }
+
         foreach ($config['encode'] as $field => $options) {
-            $value = $meta->getReflectionProperty($field)->getValue($object);
+            $value = $meta->getFieldValue($object, $field);
             $method = $options['type'];
             $encoded = $method($options['secret'].$value);
-            $meta->getReflectionProperty($field)->setValue($object, $encoded);
+            $meta->setFieldValue($object, $field, $encoded);
         }
         // recalculate changeset
         $ea->recomputeSingleObjectChangeSet($uow, $meta, $object);
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1039,24 +1039,6 @@ parameters:
 			path: tests/Gedmo/DoctrineExtensionsTest.php
 
 		-
-			message: '#^Property Gedmo\\Tests\\Mapping\\Fixture\\Attribute\\TranslatableModel\:\:\$title \(string\|null\) is never assigned string so it can be removed from the property type\.$#'
-			identifier: property.unusedType
-			count: 1
-			path: tests/Gedmo/Mapping/Fixture/Attribute/TranslatableModel.php
-
-		-
-			message: '#^Property Gedmo\\Tests\\Mapping\\Fixture\\Attribute\\TranslatableModel\:\:\$titleFallbackFalse \(string\|null\) is never assigned string so it can be removed from the property type\.$#'
-			identifier: property.unusedType
-			count: 1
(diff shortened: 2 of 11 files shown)


src/Tool/ORM/Entity/Meta.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\ORM\Mapping\PropertyAccessors\PropertyAccessor;
+
+class Meta
+{
+    /**
+     * @param ClassMetadata<object> $meta
+     */
+    public static function getProperty(ClassMetadata $meta, string $propertyName): \ReflectionProperty|PropertyAccessor|null
```
mbabker: This isn't compatible with PHP 7.4, union types aren't available.

tests/Gedmo/Mapping/Mock/Extension/Encoder/EncoderListener.php
```diff
@@ -84,10 +85,10 @@ private function encode(EventAdapterInterface $ea, object $object, array $config
         $meta = $om->getClassMetadata(get_class($object));
         $uow = $om->getUnitOfWork();
         foreach ($config['encode'] as $field => $options) {
-            $value = $meta->getReflectionProperty($field)->getValue($object);
+            $value = Meta::getProperty($meta, $field)->getValue($object);
```
mbabker: The test failures around this line mean if you're going to use a static helper (which really needs to be flagged `@internal`), it needs to support the persistence `ClassMetadata` interface and not a single implementation of it.  There are a lot of code paths which are supporting MongoDB ODM and ORM (this one included, see the `Gedmo\Tests\Mapping\ExtensionODMTest::testGeneratedValues` failure on PHP 8.5) so a bulk search/replace isn't feasible.

mbabker (changes requested):

tests/Gedmo/Mapping/Mock/Extension/Encoder/EncoderListener.php
Jean85: Roger. I'm pushing an alternative approach, it should be enough because it relies on direct method existence checks; unfortunately the involved methods are not present in the interface.

Jean85:
@mbabker I've changed my approach as suggested, but I've discovered some additional stuff:
 * all code paths that I changed are in tests -- should I move the helper to the test folder?
 * I opened this PR because I thought I needed to solve some deprecations being triggered on my project, while all is already solved in 3ba0096a0a95022e9fa4ade7ee571794fe29b9d5, I just need to update... I can still complete this PR if you want

mbabker:
> * I opened this PR because I thought I needed to solve some deprecations being triggered on my project, while all is already solved in [3ba0096](https://github.com/doctrine-extensions/DoctrineExtensions/commit/3ba0096a0a95022e9fa4ade7ee571794fe29b9d5), I just need to update... I can still complete this PR if you want

At this point I'd just change it similar to that commit/PR.  After https://github.com/doctrine/persistence/pull/451 `getFieldValue()` and `setFieldValue()` are part of the metadata interface and not just something provided by every implementation, so there's no need for consumers to know if they need a reflection property or a property accessor (that's all handled by the metadata).

Jean85:
@mbabker I did it; the two failing tests are unrelated, I see them failing on main too, locally at least.

Jean85:
@mbabker now the build is green; I had to fix it with f0960ac, let me know if that's ok.

mbabker (approved):

Jean85:
Merged main, CI is green, is this ready to merge?

Jean85:
@phansys rebased 👍

phansys:
Thank you @Jean85!

PR #3037, opened 2026-02-23
feat: allow doctrine/collections 3

Chris53897:
needs a ci run if affected from BC breaks.
https://github.com/doctrine/collections/releases/tag/3.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/collections": "^1.2 || ^2.0",
+        "doctrine/collections": "^1.2 || ^2.0 || ^3.0",
         "doctrine/deprecations": "^1.0",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "doctrine/persistence": "^2.2 || ^3.0 || ^4.0",
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/collections": "^1.2 || ^2.0 || ^3.0",
+        "doctrine/collections": "^1.2 || ^2.0 || ^3.1",
         "doctrine/deprecations": "^1.0",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "doctrine/persistence": "^2.2 || ^3.0 || ^4.0",
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/collections": "^1.2 || ^2.0 || ^3.1",
+        "doctrine/collections": "^1.2 || ^2.0 || ^3.0",
         "doctrine/deprecations": "^1.0",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "doctrine/persistence": "^2.2 || ^3.0 || ^4.0",
--- a/src/Tree/Strategy/ORM/Nested.php
+++ b/src/Tree/Strategy/ORM/Nested.php
@@ -480,7 +480,7 @@ class Nested implements Strategy
                 $criteria->andWhere(Criteria::expr()->eq($config['level'], 0));
                 $order = 'ASC';
                 if (function_exists('enum_exists')) {
-                    $order = enum_exists(\SortDirection::class) ? \SortDirection::Ascending : (enum_exists(Order::class) ? Order::Ascending : 'ASC');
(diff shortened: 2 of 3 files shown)


Chris53897:
The unrelated failing tests should be fixed in a seprate PR.
`doctrine/collections 3` is only in the ci runs for PHP >= 8.4

mbabker:
FWIW this is blocked by upstream:

ORM - https://github.com/doctrine/orm/pull/12341 (merged, needs 3.7 release)
MongoDB ODM - https://github.com/doctrine/mongodb-odm/issues/2986 (no PR yet it appears)

Chris53897:
Thanks for pointing that out. I added `WIP` to the title

VincentLanglet:
Hi @Chris53897 ; https://github.com/doctrine/orm/releases/tag/3.7.0 is released :)

Chris53897:
To support all 3 versions deprecations free this line in src/Tree/Strategy/ORM/Nested.php needs to be changed (or ignored in PHPStan).
As Order is deprecated in favor of SortDirection. Available via PHP 8.6 polyfill.

Actual
`$criteria->orderBy([$config['right'] => function_exists('enum_exists') && enum_exists(Order::class) ? Order::Ascending : Criteria::ASC]);`
or simple solution
`$criteria->orderBy([$config['right'] => function_exists('enum_exists') && enum_exists(Order::class) ? Order::Ascending : 'ASC']);`

Please let me know how you want this to be handled.
a) ignored
b) hardcoded
c) Solution with SortDirection

In my opionen the other 2 PHPStan Errors are not directly related.

VincentLanglet:
Can't you use `Criteria` when it's defined (which means an old version) and `ASC` if not ?

```
$criteria->orderBy([$config['right'] => 
      function_exists('enum_exists') && enum_exists(Order::class) ? Order::Ascending : (
           function_exists('enum_exists') && enum_exists(Criteria::class) ? Criteria::ASC : 'ASC'
      )
]);
```
?

mbabker:
> Can't you use `Criteria` when it's defined (which means an old version) and `ASC` if not ?

`Criteria` was never an enum, the deprecation here is around the `ASC` and `DESC` constants.  TBH, I'd say roll with the "simple" solution of just using the string values when neither of the enums are available (and `symfony/polyfill-php86` makes the native `SortDirection` available, so it's theoretically possible to use it, but, that also runs the risk of having compat issues with Collections 1.x or 2.x `¯\_(ツ)_/¯`).

In my mind, a practical world solution is this (assuming it doesn't break older versions, which it probably will, considering Collections 3.1 is when the native enum support was added):

```php
use Doctrine\Common\Collections\Order;

$order = 'ASC';

if (function_exists('enum_exists')) {
    $order = enum_exists(\SortDirection::class) ? \SortDirection::Ascending : (enum_exists(Order::class) ? Order::Ascending : 'ASC');
}
```

Chris53897:
Thanks @mbabker and @VincentLanglet 

@phansys ready for review

composer.json
```diff
@@ -41,7 +41,7 @@
     },
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/collections": "^1.2 || ^2.0",
+        "doctrine/collections": "^1.2 || ^2.0 || ^3.0",
```
whataboutpereira:
```suggestion
        "doctrine/collections": "^1.2 || ^2.0 || ^3.1",
```

doctrine/orm itself requires ^3.1, so maybe align with that.

src/Tree/Strategy/ORM/Nested.php
```diff
@@ -478,7 +478,10 @@ public function updateNode(EntityManagerInterface $em, $node, $parent, $position
                 $criteria->andWhere(Criteria::expr()->eq($config['root'], $node->$method()));
                 $criteria->andWhere(Criteria::expr()->isNull($config['parent']));
                 $criteria->andWhere(Criteria::expr()->eq($config['level'], 0));
-                $criteria->orderBy([$config['right'] => function_exists('enum_exists') && enum_exists(Order::class) ? Order::Ascending : Criteria::ASC]);
+                $criteria->orderBy([$config['right'] => function_exists('enum_exists') && enum_exists(Order::class) ? Order::Ascending : (
+                    function_exists('enum_exists') && enum_exists(Criteria::class) ? Criteria::ASC : 'ASC'
```
HypeMC, line 482: I'm wondering why `enum_exists(Criteria::class)` is needed here. `Criteria` has never been an enum and still isn't one, so I don't think this check is correct.

Also, `Order` is deprecated as of v3.1. `\SortDirection` should be used instead.

Wouldn't this be something like:

```php
enum_exists(\SortDirection::class)
    ? \SortDirection::Ascending
    : (enum_exists(Order::class) ? Order::Ascending : 'ASC');
```

PR #3045, opened 2026-03-26
[Tree] Nested set treeRootNode option - fix tricky and improve performance

wilfi00:
Hello!

I came across a "bug" when using the `recover` method with the `treeRootNode` option:
```
->recover(
            [
                'treeRootNode' => $rootStructure
            ]
);
```

In `NestedTreeRepository`, the code loops through each root node and compares the current node with the one provided in the options. However, comparing objects can be tricky in PHP, and in my case, it wasn't working.

My root node is a Doctrine entity. When I pass this node as an option, it has a fully instantiated `User` attribute. But `getRootNodes()` returns an entity with a lazy-loaded `User` object. Because of this, the comparison was wrongly returning false.

The attribute of the root node I give in option:
image

The attribute of the root node being iterated by `NestedTreeRepository`:
image

I suggest a refactor to unify the logic whether it's a forest, a specific root node, or a single tree. This way, we avoid the direct object comparison entirely, and it works like a charm.

I am not very familiar with this type of contribution, so feel free to request changes or ask questions :pray: 

edit: I've also updated the `verify` method to apply the same logic.

Thank you,
Johan

### DIFF
--- a/src/Tree/Entity/Repository/NestedTreeRepository.php
+++ b/src/Tree/Entity/Repository/NestedTreeRepository.php
@@ -1034,29 +1034,20 @@ class NestedTreeRepository extends AbstractTreeRepository
 
         // if it's a forest
         if (isset($config['root'])) {
-            foreach ($this->getRootNodes($options['sortByField'], $options['sortDirection']) as $root) {
-                // if a root node is specified, recover only it
-                if (null !== $options['treeRootNode'] && $options['treeRootNode'] !== $root) {
-                    continue;
-                }
-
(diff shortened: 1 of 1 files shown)


wilfi00:
Could I have an answer to this, please? Thanks :pray:

phansys (commented):
Thanks for your first contribution @wilfi00!

As you are reporting something that isn't working for your use case, the PR requires a test covering the expected behavior to avoid regressions.
An entry at `Fixed` section from `CHANGELOG.md` is also required to explain other users what this PR fixes.

PR #3046, opened 2026-03-28
Fix Loggable template bounds for Symfony TypeInfo compatibility

### LINKED ISSUE #3044: Incompatibility with symfony/type-info > 8.0.4
Hello, 
I'm reaching you, but I'm not sure this is an issue in the Gedmo side. 

Since symfony/type-info 8.0.6 (this commit: https://github.com/symfony/type-info/commit/4855ceea609b2c09e48ff76e12a97a3955531735), they also gather types thanks to @phpstan-template and @psalm-template. 

For example in AbstractLogEntry class we can find this UnionType `Loggable|object`

```php
/**
 * @phpstan-template T of Loggable|object
```

Which break in type-info UnionType class because we can't create an UnionType with both a class and object
(https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/TypeInfo/Type/UnionType.php#L74)

```php
if (\in_array(TypeIdentifier::OBJECT, $builtinTypesIdentifiers, true) && \count(array_filter($this->types, fn (Type $t): bool => $t instanceof ObjectType))) {
    throw new InvalidArgumentException('Cannot create union with both "object" and class type.');
}
```

Thanks you

Prometee:
Fixes #3044.
## What
This PR updates 3 Loggable PHPDoc template bounds that currently combine `Loggable` and `object` in the same bound:
- `src/Loggable/LoggableListener.php`
- `src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php`
- `src/Loggable/LogEntryInterface.php`
The bound is simplified to `object`.
## Why
Recent versions of `symfony/type-info` parse `@phpstan-template` / `@psalm-template` annotations and reject bounds combining `object` with a specific class/interface.
This causes the incompatibility reported in #3044.
## Scope / BC
- PHPDoc-only change
- No runtime behavior changes
- No API signature changes

### DIFF
--- a/src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
+++ b/src/Loggable/Entity/MappedSuperclass/AbstractLogEntry.php
@@ -15,7 +15,7 @@ use Gedmo\Loggable\LogEntryInterface;
 use Gedmo\Loggable\Loggable;
 
 /**
- * @phpstan-template T of Loggable|object
+ * @phpstan-template T of object
  *
  * @phpstan-implements LogEntryInterface<T>
  *
--- a/src/Loggable/LogEntryInterface.php
+++ b/src/Loggable/LogEntryInterface.php
@@ -12,7 +12,7 @@ namespace Gedmo\Loggable;
 /**
  * Interface to be implemented by log entry models.
  *
- * @phpstan-template T of Loggable|object
+ * @phpstan-template T of object
  *
  * @author Javier Spagnoletti <phansys@gmail.com>
  */
--- a/src/Loggable/LoggableListener.php
+++ b/src/Loggable/LoggableListener.php
@@ -36,7 +36,7 @@ use Gedmo\Tool\Wrapper\AbstractWrapper;
  *   versioned?: string[],
  * }
  *
- * @template T of Loggable|object
+ * @template T of object
  *
  * @phpstan-extends MappedEventSubscriber<LoggableConfiguration, LoggableAdapter>
  */


mbabker:
I dare suggest this should go the other way around and just be an object.

None of the extensions require entities to implement the marker interfaces (of which this Loggable interface is one), so the generic type only supporting the marker interface would create other static analysis issues if you aren't implementing them.

(Long term since they offer no value to this library, maybe they should just be deprecated and let applications have their own marker interfaces if they so choose?)

mbabker (approved):

chrisyawman:
Hello, can we get this merged?

lruozzi9:
Hello @phansys, can you have a look at this PR?

lruozzi9 (approved):

laurentdabbb:
You released a new version without merging this fix ? 😞

phansys:
I closed and immediately reopened this PR to re-trigger the GitHub Actions workflows.

The workflows had never run against the current head commit (`9824f91`): there were zero check runs associated with it, so the PR appeared "clean" only because no checks had reported - not because any check had passed. As a fork PR, the workflows were gated and no run had been queued to approve.

The reopen emitted a fresh `pull_request` event, which queued the workflows. They are now in the `action_required` state, awaiting maintainer approval to run.

No code was changed by this action.

Prometee:
@phansys I rebase onto `main` and force push. Can you allow the CI to run?
If you can also review this PR, this PR is important and could fix issue for several people.

phansys:
Thanks for your patience @Prometee.

Please, rebase the commit to inherit the recent fixes for the CI suite.

Additionally, if this is a fix, please add an entry in the `CHANGELOG.md` and if possible, the related tests to avoid regressions.

Prometee:
@phansys rebased, `CHANGELOG.md` edited and regression tests added.

phansys:
Thank you @Prometee!

PR #2889, opened 2024-10-18
Reduce deprecations due to FieldMapping array access

Jean85:
Fixes #2827, at least in large part.

I did a simple regex find&replace of access to `$mapping['...']`, and then:
 * added parentheses everywhere, since `??` (null coalescing) has one of the [lowest operator precedence in PHP](https://www.php.net/manual/en/language.operators.precedence.php)
 * removed the parentheses (with another regex) when:
     * it was already inside a pair of them
     * it was an argument (so already separated by parentheses and\or a comma)
 * reviewed all the changes manually and:
     * reverted the changes in the `ODM` class (since there is no deprecation there)
     * reverted all the times where it was inside an `isset` call (it wouldn't work)
     * removed the parentheses manually when it was already followed by a `??` (so it's a concatenation)

I realize this may not cover all the places where this deprecation may be triggered, but I'm sure that all my changes are covering some, so it's a good start.

PR #3054, opened 2026-07-15
fix: avoid deprecation with nullable fields

VincentLanglet:
Hi @phansys 

The last PR from @Jean85 (https://github.com/doctrine-extensions/DoctrineExtensions/pull/2889) works fine except for the SlugListener.

Since `length` and `nullable` are optional properties they are nullable ; which means something like
```
$mapping->nullable ?? $mapping['nullable'] ?? null
```
might still try to access to the array.

All the CI failure are unrelated to this PR, but I fixed them in another commit.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -73,12 +73,6 @@ parameters:
 			path: src/Loggable/Entity/Repository/LogEntryRepository.php
 
 		-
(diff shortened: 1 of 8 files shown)


phansys:
Thank you @VincentLanglet!

VincentLanglet:
Thanks ; does a 3.22.1 release is possible @phansys ? :)

PR #3065, opened 2026-09-20
Update CI config to pin MongoDB packages on PHP < 8.1

phansys:
Recent releases of `mongodb/mongodb` (>= 1.21) and `doctrine/mongodb-odm` (>= 2.6) require PHP ^8.1. Since the older, PHP 7.4/8.0-compatible `mongodb/mongodb` releases (<= 1.20) are all covered by security advisories, Composer's advisory policy blocks their installation. As a result, the PHP 7.4 and 8.0 CI legs began failing at the `composer install` step, which in turn masked and then exposed a number of latent test-suite issues on those legs.

This PR restores a fully green matrix without dropping any core functionality. The MongoDB ODM integration remains a first-class, tested feature on every PHP version it can run on.

### MongoDB ODM on PHP < 8.1

The MongoDB ODM stack cannot be installed on PHP 7.4/8.0 with the modern releases, and its only PHP < 8.1-compatible releases carry security advisories. To keep the matrix green while preserving ODM support:

- On the standard PHP 7.4/8.0 legs, `doctrine/mongodb-odm` is removed and the ODM test suites are excluded through a dedicated `phpunit.no-odm.xml.dist` configuration (path-level exclusion, so PHPUnit never autoloads classes that reference the absent package).
- A dedicated `mongodb-odm: legacy` matrix leg (PHP 7.4) installs the last ODM stack that runs on PHP < 8.1 (`doctrine/mongodb-odm >= 2.3 < 2.6`, `mongodb/mongodb >= 1.10 < 1.21`) and runs the full test suite, including the ODM tests. Security-advisory blocking is disabled only for this leg (`COMPOSER_NO_SECURITY_BLOCKING`).
- Modern ODM coverage is retained unchanged on the PHP >= 8.1 legs.

This is deliberate: these ODM combinations worked in previous releases, so removing their support in a minor version would be a backward-compatibility break for users who cannot upgrade PHP immediately. Every temporary branch introduced for this purpose is annotated with an `@todo: Remove when support for PHP < 8.1 is dropped.` marker so it can be cleaned up when the PHP floor is eventually raised.

### Test-suite compatibility fixes surfaced by the unblocked legs

Once installation was unblocked, the PHP 7.4 "Without Annotations" leg revealed tests that assumed an annotation reader was always available:

- The `*MappingTest` data providers gated the "Model with annotations" data set on 
  `class_exists(Doctrine\ORM\Mapping\Driver\AnnotationDriver::class)`. That ORM driver class exists on ORM 2 even when `doctrine/annotations` is absent, so the data set was still produced without a usable reader. The guard now uses `class_exists(Doctrine\Common\Annotations\AnnotationReader::class)`, the actual indicator of annotation availability.
- `CustomDriverTest` is skipped when neither attribute mapping (PHP >= 8.0) nor `doctrine/annotations` is available, since its extension configuration requires a reader.

### Static analysis

`ExtensionMetadataFactoryReaderTest` instantiates the ORM `AnnotationDriver`, which no longer exists in `doctrine/orm` 3.x (the version PHPStan analyzes on the highest dependency set). Baseline entries were added for this file, consistent with how every other equivalent test site is handled.

### Testing

Verified locally (PHP 8.2, ORM 3 highest, `doctrine/annotations` present):

- `vendor/bin/phpstan analyse` reports no errors.
- The Mapping test suite passes with the annotation data sets exercised.
- `php-cs-fixer` reports no fixable files.

The PHP < 8.1 legs (including the new legacy ODM leg and the "Without Annotations" leg) are validated by CI, since the legacy MongoDB stack and the annotation-removal scenario cannot be reproduced on a modern local environment.


phansys:
This PR is a step before #2935.

